PR #35, opened 2019-04-10
WIP: counterfactuals

jklaise:
This pull request will implement the counterfactual search algorithm from [Wachter et al. (2017)](https://arxiv.org/abs/1711.00399).

- [x] Define a prediction function to be optimized (options: 'same' to optimize towards a target probability of the same class, 'other' - any other class different from the predicted class, and an integer value - find a counterfactual with a specific class)
 - [x] Numerical gradient function for the black-box prediction function (can use `statsmodels.tools.numdiff.approx_fprime`
 - [x] Distance function (`L1`)
 - [ ] Optional `.fit` method if training set is available - scale distance by MADs (use `functools.partial` to make it a 2-argument distance)
 - [x] Calculate gradients of the Wachter loss
 - [x] Minimize loss for some number of steps or until a maximum `lambda` is reached
 - [x] Incorporate tolerance around the target probability
 - [x] Sensible defaults for allowed feature ranges and epsilons in numerical gradient evaluations
 - [ ] Incorporate distance loss into the TF graph for speed
 - [ ] Support TF and Keras models by directly back-propagating through to get prediction gradients
 - [ ] Use bisection to increase/decrease `lambda` to find the counterfactual with the highest possible `lambda` (closest to the test instance)
 - [ ] Look into learning rate scheduling, particularly to accelerate learning if prediction gradients are small

### DIFF
--- a/alibi/explainers/__init__.py
+++ b/alibi/explainers/__init__.py
@@ -6,8 +6,10 @@ from .anchor.anchor_tabular import AnchorTabular
 from .anchor.anchor_text import AnchorText
 from .anchor.anchor_image import AnchorImage
 from .cem.cem import CEM
+from .counterfactual.counterfactual import CounterFactual
 
 __all__ = ["AnchorTabular",
            "AnchorText",
            "AnchorImage",
-           "CEM"]
+           "CEM",
+           "CounterFactual"]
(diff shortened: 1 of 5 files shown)


jklaise:
Closing in favour of #78

PR #78, opened 2019-05-10
WIP: Counterfactual instances

jklaise:


jklaise:
This pull request will implement an improved counterfactual search algorithm based on [Wachter et al. (2017)](https://arxiv.org/abs/1711.00399).

TODO:
 - [ ] Optional `.fit` method if training set is available - scale distance by MADs (use `functools.partial` to make it a 2-argument distance)
 - [ ] Incorporate distance loss into the TF graph for speed
 - [ ] Support TF and Keras models by directly back-propagating through to get prediction gradients (requires TF version of sorting and getting the highest non-test class probability for `target_class=other`)
 - [ ] Fix bug in learning rate scheduling (global step updating but learning rate is constant)
 - [ ] Optimize search speed / sensible hyperparameter choices

Optional:
 - [ ] Use FISTA for optimization
 - [ ] Option to offer different losses instead of just L1/L1 scaled by MADs

PR #80, opened 2019-05-10
WIP: Influence Functions

jklaise:
This will implement the influence function method from https://arxiv.org/abs/1703.04730 . As the method requires access to the full model (to backpropagate end-to-end), the initial work is on supporting Keras models.


jklaise:
TODO: 

The algorithm is functional (meaning it works and returns influential instances which seem reasonable to the eye), however it is not converging to the right thing. I.e. experiments with leave-one-out retraining show that the influence scores are not correlated with the actual loss uplift, so there is a bug in the core algorithm.

The code is mostly based on a re-implementation by https://github.com/darkonhub/darkon, the original research code is available at https://github.com/kohpangwei/influence-release . I suspect there may be a difference in the two code bases which are causing the issue in our implementation.

jklaise:
Have realized that the validation difficulties were caused by using SGD on the leave-one-out re-training, in the paper the authors validate the method on multinomial logistic regression, but crucially they use the LBFGS solver which is more robust and coupled with the convexity of logistic regression is the perfect way to validate the influence function approximation empirically. I will see if I can replicate the results using this approach, potentially using pre-trained scikit-learn weights to initialize a frozen Keras model.

In other news, have decided that the datafeeder abstraction is unnecessary. Will remove that and re-work the api to conform to `.fit` and `.score`.

jklaise:
Closing as any new updates or version will be written in TF2.0 + PyTorch.

PR #110, opened 2019-07-03
Prune lexemes with no word vectors before finding synonyms

jklaise:


axsaucedo:
Based on [Spacy docs](https://spacy.io/api/lexeme#has_vector) lgtm

PR #113, opened 2019-07-09
WIP: Initial explainer API

jklaise:
The proposed explainer+explanation API is inspired by MS Interpret.

Current features:

- Any explainer inherits from `Explainer` and must implement `explain` which returns an explanation object
- Some explainers have an optional `fit` using data, this is achieved using inheritance from `FitMixin` to force implementing `fit`
- Any explanation object inherits from `Explanation` and must implement `data` method which can be called (with optional keys, corresponding to a specific sample in a batch) to return a JSON serializable dict or list
- Rationale for an explanation object - allows to extend definition to implement various method acting on the explanation, e.g. `plot`

To discuss:
- Should local/global explanations have different methods? E.g. MS Interpret define `explain_local` and `explain_global` and expose available methods via `available_explanations` attribute
- The signature of the `data` method could be more loose - allow slicing
- Other attributes, what are the explainer/explanation types? (`local`, `global`, or something more specific like `feature_attribution`?)
- Maybe specifying `np.ndarray` in abstract method signatures is too strict?


jklaise:
I think we can actually push the `local` vs `global` distinction in the `Explanation` implementations instead of the `Explainer` stuff. I.e. for most things local explainers will operate on a batch of features `X` and perform some calculation for each `x` in `X`, while global methods will still take `X` but calculate something across all of `X`, so the explainer signature should be unchanged, while what is returned should reflect the differences.

jklaise:
Moving towards the following spec for the data of the explanation:

```json
meta: {},
global: {},
local: {
 "x1": {},
 "x2": {},
 ...
}
```

`meta` field would contain all the metadata relating to how the explanation was created (e.g. hyperparameters of the explainer) as well as metadata relevant the explanation itself (e.g. `"type":"blackbox"`, `"scope":"local"`).

All explanations will have a `global` and a `local` field relating to global and local explanations. In most cases one of these will be empty, but some explainers would populate both fields. Additionally, the `local` field will always be indexed by the instances explained in the batch call to `explain`.

Some examples of proposed spec:

PDP (1-dimensional):
`pdp.explain(X, features, dim=1)` where `features` is the features for which to compute PDP (could be `"all"`). Returns:

```json
meta: {"dim":1}
global: {"feature1" :{...},
         "feature2" :{...},
         ...
}
local: {
 "x1": {"feature1": {...},
        "feature2": {...},
         ...},
 "x2": {...},
...
}
```

`local` here containes the individual ICE lines.

PDP (2-dimensional) would be the same, except the features are indexed by the `n choose 2` products of the `features`, e.g. `"feature1 x feature2" : {}`

SHAP:
`shap.explain(X, ...)`. Returns:

```json
meta: {...}
global: {"feature1" :{"scores:" [...]},
         "feature2" :{...},
         ...
}
local: {
 "x1": {"scores": [...], ...},
 "x2": {...},
...
}
```
Here, even though SHAP is a local method, we can populate the `global` field with a feature-transposed version of the data. This would ease plotting dependence plots (for a given feature, plot SHAP values vs feature values) or summary plots.

jklaise:
- Relaxed the signatures of the abstract `explain` and `fit` methods as currently e.g. `AnchorText` takes `str` instead of `np.ndarray`.
- Maybe worth thinking of ~not~ defining the methods `fit` and `explain` as part of the interface as any changes to arguments will conflict with the assumed signature (e.g. `fit` may need both features `X` and targets `y` for some methods, there is no clean way of allowing this in a subclass without adding boilerplate to methods that wouldn't require `y`, would have to explicitly set `y=None` for all).
- Implemented `AnchorTabular` as an example of the new API. Immediate concern is that accessing data is too cumbersome, e.g. need to call `explanation.data['local'][0]['precision']` to get the precision of the first datapoint in the batch. Could expose a utility function `def local(ix)` that returns the instance of interest, e.g. `explanation.local(0)['precision']`. Unfortunately cannot do the same with `global` as it's a reserved keyword. Also, need to think about how to make functions getting data and properties acting as attributes be as consistent as possible.

jklaise:
- Implemented `Sequence` protocol for `Explanations`, now one can access local instance explanations by direct indexing, e.g. `explanation[0]['precision']`

jklaise:
- Renamed `global` to `overall` to allow exposition via attributes, i.e. can now call `explanation.global` and `explanation.local[0]`.
- Populate `name` field in `meta` by default using the Explainer/Explanation class name. Remains to discuss whether we want the same name for explainer and explanation, e.g. currently this would be `AnchorTabular` for `explainer.meta["name"]` and `AnchorTabularExplanation` for `explanation.meta["name"]`

jklaise:
- Renamed `global` to `overall` to allow exposition via attributes, i.e. can now call `explanation.global` and `explanation.local[0]`.
- Populate `name` field in `meta` by default using the Explainer/Explanation class name. Remains to discuss whether we want the same name for explainer and explanation, e.g. currently this would be `AnchorTabular` for `explainer.meta["name"]` and `AnchorTabularExplanation` for `explanation.meta["name"]`

jklaise:
Update: the main blocker for merging this is that we need to make a decision how to deal with batch explanations (i.e. passing multiple instances to explain at once). The proposed API would assume batches, requiring one to index into the explanation object to retrieve explanations for each instance (`explanation[i]`). However, the actual support of batch explanations within alibi seems to be some time away, so we need to think whether explicit batch indexing is the way to go for now.

jklaise:
Closing in favour of #166 .

PR #120, opened 2019-07-29
Initial TF session management

jklaise:
We want to move the TF session management internally to the explainers that use TF, this is a first attempt taking `CounterFactual` as an example.

The main difficulty is that with `tf.keras` this works out of the box (tests passing etc.), but using `keras` is throwing `FailedPreconditionError`, so there is some incompatibility with initialization between `tf.keras` and `keras`.


jklaise:
This may be solved by conditionally importing either `tensorflow.keras.backend` or `keras.backend` depending on the model passed in to get the model's session.

jklaise:
Update:

- `CounterFactual` works with both `tf.keras` and `keras` models now
- `CounterFactualProto` is more complex as it can have multiple models passed in (model itself, autoencoder and encoder). Currently works when all components are made with either `tf.keras` or `keras`, but no mix-and-match. Left to investigate whether this can be supported automatically. Worst case, we can have `sess` as an optional argument if people want to pass in mixed components, so the burden then would be on the user to match up the sessions.

jklaise:
Remaining things to land this: CEM and CFProto example notebooks and documentation needs updating.

jklaise:
@arnaudvl I'm puzzled how this can work in the black-box case (which is the notebook example) as `self.predict` would be a balck-box function and `self.adv, self.adv_s` are tensors: https://github.com/SeldonIO/alibi/blob/d7def9c32eaf84204f2bf4b45b2c3e0e4b4dade8/alibi/explainers/cfproto.py#L231-L234

jklaise:
After chatting to @arnaudvl it makes sense that a black-box function which is a keras model in disguise (i.e. via `predict=model.predict`) would not work, as it already has an attached session which we discard when initializing `tf.Session` for black-box functions. So we need a way to get the default session regardless if the passed object is a model or a black-box function. This gets a bit complicated as we still need to somehow check that what is passed in is derived from `tf.keras` or `keras`.

jklaise:
The only thing left to do is to produce a good example for the CEM MNIST notebook - this proves to be quite tricky as the algorithm is quite sensitive to the trained model and hyperparameters.

Other than that, the following (across CEM, CounterFactual and CounterFactualProto) are supported:

- `tf.keras` models
- `keras` models
- black-box models (e.g. `sklearn` models via passing `predict_proba` as a function
- black-box models that are actually `tf.keras` models (e.g. via passing `model.predict`)
- any other `tf` or `keras` model via passing in a `tf` session via the optional `sess` keyword parameter

PR #125, opened 2019-08-22
Add example of how no anchor is found when observation from minority …

### LINKED ISSUE #101: Document the performance of Anchors on imbalanced datasets
Currently the [main example](https://github.com/SeldonIO/alibi/blob/master/examples/anchor_tabular_adult.ipynb) on anchor explanations on tabular datasets will often fail to find a valid anchor for the positive (>$50k) class. This is due to the imbalanced dataset (roughly 25:75 high:low earner proportion), so during the sampling stage feature ranges corresponding to low-earners will be oversampled. This is a feature because it can point out an imbalanced dataset, but it can also be fixed by producing balanced datasets to enable anchors to be found for either class.

MarcoGorelli:
…class is passed

closes #101

### DIFF
--- a/examples/anchor_tabular_adult.ipynb
+++ b/examples/anchor_tabular_adult.ipynb
@@ -374,6 +374,64 @@
     "print('Precision: %.2f' % explanation['precision'])\n",
     "print('Coverage: %.2f' % explanation['coverage'])"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### ...or not?\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


arnaudvl:
Thanks for your contribution @MarcoGorelli ! We also intend to include more sophisticated (incl. balanced) sampling strategies. Because by keeping the features in the proposed anchor fixed while naively sampling the other features, out-of-distribution samples can be generated which does not result in high quality explanations.

MarcoGorelli:
@arnaudvl happy to help, and thanks for your great library!

Issue #28, opened 2019-03-28
Create custom docker image for CI


jklaise:
Our CI requires a lot of custom dependencies and even datasets (e.g. Spacy corpus) which get downloaded and installed every time on the current base Python CI image. Creating our own image which has all the dependencies would make the process significantly faster.

MarcoGorelli:
Currently, the `.travis.yml` uses three Python images (`3.5`, `3.6`, and `3.7`). Should there be three different custom images, one for each of these versions of Python?

jklaise:
I think one could use a single image and manage the different Python versions for CI using virtualenv/conda. That being said, the CI docker image would probably be distinct from a docker image containing `alibi` to be used immediately (it would only include a python interpreter, alibi and all its dependencies), in which case it would make sense to have different versions. Were you thinking of the CI or the user facing image?

MarcoGorelli:
I was thinking of the CI (speeding up the 
```install:
  - pip install -r requirements/requirements_all.txt
  - python -m spacy download en_core_web_md
```
part)

jklaise:
Yes, that would be the main reason for using a custom image (and we could use it for new builds and releases too). Were you interested in contributing a Dockerfile for this?

MarcoGorelli:
I'm interested in attempting this - yes. I was thinking that we could have a Dockerfile of the kind (possibly passing in the different Python versions as an argument)
```
FROM python:3.7

COPY requirements/requirements_all.txt ./

RUN pip install -r requirements_all.txt
RUN python -m spacy download en_core_web_md
```
, building a Docker image, pushing it to DockerHub (e.g. as marcogorelli/alibi-ci:latest), and then using this in the travis yml file as follows:
```
dist: xenial
language: python
image: marcogorelli/alibi-ci:latest
before_install:
  - sudo apt-get install pandoc
script:
  - flake8 alibi
  - mypy alibi
  - travis_wait pytest
  - make build_pypi
  - make build_docs
after_success:
  - codecov
```
(I could even add the `apt-get install pandoc` step to the Dockerfile).

Before I run any of this, I just wanted to check how we would go about such a pull request.

1. Is this what you had in mind? At present, I wouldn't know how to manage different Python versions with virtualenv/conda and just a single Docker image, but I can look into it.
2. The travis file shouldn't reference a Docker image that's in my name. So, should the process be:
- 1 pull request with Dockerfile
- you build the Docker image, and push it to Docker hub, and let me know its tag
[...]

PR #126, opened 2019-08-28
WIP (don't merge!) Docker ci

MarcoGorelli:
Just a test to see whether using a Docker image during the CI process would work / save time / be worth it. Even if it does work, there's still other work that needs to be done to it anyway.


MarcoGorelli:
Likely not going to help (see comments in #28)

PR #147, opened 2019-09-20
Colour pertubations according to black-red cmap

### LINKED ISSUE #143: Could the CEM explanations contain extra colours?
The [CEM paper](https://arxiv.org/pdf/1802.07623.pdf) shows an example image which is made very clear via the use of different colours (cyan for pertinent positive mode, pink for pertinent negative):
Screenshot from 2019-09-17 19-13-13

On the other hand, the `cem_mnist` example from the alibi docs contains black-and-white images. I think this makes the pertinent negative example a little hard to understand, as it's not obvious which are the extra pixels which are minimally but critically absent. Could they be coloured in?

I can work on a PR if you're interested in this.

MarcoGorelli:
closes #143 

Just an idea for how to display pertinent negatives - it was an excellent exercise to go through the notebook in detail anyway, so I won't feel offended if you don't use this :)


jklaise:
Thanks for the contribution! It looks interesting, but I'm not sure this is the best way to convey the PN pixel changes (maybe a different colour map would make it more intuitive?). The thing with the CEM algorithm is that it is very sensitive to hyperparameters, even between instances and PN/PP mode. The images in the original paper look quite good coloured in, but we suspect those examples were a bit cherry-picked. Similarly, we had to cherry-pick examples for the notebook just because for a lot of them the PN/PPs are not particularly visually interpretable.

MarcoGorelli:
Yes, agreed - I'll close this for now then

Issue #151, opened 2019-10-03
Accumulated Local Effects (ALE)


jklaise:
A global explanation method, see https://christophm.github.io/interpretable-ml-book/ale.html . Now that we have distance measures between categories in Alibi we are in a position to implement this for categorical features as well as numerical ones.

jklaise:
Minimal requirements:
- 1-way numerical ALE
- 2-way numerical ALE
- 1-way categorical ALE
- Clean visualisations for all of the above

Secondary requirements (probably another PR):
- 2-way categorical ALE
- 2-way numerical/categorical mixed ALE
- caching of previously calculated results
- batching for large datasets
- ability to do all 1-ways/2-ways in parallel

jklaise:
Initial implementation in #152 .

PR #152, opened 2019-10-07
Accumulated Local Effects (ALE)

jklaise:
Addresses #151 

Edit 04/05/2020:
Requirements for merging:
 - [x] 1-way numerical features ALE
 - [x] Flexible plotting using `matplotlib`
 - [x] 2 examples (1 regression, 1 classification)
 - [x] Method documentation
 - [x] Usage documentation
 - [x] Implementation tests 
 - [x] Convergence test (on Linear regression) 
 - [x] Add references to docs, readme etc. (can be done in a separate PR)

Edit 03/04/2020:
 - Reduce scope to support 1-way numerical and categorical features in the first release of ALE.

Previous:
```
- [x] 1-way numerical features
- [ ] 2-way numerical features
- [ ] 1-way categorical features
- [ ] 2-way mixed numerical/categorical features
- [ ] plotting
  ~~- need different interpolation methods (e.g. constant on any grid square or linear interpolation from the middle point?)~~ (ALE definition prescribes linear interpolation between interval endpoints)
  - [x] 1-way numerical plots (`altair`)
  - [ ] 2-way numerical plots
  - [ ] 1-way categorical plots
  - [ ] 2-way mixed numerical/categorical plots
- [ ] support for classifiers
  - [x] 1-way numerical features done, should generalize as only need to take care to vectorize calculations
```


alibi/explainers/ale.py
```diff
@@ -0,0 +1,66 @@
...
+    ALE values for the feature
+
+    """
+    # TODO the following onlt works for regression
+    q = get_quantiles(X[:, feature], num_points=num_points)
+    ale = np.zeros(len(q) - 1)
+
+    for i in range(len(q) - 1):
+        # TODO this misses the data point equal to the upper quantile
+        subset = X[(q[i] <= X[:, feature]) & (X[:, feature] < q[i + 1])]
```
MarcoGorelli: Just a suggestion (which takes care of the TODO)

```
    indices = np.digitize(X[:, feature], q)
    for i in range(1, num_points):
        if i == num_points - 1:
            subset = X[indices >= i]
        else:
            subset = X[indices == i]
```

jklaise: This wouldn't quite work, first the order would be reversed (using side="right" would be closer to the original code. More importantly, this will still not pick up the last point (e.g. x[indices==num_points]. I think the point equal to the last quantile should be merged with the previous quantile, although it's probably not a big deal if it's not used at all.

MarcoGorelli: First - you're right, the bottom snipped should use `side="right"`, sorry about that, but I think the top snippet works.

If we start with
```
import numpy as np

np.random.seed(42)
X = np.random.randn(5, 2)
feature = 0
values = X[:, feature]
num_points = 3

percentiles = np.linspace(0, 100, num=num_points)
q = np.percentile(values, percentiles, axis=0)
```

then

```
for i in range(len(q) - 1):
    subset = X[(q[i] <= X[:, feature]) & (X[:, feature] < q[i + 1])]
    print(subset)
```
and
```
indices = np.searchsorted(q, X[:, feature], side="right")
for i in range(1, num_points):
    subset = X[indices == i]
    print(subset)
```
produce the same output.

Both miss the last point, so to capture it, we can change it to
```
indices = np.searchsorted(q, X[:, feature], side="right")
for i in range(1, num_points):
    if i == num_points - 1:
        subset = X[indices >= i]
    else:
        subset = X[indices == i]
    print(subset)
```
then we do capture the point equal to the uppermost quantile (here, `[ 1.57921282  0.76743473]`).

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+             fig_kw: Optional[dict] = None) -> 'np.ndarray':
+    """
+    Plot ALE curves on matplotlib axes.
+
+    Parameters
+    ----------
+    exp
+        An `Explanation` object produced by a call to the `ALE.explain` method.
+    features
+        A list of features for which to plot the ALE curves. Can be a mix of integers denoting
```
alexcoca, line 328: By this time I already knew they were lists.

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+        feature index or strings denoting entries in `exp.feature_names`. Defaults to 'all'.
+    targets
+        A list of targets for which to plot the ALE curves. Can be a mix of integers denoting
+        target index or strings denoting entries in `exp.target_names`. Defaults to 'all'.
+    n_cols
+        Number of columns to organize the resulting plot into.
+    sharey
+        A parameter specifying whether the y-axis of the ALE curves should be on the same scale
+        for several features. Possible values are 'all', 'row', None.
```
alexcoca, line 337: backticks as these are values params can take?

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+    constant
+        A parameter specifying whether the constant zeroth order effects should be added to the
+        ALE first order effects.
+    ax
+        A matplotlib axes or a numpy array of matpotlib axes to plot on.
```
alexcoca, line 342: A `matplotlib` axes _object_ [...]

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+    import matplotlib.pyplot as plt
+    from matplotlib.gridspec import GridSpec
+
+    if features == 'all':
+        features = range(0, len(exp.feature_names))
+    else:
+        for ix, f in enumerate(features):
+            if isinstance(f, str):
+                try:
+                    f = np.argwhere(exp.feature_names == f).item()
```
alexcoca, line 362: is `feature_names` not a list? we could use `feature_names.index()` and avoid casting to numpy array etc

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+            features[ix] = f
+    n_features = len(features)
+
+    if targets == 'all':
+        targets = range(0, len(exp.target_names))
+    else:
+        for ix, t in enumerate(targets):
+            if isinstance(t, str):
+                try:
+                    t = np.argwhere(exp.target_names == t).item()
```
alexcoca, line 374: See above.

alibi/explainers/ale.py
alexcoca, line 328: I think we should be clear that the entries of the list can be that mix. If I pass simply `features = 'sex'` then I'll get errors in ` L359`

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+    Parameters
+    ----------
+    exp
+        An `Explanation` object produced by a call to the `ALE.explain` method.
+    features
+        A list of features for which to plot the ALE curves. Can be a mix of integers denoting
+        feature index or strings denoting entries in `exp.feature_names`. Defaults to 'all'.
+    targets
+        A list of targets for which to plot the ALE curves. Can be a mix of integers denoting
+        target index or strings denoting entries in `exp.target_names`. Defaults to 'all'.
```
alexcoca, line 332: Same comment as for `features` regarding errors.

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+        if isinstance(ax, plt.Axes):
+            ax = np.array(ax)
+        if ax.size < n_features:
+            raise ValueError(f"Expected ax to have {n_features} axes, got {ax.size}")
+        axes = np.atleast_2d(ax)
+        axes_ravel = axes.ravel()
+        fig = axes_ravel[0].figure
+
+    # make plots
+    if line_kw is None:
```
alexcoca, line 417: maybe would be nice to have the defults and properties handling at the start of the fcn?

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+                              constant=constant,
+                              ax=ax_ravel,
+                              legend=not ix,  # only one legend
+                              line_kw=line_kw)
+
+    # if explicit labels passed, handle the legend here as the axis passed might be repeated
+    if line_kw['label'] is not None:
+        axes_ravel[0].legend()
+
+    default_fig_kw = {'tight_layout': 'tight'}
```
alexcoca, line 435: same comments as above.

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+
+    ax.set_xlabel(exp.feature_names[feature])
+    ax.set_ylabel('ALE')
+
+    if legend:
+        # if no explicit labels passed, just use target names
+        if line_kw['label'] is None:
+            ax.legend(lines, exp.target_names[targets])
+
+    return ax
```
alexcoca, line 479: Your plotting code is excellent.

alibi/explainers/tests/test_ale.py
```diff
@@ -0,0 +1,83 @@
...
+from alibi.explainers.ale import ale_num, adaptive_grid, get_quantiles
+from alibi.api.defaults import DEFAULT_DATA_ALE, DEFAULT_META_ALE
+
+
+@pytest.mark.parametrize("min_bin_points", [1, 4, 10])
+def test_ale_num_linear_regression(min_bin_points):
+    """
+    The slope of the ALE of linear regression should equal the learnt coefficients
+    """
+    X, y = load_boston(return_X_y=True)
```
alexcoca, line 14: I'm more of a fan of the approach where one has a fixture for data, one for the model an combines the two rather than hardcoding in the test. The test needs to check something about ale function, and it's nice if we can abstract the model and the data away (maybe there comes a time when we need to check why this fails in production and it would be nice to have the ability to add datasets/models). This results in harder work as additional code as other parts of the codebase show, but makes testing more flexible (but maybe harder to implement the first time).

alibi/explainers/tests/test_ale.py
```diff
@@ -0,0 +1,83 @@
...
+    lr = LinearRegression().fit(X, y)
+    for feature in range(X.shape[1]):
+        q, ale, _ = ale_num(lr.predict, X, feature=feature, min_bin_points=min_bin_points)
+        assert np.allclose((ale[-1] - ale[0]) / (X[:, feature].max() - X[:, feature].min()), lr.coef_[feature])
+
+
+@pytest.mark.parametrize("min_bin_points", [1, 4, 10])
+def test_ale_num_logistic_regression(min_bin_points):
```
alexcoca, line 22: Same comment as above

alibi/explainers/tests/test_ale.py
```diff
@@ -0,0 +1,83 @@
...
+@pytest.mark.parametrize("min_bin_points", [1, 4, 10])
+def test_ale_num_linear_regression(min_bin_points):
+    """
+    The slope of the ALE of linear regression should equal the learnt coefficients
+    """
+    X, y = load_boston(return_X_y=True)
+    lr = LinearRegression().fit(X, y)
+    for feature in range(X.shape[1]):
+        q, ale, _ = ale_num(lr.predict, X, feature=feature, min_bin_points=min_bin_points)
+        assert np.allclose((ale[-1] - ale[0]) / (X[:, feature].max() - X[:, feature].min()), lr.coef_[feature])
```
alexcoca, line 18: NB: Check [this](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html) out! Think this is also something we should standardize across our codebase.

alibi/explainers/tests/test_ale.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+@pytest.mark.parametrize("min_bin_points", [1, 4, 10])
+def test_ale_num_logistic_regression(min_bin_points):
+    """
+    The slope of the ALE curves performed in the logit space should equal the learnt coefficients.
+    """
+    X, y = load_iris(return_X_y=True)
+    lr = LogisticRegression(max_iter=200).fit(X, y)
```
alexcoca, line 27: I remember the anchor tabular test paramterized this already?

doc/source/overview/algorithms.md
```diff
@@ -27,6 +28,11 @@ Key:
  - **Global** - explains the model with respect to a set of instances
  - **Cat. data** - support for categorical features
  - **Train** - whether a training set is required to fit the explainer
+
+**Accumulated Local Effects (ALE)**: calculates first-order feature effects of a model with
```
alexcoca, line 32: feature effects _on_ the model?

doc/source/methods/ALE.ipynb
```diff
@@ -0,0 +1,285 @@
+{
```
alexcoca, line 1: <p>I would abide by the rules of academic objectivity in the 2nd paragraph and simply say.</p><p> </p><p>_ALE addresses some key shortcomings of Partial Dependence Plots, a popular method of estimating first order feature effects. We discuss these limitiations and motivate ALE after presenting the method usage._</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

alexcoca, line 1: <ul><li>_a default argument, <code>min_bin_points = 4</code>, -&gt; commas are missing</li><li>I feel we could spell _data related_ as <em>data-related</em></li><li>random comment: what is your opinion on when having <code>min_bin_points</code> actually hurts us? Can we think of an example of a feature where this is not so big of a deal?</li><li>what exactly do those values in <code>ale0</code> represent. A quick explanation (if possible) or reference to somewhere where I can get an explanation would be helpful.</li><li>The construction _one for each feature_ is both parenthesized and used as an incident sentence - we can be consistent?</li><li>Not sure what ALE_iris is doing there? I think that's an img that's not rendering ...</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

alexcoca, line 1: <p>I like this, maybe it's a better name than _Theoretical overview_ and we should use it throughout the entire library.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

alexcoca, line 1: <ul><li>missing a comma in $x = (x_1, ...x_d)$ before $x_d$?</li><li>missing a comma after _$d$ features_ ?</li><li>missing comma after <em>notation</em> in _To simplify notation_?</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

alexcoca, line 1: <ul><li>I think you may have forgotten the space before/after $$\hat{M}(x_1)$$ as this does not render well ... Potentially as blank line is missing after $$M(x_1) = ...$$.</li><li>So the estimation of the conditional tells us to fix $x_1$ and then check for some instances that have a similar enough $x_1$. How do we select N(x1) in practice? </li><li>Esentially, what we are saying is that fixing x1 to some value changes the distribution of x2 (which could we see from the plots below), and that this biases M(x1), right?</li><li>Maybe worth stating this [_ Furthermore, suppose height doesn't actually have any effect on the prediction, only weight does_] mathematically as the case where the function _f(x_1, x_2) = x_1_ has _M(x_2) \neq 0_ which is not intuitive. </li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

alexcoca, line 1: <p>NB: This image does not appear in the rendered docs.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

alexcoca, line 1: <ul><li>In NB review, this doesn't render but it's fine in the docs. </li><li>I can't imagine what _vertically centered_ means. Is there anywhere in the examples (which I haven't read yet) which shows us the differencce that constant makes? </li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

alexcoca, line 1: <ul><li>Here I would say _replace the first feature value_ simply _replace $x_1$ with_ as that's easier to parse.</li><li>Just to make it clear that $x_{\1}^{(i)}$ means we leave all the other features unchanged? In which case it would be worth adding _and leave the rest of the features unchanged_ to make it easier to read quickly?</li><li>For c1, is calculated quanity = ALE(x1)? If you can it would be nice to discuss the different options for estimating this expectations (from a practical standpoint) as I have seen the code does consider some options.</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

alexcoca, line 1: <p>Plot not shown in rendered docs.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

alexcoca, line 1: <p>Plots do not render</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

alexcoca, line 1: <p>Nice presentation, well done. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

examples/ale_regression_boston.ipynb
```diff
@@ -0,0 +1,790 @@
+{
```
alexcoca, line 1: <p>I would have put _feature effects on the model_ to emphasise this is a global method (so applies to <em>any</em> prediction not specified predictions (_the predictions_))</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

alexcoca, line 1: <p>We usually do <code> from alibi.explainers import ALE</code>  ?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

alexcoca, line 1: <p>It's usually helpful to print the feature names so that the user has a clue what are the independent variables if they don't run the nb and just look at the example.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

alexcoca, line 1: <p>The equation does not appear in the rendered docs.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

alexcoca, line 1: <p>Would have been nice to set  <code>FEATURE</code>  to be a string and then do <code>feature_names.index(FEATURE)</code> to get the axis labels.  That way, the user can also redo your work for other features if they are interested and don't want to print the list and count the position of the index they care about...</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

alexcoca, line 1: <p>We could make this easier to read by putting a full stop after _one might expect_ and write _However,_</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

alexcoca, line 1: <ul><li>We can separate _in this case_ in between commas as it's an incident construction?</li><li>I think we should refer to the present plot given one is plotted so we can say "The ALE on the y-axis of the plot above ..." or even "The units of the y-axis of the plot above" ... since it's clear from the label we are plotting the ALE value...</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='33'/>

alexcoca, line 1: <p>Missing : </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='40'/>

alexcoca, line 1: <p>This is helpful to understand -&gt; This helps understand</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='44'/>

alexcoca, line 1: <ul><li>maybe "danger" and "safely" are strong words, nothing's at stake. I would use terms like "unrealistic", "not a reliable estimate" etc.</li><li>Last sentence is somewhat redundant, I was well aware by that point that you are interpolating between the points. </li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='46'/>

alexcoca, line 1: <ul><li>Could start a new sentence AFTER "with respect to the mean feature effect". Now I don't really get what you're comparing the magnitude and shape against ? The previous example for linear model? If so, you've already said plots are nonlinear and the reader can hopefully understand this is a different model and draw this conclusion 4 themselves. No need for it to be spelled out. </li></ul><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='56'/>

alexcoca, line 1: <ul><li>Missing comma after <em>plot</em>s</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='57'/>

alexcoca, line 1: <ul><li>Missing comma after _To do this_</li></ul><p> </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='59'/>

alexcoca, line 1: <p>Missing comma before <em>as evidenced</em>  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='61'/>

alexcoca, line 1: <p>Has anyone proved consistency for ALE? What I mean is that if we take two models A and B, and we know feature x is more important in B vs A, then does ALE tell us the same story? I think we should be able to generate some functions of the same variables to check this empirically. This might be a future development and we can talk about it.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='63'/>

examples/ale_classification.ipynb
```diff
@@ -0,0 +1,787 @@
+{
```
alexcoca, line 1: <p>You should state why I should read the regression example (e.g., "to familiarise yourself with how to interpret ALE plots is simple settings")</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

alexcoca, line 1: <ul><li>Printing feature name would be helpful</li><li>Printing target names would be helpful as below you print some logits and I don't know/have to 2nd guess to which class they correspond </li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

alexcoca, line 1: <p>another way to lay these out would be:</p><ul><li>each row is the same target</li><li>each col is the feature</li></ul><p>I think that could be a bit less sore on the eye but I don't mind the current version too much as it is simple enough</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

alexcoca, line 1: <ul><li>I think we can cross out should in "what we should expect" -&gt; what we expect sounds perfectly fine to me!</li><li>a comma could be added before <em>which means</em> </li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

alexcoca, line 1: <ul><li>vice versa for the other two classes -&gt; this is inaccurate as the effect is anything but large pos/neg for the vericolor cls</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

alexcoca, line 1: <p>Ok, a bit confused here as there is no positive effect of 0.69 in that vector. Maybe we could be a bit more precise and say that all the effects are due to the combined feats except petal length.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

alexcoca, line 1: <ul><li>"which reflcts the non-linear nature of applying the softmax" -&gt; " due the softmax transformation applied to the logits" would be a bit more concise</li><li><em>in this case</em> should be in between commas</li><li>I think i.e. also takes a comma since if you were to be less fancy and say _that is_ you'd need a comma</li><li>Easier to read if we put a full stop of ; after <em>decrease in probability </em><strong><em>of </em></strong><em>(not to?) class </em>and a comma after <em>In fact </em>(or leave it out altogether).</li><li>Maybe worth stating that ALE values for the same interval or something (or the same feature vals) ...</li></ul><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='37'/>

alexcoca, line 1: <ul><li>"virtually flat" could be simplified as "flat" or "has almost zero slope"?</li><li><em>the per-class histograms are simply overlapping too much to offer much discriminative power to the linear model -&gt; the overlap in the per-class histograms of this feature show that it does not increase the model discriminative power</em></li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='39'/>

alexcoca, line 1: <p>I feel like the pause line should have some blank space around it?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='42'/>

alexcoca, line 1: <ul><li>Forgot a space before the bullets start (bullets don't show in rendered version)</li><li>_Whils_ -&gt; W<em>hile</em></li><li>Second bullet point does not seem to be true for the GBM </li><li>Both models do perfectly on test data, so we cannot talk simply about an "increase in discriminative power"; we can say that the models use different features to discern between different classes. An interesting experiment would be to see what would happen if we excluded petal width data, refit the models and re-assessed them. </li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='60'/>

alibi/explainers/__init__.py
```diff
@@ -2,6 +2,7 @@
 The 'alibi.explainers' module includes feature importance, counterfactual and anchor-based explainers.
 """
 
+from .ale import ALE, plot_ale
```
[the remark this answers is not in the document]
jklaise: How do you mean? `plot_ale` takes an `Explanation` object, we had a discussion earlier how we with the setup we have now.

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+                 feature_names: List[str] = None,
+                 target_names: List[str] = None) -> None:
+        """
+        Accumulated Local Effects for tabular datasets. Current implementation supports first order
+        feature effects of numerical features.
+
+        Parameters
+        ----------
+        predictor
+            A callable that takes in an NxF array as input and outputs an NxT array (N - number of
```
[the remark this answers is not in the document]
jklaise: Yea, probably stuff for another PR on updating docstrings. There are some that are out of date too (e.g. CF still say they return a `dict` instead of an `Explanation`).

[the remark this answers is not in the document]
jklaise: Opened an issue for this (#229), so resolving.

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+            data points, F - number of features, T - number of outputs/targets, e.g. 1 for ordinary
+            regression, >=2 for classification).
+        feature_names
+            A list of feature names used for displaying results.
```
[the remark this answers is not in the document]
jklaise: Agree to disagree :)

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+        X
+            An NxF tabular dataset used to calculate the ALE curves, this is typically the training dataset
+            or a representative sample.
+        min_bin_points
+            Minimum number of points each discretized interval should contain to ensure more precise
+            ALE estimation.
+
+        Returns
+        -------
+            An `Explanation` object containing the data and the metadata of the calculated ALE curves.
```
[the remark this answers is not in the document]
jklaise: Another one for #229 I think (also I checked your SHAP code, they don't do this...)

alibi/explainers/__init__.py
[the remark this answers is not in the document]
alexcoca: My argument here is that the individual files will become a bit hard to navigate. Taking a naive example in SHAP, there are lots of different visualisations and each involve quite a bit of code. I think it can get very messy if we don't do something like `alibi.explainers.visualisation` and mirror the file structure. I think it would be easier to maintain, as the code would be easier to track down (e.g., for SHAP you don't need to scroll over the first 1000 lines of wrapper code etc)

alibi/explainers/ale.py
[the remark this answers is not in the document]
alexcoca: The main point is that for complex types it gets too much to say a "a dictionary of tuples where each tuple has a list with a strings" ....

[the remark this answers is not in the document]
jklaise: Agreed, I wouldn't do this for a complex type but I don't think there's any harm in saying it's a list to make the sentence more stand-alone.

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+        self.target_names = np.array(self.target_names)
+
+        feature_values = []
+        ale_values = []
+        ale0 = []
+        feature_deciles = []
+
+        # TODO: use joblib to paralelise?
+        for feature in range(n_features):
+            q, ale, a0 = ale_num(self.predictor,
```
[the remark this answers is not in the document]
jklaise: Hmm, this is why we need something like `black` :), does the same rule apply to methods or just functions? Also, with automatic formatting in Pycharm putting the first argument on a separate line will de-indent everything quite a lot to the left. Happy to go with either.

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+        # I've replaced this with feature deciles which is coarser but has constant space complexity
+        # as opposed to a rugplot. Alternatively, could consider subsampling to produce a rug with some
+        # maximum number of points.
+        return self.build_explanation(ale_values=ale_values,
+                                      ale0=ale0,
+                                      constant_value=constant_value,
+                                      feature_values=feature_values,
+                                      feature_deciles=feature_deciles)
+
+    def build_explanation(self,
```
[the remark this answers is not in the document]
jklaise: Maybe, although for an internal function this is not so crucial. The main thing is to accurately document what is returned to the user.

Agree on consistency in the future.

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+        of points in each resulting bin.
+        """
+        q = np.unique(get_quantiles(values, num_points=n))
+        indices = np.searchsorted(q, values, side='left')
+        indices[indices == 0] = 1
+        interval_n = np.bincount(indices)
+        return int(np.all(interval_n[1:] > min_bin_points))
+
+    # bisect
+    num_points = bisect_fun(fun=lambda x: 1 - minimum_satisfied(x), target=0.5, lo=0, hi=len(values)) - 1
```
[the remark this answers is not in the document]
jklaise: Good point.

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+
+    Returns
+    -------
+    q
+        Unique quantiles.
+    num_points
+        Number of non-unique points the feature array was subdivided into.
+    """
+
+    def minimum_satisfied(n: int) -> int:
```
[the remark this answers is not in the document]
jklaise: Not sure it's worth the additional boilerplate for this, as you said this works because `values` never changes. On the flipside I do use something similar in the tests so maybe it makes sense to factor this out...

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+        """
+        q = np.unique(get_quantiles(values, num_points=n))
+        indices = np.searchsorted(q, values, side='left')
+        indices[indices == 0] = 1
+        interval_n = np.bincount(indices)
+        return int(np.all(interval_n[1:] > min_bin_points))
+
+    # bisect
+    num_points = bisect_fun(fun=lambda x: 1 - minimum_satisfied(x), target=0.5, lo=0, hi=len(values)) - 1
+    assert minimum_satisfied(num_points)
```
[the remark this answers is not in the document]
jklaise: Yep, happy to take it out as I have similar code in the actual test, although imo there's no harm having simple asserts in code.

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+    q = np.unique(get_quantiles(values, num_points=num_points))
+
+    return q, num_points
+
+
+def ale_num(
```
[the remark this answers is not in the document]
jklaise: No particular reason other than what you said - easier to test and understand, also avoids the temptation of changing the state of the object.

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+    df = pd.DataFrame(concat)
+
+    avg_p_deltas = df.groupby(df.shape[1] - 1).mean().values  # groupby indices
+
+    # accummulate over intervals
+    accum_p_deltas = np.cumsum(avg_p_deltas, axis=0)
+
+    # pre-pend 0 for the left-most point
+    zeros = np.zeros((1, accum_p_deltas.shape[1]))
+    accum_p_deltas = np.insert(accum_p_deltas, 0, zeros, axis=0)
```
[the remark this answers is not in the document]
jklaise: I remember having trouble with concat but I don't remember why...

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+
+    # mean effect, R's `ALEPlot` and `iml` version (approximation per interval)
+    ale0 = (((accum_p_deltas[:-1, :] + accum_p_deltas[1:, :]) / 2) * interval_n[1:, np.newaxis]) \
```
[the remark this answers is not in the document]
jklaise: It's a weighted average of the midpoint values over the intervals. The weights are the number of points in each interval (`interval_n`) and the calculation is being normalized by the total number of points.

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+               .sum(axis=0) / interval_n.sum()
+
+    # crude approximation (assume datapoints on interval endpoints)
```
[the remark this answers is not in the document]
jklaise: This is just for future reference, I used the same version the authors used to get exact numbers, afterwards I realized you can be more precise using information of the quantiles (instead of just averaging the midpoints). The crude estimate is not useful but it's the first thing you could think of if approached this.

alibi/explainers/ale.py
jklaise: It is actually an array (conversion to array happens in `explain`). I think I decided to go with arrays because the functionality was more consistent with all the plotting code etc. This is one for #227 imo.

alibi/explainers/tests/test_ale.py
jklaise: Agree!

alibi/explainers/ale.py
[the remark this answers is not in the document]
alexcoca: Agree to disagree :)

alibi/explainers/ale.py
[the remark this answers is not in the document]
alexcoca: There are some places where there is a ref to docs. But it would be nice to standardize when we do #229.

alibi/explainers/ale.py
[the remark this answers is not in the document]
alexcoca: I think both methods and functions. Me and Arnaud use this (seldon core do too). Important is to be consistent . I personally hated the de-indent but learned to live with it!

alibi/explainers/ale.py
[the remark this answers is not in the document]
alexcoca: I get upset when people use the closure of the function, but that's perhaps a personal preference. The docs and everything are a lot less clear/useful. imho it's just an additional line of code stating 

```python
my_fun = functools.partial(_my_fun, values=values)
```
which would also let you serialise the function if you ever needed to!

alibi/explainers/ale.py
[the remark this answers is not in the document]
alexcoca: I agree there is no harm, but they should at least say something to the user as it's a bit upsetting to get an `assert False` error ...

alibi/explainers/ale.py
alexcoca: Yes, we'll decide what the best thing is then. I thought everywhere in the codebase `feature_names` was a list. Missed the line where the conversion happened!

alibi/explainers/tests/test_ale.py
alexcoca: Let's start a testing standard doc to develop in time ...

alibi/explainers/ale.py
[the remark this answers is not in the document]
jklaise: Ah, but asserts shouldn't be used for validation/error handling as they can get compiled away in packaged code. Asserts in code should never really be possible to trigger by anything the user does.

alibi/explainers/ale.py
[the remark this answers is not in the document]
jklaise: I will resolve this then...

alibi/explainers/tests/test_ale.py
jklaise: I agree with this, we will have to rerorganize testing utilities and fixtures at some point as there is quite a lot going on already. I will try to make this change using fixtures, it seems a bit involved as I need access both to the model and the dataset it's parametrized by.

jklaise: I've re-parametrized tests using some new and existing fixtures, please have a look. Also, I haven't updated `SUPPORTED_DATASETS` and `get_dataset` as they were missing some values already and I wasn't clear where these are used in the tests?

alibi/explainers/tests/test_ale.py
```diff
@@ -1,31 +1,41 @@
...
     The slope of the ALE of linear regression should equal the learnt coefficients
     """
-    X, y = load_boston(return_X_y=True)
-    lr = LinearRegression().fit(X, y)
+    lr, _ = lr_regressor
+    X = dataset['X_train']
+
     for feature in range(X.shape[1]):
         q, ale, _ = ale_num(lr.predict, X, feature=feature, min_bin_points=min_bin_points)
         assert_allclose((ale[-1] - ale[0]) / (X[:, feature].max() - X[:, feature].min()), lr.coef_[feature])
```
alexcoca, line 23: Can probs use the same syntax as the next test :)

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+        if fun(mid) < target:
+            lo = mid + 1
+        else:
+            hi = mid
+    return lo
+
+
+def adaptive_grid(values: np.ndarray, min_bin_points: int = 1) -> Tuple[np.ndarray, int]:
+    """
+    Find the optimal number of points to subdivide the feature range into
```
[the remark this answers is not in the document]
jklaise: Replaced with `    Find the optimal number of quantiles for the range of values so that each resulting bin
    contains at least `min_bin_points`. Uses bisection.`

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+        Array of feature values.
+    min_bin_points
+        Minimum number of points each discretized interval should contain to ensure more precise
+        ALE estimation.
+
+    Returns
+    -------
+    q
+        Unique quantiles.
+    num_points
```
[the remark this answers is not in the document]
jklaise: Renamed all occurences of `num_points` to `num_quantiles`

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+def adaptive_grid(values: np.ndarray, min_bin_points: int = 1) -> Tuple[np.ndarray, int]:
+    """
+    Find the optimal number of points to subdivide the feature range into
+    so that each bin has at least `min_bin_points`. Uses bisection.
+
+    Note: This is a heuristic procedure since the bisection algorithm is applied
+    to a function which is not monotonic. This will not necessarily find the
+    maximum number of quantiles the interval can be subdivided into to satisfy
+    the minimum number of points in each resulting bin.
+
```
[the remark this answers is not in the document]
jklaise: Changed `quantiles`-->`bins`

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+    -------
+    q
+        Unique quantiles.
+    num_points
+        Number of non-unique points the feature array was subdivided into.
+    """
+
+    def minimum_satisfied(n: int) -> int:
+        """
+        Calculates whether the partition into n quantiles has the minimum number
```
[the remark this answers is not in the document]
jklaise: Renamed to "        Calculates whether the partition into bins induced by n quantiles
        has the minimum number of points in each resulting bin."

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+    p_high = predictor(z_high)
+
+    # finite differences
+    p_deltas = p_high - p_low
+
+    # make a dataframe for averaging over intervals
+    concat = np.column_stack((p_deltas, indices))
+    df = pd.DataFrame(concat)
+
+    avg_p_deltas = df.groupby(df.shape[1] - 1).mean().values  # groupby indices
```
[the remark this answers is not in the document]
jklaise: Probably one for another PR

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+
+    # accummulate over intervals
+    accum_p_deltas = np.cumsum(avg_p_deltas, axis=0)
+
+    # pre-pend 0 for the left-most point
+    zeros = np.zeros((1, accum_p_deltas.shape[1]))
```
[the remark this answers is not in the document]
jklaise: Maybe I'm missing something, but all the arrays are always 2-dimensional, it even gets validated after `explain` call.

alibi/explainers/ale.py
jklaise: I've changed the type signature to `Union[List[Union[int, str]], Literal['all']]`, although we could be even more epxlicit in the docstring.

alibi/explainers/ale.py
jklaise: See above.

alibi/explainers/ale.py
jklaise: This does require a dependency on the `typing-extensions` module which I've added.

doc/source/methods/ALE.ipynb
jklaise: <ul><li><code>min_bin_points</code>  can be quite sensitive, but would be a case-by-case basis, hard to come up with a general example</li><li>Reworded: "<code>ale0</code> - a list of "centering" values, one for each feature, used by the algorithm to center the <code>ale_values</code> (i.e. the sum of <code>ale_values</code> and <code>ale0</code> will be the uncentered ALE."</li><li> ALE_iris renders fine for me both on the notebook and the static docs, will need to check with you</li><li>All other small points addressed</li></ul>

jklaise: <ul><li>Added blank lines although it rendered fine for me</li><li>Estimating conditional distributions is a hard problem in general, in practice this is done by binning data using small enough quantiles (but big enough so that <code>min_bin_points</code>  is satisfied. I mention this in the estimation formula: "Here z0, z1,... is a sufficiently fine grid..."</li><li>It's not the change of the distribution of x2 that biases M(x1) but rather that M(x1) cannot distinguish between effects of x1 and effects of other features correlated with x2</li><li>Stated!</li></ul>

jklaise: <p>It does...</p>

jklaise: <p>So the idea is that if you don't center it then ALE(x) depends on min(x1) which is kind of an arbitrary point, by centering the plot you essentially get rid of this dependency but also get as a benefit that the ALE is zero in expectation over the marginal distro p(x), meaning that you have the nice interpretation that ALE(x)=0 means that at x you observe an effect that is average wrt p(x). I agree that it's not super clear, will think about this more, but I spent a lot of time understanding this as the paper and Molnar's chapter are not super clear.</p>

jklaise: <ul><li>For calculating c1 I didn't think it's worth going into detail as it's more of a small thing. Agree that calculated quantity is a big ambiguous, but I haven't given it a name (it's the first term in ALE(x1)^). I don;t think it's a big deal as the paragraph describing the calculation proceeds in a linear fashion so you end up with a "calculated quantity".</li></ul>

jklaise: <p>It does...</p>

jklaise: <p>They do...</p>

examples/ale_regression_boston.ipynb
jklaise: <p>It does for me, added a blank line.</p>

jklaise: <p><code>feature_names</code>  is a numpy array though so it's a bit tricky, I've done this:</p><p> </p><pre class="ql-syntax" spellcheck="false">FEATURE = 'RM'
index = np.where(feature_names==FEATURE)[0][0]


fig, ax = plt.subplots()
ax.scatter(X_train[:, index], lr.predict(X_train));


ax.set_xlabel(FEATURE);
ax.set_ylabel('Value in $1000\'s');
</pre>

jklaise: <p>Rephrased to</p><p>"For linear models this is not an issue as we know the effect is linear across the whole range of the feature, however for non-linear models linear interpolation in feature areas with no data could be unreliable. This is why the use of deciles can help assess in which areas of the feature space the estimated feature effects are more reliable."</p>

jklaise: <p>I like spelling things out, I do kind of want to draw attention to shapes and magnitudes. Rephrased:</p><p> </p><p>"Because the model is no longer linear, the ALE plots are non-linear also and in some cases also non-monotonic. The interpretation of the plots is still the same—the ALE value at a point is the relative feature effect with respect to the mean feature effect, however the non-linear model used shows that the feature effects differ both in shape and magnitude when compared to the linear model."</p>

jklaise: <p>Yea, this is not really studied at all as far as I can tell, but it's a natural idea to compare different models and I wanted to include the plot to show that you can do it very easily. Have been careful about the interpretation, but as you say would have to spend time on studying this in more detail.</p>

examples/ale_classification.ipynb
jklaise: <p>I considered that way the first time round, I ended up going with this because the number of plots is independent of the number of targets. In any case, targets can be switched off to focus on any subset of classes.</p>

jklaise: <p>I changed the order around to be <code>mean_logits - subset_logits</code>  so the interpretation is more clear. Updated text to</p><p> </p><p>"</p><p>Now if we subtract the logits of the instances for which petal length is ~3.8 from the mean logits, because $\text{ALE}(3.8)\approx 0$ for petal length, any difference from zero must be due to the combined effect of all other features (except petal length):</p><p>"</p><p>...code...</p><p>"For example, the remaining 3 features combined must be responsible for a positive effect of around ~0.69 on predicting instances with petal length ~3.8cm to be of class <code>setosa</code>."</p>

jklaise: <ul><li>Seems like British English doesn't use comma after i.e.</li></ul>

jklaise: <p>That's not traditional...</p>

alibi/explainers/ale.py
```diff
@@ -0,0 +1,479 @@
...
+    """
+    q, _ = adaptive_grid(X[:, feature], min_bin_points)
+
+    # find which interval each observation falls into
+    indices = np.searchsorted(q, X[:, feature], side="left")
+    indices[indices == 0] = 1  # put the smallest data point in the first interval
+    interval_n = np.bincount(indices)  # number of points in each interval
+
+    # predictions for the upper and lower ranges of intervals
+    z_low = X.copy()
```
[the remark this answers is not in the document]
jklaise: Did some benchmarks with small-ish data sets (15k rows, ~400 features) and didn't see much difference in total runtime or memory consumption. Will raise an issue to address this, will be more useful when we can support bigger datasets via paralelising.

PR #154, opened 2019-10-08
Improve anchor tabular sampling

alexcoca:
A modified sampling implementation of the AnchorTabular algorithm. 

**Algorithmic summary:**

- The new algorithm draws the samples from the training set as a whole,  from the set of rows that contains the continuous features that fall in the same bins as the instance to be explained
- If the set of the training set rows that contains the aforementioned not have sufficient samples, the algorithm determines the maximum number of continuous features in the anchor s.t. enough samples can be drawn. The values for the remainder continuous features are drawn from the set of values in the training set that come from the same bin, independently
- Categorical variables are subsequently set to the value in the instance to be explained


alibi/utils/discretizer.py
```diff
@@ -1,44 +1,44 @@
...
-from typing import Dict, Callable, List
+from typing import Dict, Callable, Tuple, List
 
 
 class Discretizer(object):
 
-    def __init__(self, data: np.ndarray, categorical_features: List[int], feature_names: List[str],
-                 percentiles: List[int] = [25, 50, 75]) -> None:
+    def __init__(self, data: np.ndarray, numerical_features: List[int], feature_names: List[str],
+                 percentiles: Tuple[int] = (25, 50, 75)) -> None:
```
MarcoGorelli: Couple of comments :)

1. `Tuple[int]` indicates a tuple of length 1, perhaps you meant `Tuple[int, ...]`?
2. Why specify `Tuple` or `List` at all, instead of just `Sequence[int]` which allows for either?

alexcoca: Hi @MarcoGorelli !

Thanks for the comments, very useful!

- Agreed, I'll make that change.
- kwargs should not contain default mutable values (strange things will happen if that gets changed by the function), the intention there was to change that behaviour. This is a situation where I considered using abstract types for typing was not desirable.

MarcoGorelli: Happy to help!

The default value shouldn't be mutable - agreed - but is there any reason to not allow someone to pass in a mutable object when calling the function from somewhere else in the codebase?
```Sequence[int] = (25, 50, 75)```
covers both cases.

Comment (jklaise) [reply alibi/utils/discretizer.py]: I think the Sequence protocol makes sense here, we should probably adopt that convention across the library (there are a few other places with mutable list type defaults atm).

Comment (alexcoca) [reply alibi/utils/discretizer.py]: Good discussion guys, we'll do those changes in a separate PR.

**Comment by arnaudvl** [comment]
Typo in `anchor_tabular` line 85: `Each value in this dict is a set is a set of training data`.

**Comment by arnaudvl** [comment]
`anchor_tabular`: remove unused `cat_feat_ids = []` in `sample_from_train`

**Comment by arnaudvl** [comment]
`anchor_tabular`: remove `# random.shuffle(uniq_feat_ids)`.

**Comment by arnaudvl** [comment]
Separate PR for benchmarking.

**Comment by arnaudvl** [comment]
`anchor_tabular`: remove `# TODO: Discuss this - could replace the entire row?` (x2).

**Comment by alexcoca** [comment]
Comments addressed!

File: alibi/explainers/anchor_tabular.py [hunk 2]
```diff
@@ -48,224 +74,273 @@ def fit(self, train_data: np.ndarray, disc_perc: list = [25, 50, 75]) -> None:
...
+            cat_enc_idx = max(self.ord_lookup.keys()) + 1
+            for cat_feat_idx in range(self.numerical_features[-1] + 1, self.categorical_features[-1] + 1):
+                self.cat_lookup[cat_enc_idx] = X[cat_feat_idx]
+                self.enc2feat_idx[cat_enc_idx] = cat_feat_idx
+                cat_enc_idx += 1
+
+    def sample_from_train(self, anchor: list, val2idx: dict, ord_lookup: dict, cat_lookup: dict, enc2feat_idx: dict,
+                          num_samples: int) -> Tuple[np.ndarray, np.ndarray]:
+        """
+        Sample data from training set but keep features which are anchor in the proposed anchor the same
```
MarcoGorelli: > which are anchor in the proposed anchor

which are present?

alexcoca: Great spot, will update! It made no sense whatsoever before!

alexcoca:
Superseded by #173

PR #160, opened 2019-10-17
Host adult and movie sentiment datasets ourselves

jklaise:
Intended to provide protection against outages of external websites, includes an option to specify a preferred URL.


axsaucedo:
Nice! Thanks @jklaise !

PR #162, opened 2019-11-04
Add flake8 and mypy as pre-commit hooks

jklaise:


adriangonz:
Why has coverage increased?

PR #163, opened 2019-11-04
WIP: Apply black formatter

jklaise:
- [x] Ran `black` on code-base
- [x] Added black as a git pre-commit hook
- [x] Modified `flake8` options to work well with `black`
- [ ] Manual checking and correcting of docstrings etc.


jklaise:
Configured black with our original 120 line length, let's see if this works without ugly line breaks @arnaudvl @alexcoca @gipster @adriangonz .

Issue #155, opened 2019-10-08
Migrate CEM and counterfactual explainer methods to TF2.0


arnaudvl:

RafalSkolasinski:
It would be awesome to see this one done! :)
As alibi-detect depends on `tensorflow>=2.0` this would allow to install both in the same virtualenv saving a lot of headache to many people.

PR #164, opened 2019-11-07
Add testing for Python 3.8

jklaise:


jklaise:
Turns out `tensorflow` isn't supported on Python 3.8, so we will just have to wait and see if they add support for the 1.x branch or once we've refactored to use 2.x #155

PR #168, opened 2019-12-04
WIP: Initial visualisation logic

jklaise:
Depends on #166.

Idea is to try out both `matplotlib` and `altair` visualisations and have a fallback to pretty-printed `plaintext` format as output of a `show(explanation)` command.


jklaise:
Closing as the scope was too broad and some of the proposals in this PR no longer relevant. From now on visualizations will be built for methods which require it (e.g. ALE #152) and will be built in matplotlib to provide flexibility of configuration by the user.

PR #194, opened 2020-03-12
Pin spacy as newest release breaks model downloading

jklaise:


jklaise:
Closing as it's been fixed by https://github.com/explosion/spaCy/issues/5140

PR #202, opened 2020-03-21
Extend fetch imagenet

### LINKED ISSUE #142: Extend fetch_imagenet to handle arbitrary categories
Currently the `fetch_imagenet` data loading utility only has a few hard-coded categories (https://github.com/SeldonIO/alibi/blob/master/alibi/datasets.py#L50), would be good to extend to handle arbitrary categories. This may involve having to pass the category identifier to the function, alternatively, we can look into producing a complete mapping between categories and identifiers (image-net.org doesn't seem to have it) and host it on the alibi repo.

daavoo:
Resolves #142

I added the `mapping` and `label_dict` as plain Python dictionaries under `alibi/utils/imagenet.py`. I though that it was a better solution than depending on requests to external gists. However I also uploaded the dictionaries as [plain jsons](https://gist.github.com/daavoo) in case the old behaviour is prefered.

I also added each of the items present in "complex" identifiers as valid categories. I.e. you could use '`Siamese cat`' and '`Siamese`' as queries in addition to the original identifier '`Siamese cat, Siamese`'


jklaise:
Hi @daavoo , great work, thanks for your contribution! On our end, we just need to decide if it's better to host the mappings as Python native objects within the library or keep them as jsons (either inside or outside the library) and then we can merge this.

jklaise:
OK, we think the best solution would be to include the 2 json files in the library (e.g. under `alibi/data`, would need to make a `MANIFEST.in` too to ensure they're packaged with the library), and then deserialize to a dict when calling `fetch_imagenet`. This would avoid creating an extra python module which just hosts data. I'm happy to make these changes on top of your PR if that sounds good?

daavoo:
Sounds great.

PR #205, opened 2020-04-06
WIP: Integrated gradients 2

gipster:


jklaise:
Looks like eager mode doesn't mesh well with the existing tests which fetch the session... Might need to check if eager mode can be selected when initializing the component (as opposed to module level), but then not sure how to disable it afterwards. v1 could become a blocker for this to get merged @alexcoca @arnaudvl .

arnaudvl:
**integrated_gradients.py**
- [ ] I feel that attributes such as *n_steps*, *method*, *return_convergence_delta* and *return_predictions* should be part of explain call and should not be class attributes.
- [ ] *forward_function* parameter; rename to *model* since the only allowed options are *tf.keras* or *keras* models anyway?
- [ ] assert statements on lines 260 and 380 should in my opinion not be there but only in the actual tests b/c they are not checks on inputs defined by users but checks on internal functioning of a function which is not exposed to the user anyway. They also test the same thing.
- [ ] assert statement to raise an error on line 46 should be changed into actual error raising
- [ ] need to check whether it is ok to have experimental TensorFlow functionality in the code (*tf.data.experimental.AUTOTUNE*) since this might eventually not be properly supported and subject to breaking change
- [ ] order imports
- [ ] new line for first arg after *self* of class init
- [ ] various typos
- [ ] *feature_names* instead of *features_names*, same in explain dict which is returned
- [ ] *baselines* parameter docstring says that it must have the same shape as X but this is not the case as it can be an *int* or *float*.
- [ ] I believe it would be good to internally support the use of model class predictions as targets as well so the user doesn't have to explicitly specify it. This would also allow *target=None*.
- [ ] use update to update explain dict to be consistent with other methods in alibi
- [ ] dtypes of *bls* need to match *X* in **_format_input_baseline**
- [ ] no need to return X in **_format_input_baseline**
- [ ] 106 in **_run_forward**: allow for forward function to return shape *(n,)* as there would be no shape[1]?

**approximation_methods.py**
- [ ] reference to original code
- [ ] change style to match our template

**alibi/api/defaults.py**
- [ ] "global" in metadata but the implemented method is only "local"

**example notebooks**
- [ ] add eager execution by default
- [ ] clean up
- [ ] extend with imagenet example
- [ ] visualize text attributions

**General**
- [ ] documentation
- [ ] For text: default background = 0's. Does this consistently correspond to e.g. padding or another neutral token across tokenizers? This needs to be clearly documented and highlighted in the example notebooks.
- [ ] To make it more user friendly, the text attributions format should probably be in terms of tokens and not in the full embedding space.

jklaise:
@arnaudvl @gipster on the two points above:

- Anything in the returned explanation should be easily serializable, I agree that it would be great to record the `layer` explained in the `params`, but if the API supports taking a layer as an object this is not straightforward. What if instead we passed an `int` pointing to the layer instead of the actual layer? Also, does IG work with models that are not sequential (if it does, then I can see how passing an `int` would reduce the capabilities)
- On the one hand if the parameter is not used in `__init__` then it makes sense to pass it in `explain`.  OTOH, @alexcoca made a point that most things the user wouldn't change between `explain` calls should be in `__init__` with defaults so that the method signature of `explain` remains uncluttered (but you could still override these via `kwargs.get()` in `explain`).
I agree that if `internal_batch_size` is in `explain` so should `n_steps`. The `method` is more tricky, could make sense in either case. Basically, the method looks like it would support a simple funcitonal API (explanations delivered via a function call without extensive setup of the object), it's up to us to decide the semantics of `__init__` vs `explain`. I'm leaning towards putting most params in `__init__`.

@gipster I unticked 2 issues related to `feature_names` as it's still under `explain` and also still spelled `features_names`.

jklaise:
A few other things going through the examples, I can patch these up once you're finished @gipster :

- No need to iterate/format metadata/data fields in the image notebooks, can just call `explanations.meta` and `explanation.data.keys()`
- The notebooks have for some reason set a `kernelspec` that's specific to you environment which results in a pop-up for anyone looking at the notebook to select a kernel, the entries in the json should be replaced to be generic as with existing notebooks

arnaudvl:
> @arnaudvl @gipster on the two points above:
> 
> * Anything in the returned explanation should be easily serializable, I agree that it would be great to record the `layer` explained in the `params`, but if the API supports taking a layer as an object this is not straightforward. What if instead we passed an `int` pointing to the layer instead of the actual layer? Also, does IG work with models that are not sequential (if it does, then I can see how passing an `int` would reduce the capabilities)
> * On the one hand if the parameter is not used in `__init__` then it makes sense to pass it in `explain`.  OTOH, @alexcoca made a point that most things the user wouldn't change between `explain` calls should be in `__init__` with defaults so that the method signature of `explain` remains uncluttered (but you could still override these via `kwargs.get()` in `explain`).
>   I agree that if `internal_batch_size` is in `explain` so should `n_steps`. The `method` is more tricky, could make sense in either case. Basically, the method looks like it would support a simple funcitonal API (explanations delivered via a function call without extensive setup of the object), it's up to us to decide the semantics of `__init__` vs `explain`. I'm leaning towards putting most params in `__init__`.
> 
> @gipster I unticked 2 issues related to `feature_names` as it's still under `explain` and also still spelled `features_names`.

Wrt using an `int` to point to the layer. There are some subtleties that need to be considered. For the class-modular approach in TF2, the model has not necessarily been built/compiled yet. So in order to point to an layer I believe it needs to be built first. This can be done via:
```python 
inputs = Input(shape=input_shape)
model.call(inputs)
```
Check a similar approach in [alibi-detect](https://github.com/SeldonIO/alibi-detect/blob/master/alibi_detect/cd/preprocess.py) for more detail.

jklaise:
@arnaudvl I looked at passing in desired layer as an `int`, I think to support the non-compiled case would require significant changes to how the gradients are currently calculated internally in `IG`, on the other hand I think that replacing the `layer` with an `int` and then doing `self.layer=model.layers[layer]` would cover the current functionality as is.

PR #214, opened 2020-05-11
Allow a list of strings of column names for dataframes in gen_category_map

### LINKED ISSUE #213: Issue with category mapper
When passing a list of column names into the ```gen_category_map``` function from [utils.data.py](https://github.com/SeldonIO/alibi/blob/master/alibi/utils/data.py), I get a value error, for using a string on an iloc slicer.

Here is the relevant code.
```python
# create the map
    category_map = dict.fromkeys(categorical_columns)
    for col in categorical_columns:
        le = LabelEncoder()
        try:
            _ = le.fit_transform(data.iloc[:, col])
        except (AttributeError, IndexError):
            raise

        category_map[col] = list(le.classes_)

    return category_map
```

I'm not sure if ```categorical_columns``` is supposed mapped to an index location, but it seems like maybe you meant to use loc? Or if this isn't right, then maybe put into type hint for ```categorical_columns``` for guidance?

```python
# create the map
    category_map = dict.fromkeys(categorical_columns)
    for col in categorical_columns:
        le = LabelEncoder()
        try:
            _ = le.fit_transform(data.loc[:, col])
        except (AttributeError, IndexError):
            raise

        category_map[col] = list(le.classes_)

    return category_map
```

jklaise:
Resolves #213.

- Can specify `categorical_columns` as `List[str]` if `data` is `pd.DataFrame`
- Simplified test, now checking for equality of `categorical_map`
- Note that even if `categorical_columns` is `List[str]` the output is always `Dict[int, list]`, the integer indices are got by calling `data.columns.get_loc(col)`.

### DIFF
--- a/alibi/utils/data.py
+++ b/alibi/utils/data.py
@@ -1,7 +1,8 @@
 import numpy as np
 import pandas as pd
 from sklearn.preprocessing import LabelEncoder
-from typing import Dict, Union
+from typing import Dict, List, Union
+
 
 # TODO: This should inherit from collections.UserDict not dict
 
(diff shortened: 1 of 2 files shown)


jklaise:
Removed last commit which patched the `scikit-image` dependent failing test for this PR to pass.

PR #218, opened 2020-05-12
Change old keras imports to tensorflow.keras imports

### LINKED ISSUE #216: Using old keras in a test
Found this line which uses `keras` and hence imports from this test file fails. Whilst not a big deal as we currently have `keras` in the dev requirements, this should still be replaced by the `tf` version:

https://github.com/SeldonIO/alibi/blob/98c61ca6cecfaf7e7dff327be9c535eac4bccb72/alibi/explainers/tests/utils.py#L7

jklaise:
Resolves #216.

Ran a pattern search through the codebase and discovered 3 places where we used imports form old keras instead of tf.keras for functionality testing.


alexcoca:
Let's check the linting errors before merging though.

jklaise:
Yea, I need to clean up the way we do requirements as currently you need to change things in multiple places when a dependency is required to have a certain range.

PR #238, opened 2020-06-02
Proposed Counterfactual Instances Implementation

alexcoca:
Hi guys,

Please find attached a proposed implementation for the counterfactual in TF2. The main goals for this were:

- Create a TF2 implementation for counterfactual
- The implementation should be modular, and few (and simple!) methods should be TensorFlow specific. Most of this code should be transferable to a base class that is framework agnostic
- Implement additional functionality such as feature scaling
- Make initial lambda optimisation optional (NB: this is 90% implemented, thought we'd discuss user experinence first)
- Improve tensorboard logging 

Things to note:
- For now, I have removed all the batch related code. I think there are some better ways to manage it. 
- I put all the stuff in one file so you don't have to trawl through files too much. I suggest you start reading at `L114`. 
- I have marked points I want to discuss with the tag `TODO: ALEX: TBD` - ignore other `TODOs` 

Hopefully, following the read we can:

- discuss the points marked 
- discuss how other "flavours" of counterfactual explanations (e.g., black-box model support, PyTorch support) could benefit from the current work (hopefully this should be clear from the way the code is written)
- discuss the future direction of this code

### DIFF
--- a/alibi/explainers/counterfactual_tf2.py
+++ b/alibi/explainers/counterfactual_tf2.py
@@ -1,0 +1,1028 @@
+import copy
+import logging
+
+import numpy as np
+import tensorflow as tf
+
+from alibi.api.interfaces import Explainer, Explanation
+from alibi.api.defaults import DEFAULT_META_CF, DEFAULT_DATA_CF
+from alibi.utils.wrappers import methdispatch
+from copy import deepcopy
+from functools import partial, singledispatch
+from tensorflow.keras.optimizers import Adam
+from tensorflow.keras.optimizers.schedules import PolynomialDecay
+from typing import Callable, Dict, List, Optional, Tuple, Union, TYPE_CHECKING
+
+
+if TYPE_CHECKING:  # pragma: no cover
+    import keras  # noqa
+
+logger = logging.getLogger(__name__)
+
+
+# TODO: ALEX: TBD: Move the casting generic function to an appropriate utility file.
+# TODO: ALEX: This is a simple draft to get the CF code to work. More generally, we would
+#  define an object/series of functions to check if PyTorch/TF/both are installed and return an
+#  appropriate casting function.
+
+@singledispatch
+def to_numpy_arr(X: Union[tf.Tensor, tf.Variable, np.ndarray]):
+
+    supported_types = [tf.Tensor, tf.Variable, np.ndarray]
+    raise TypeError(f"Expected input object type to be one of {supported_types} but got type {type(X)}!")
+
+
+@to_numpy_arr.register
+def _(X: np.ndarray) -> np.ndarray:
+    return X
+
+
+@to_numpy_arr.register
+def _(X: tf.Tensor) -> np.ndarray:
+    return X.numpy()
+
+
+@to_numpy_arr.register
+def _(X: tf.Variable) -> np.ndarray:
+    return X.numpy()
+
+
+# TODO: ALEX: TBD I think it's best practice to define all the custom
+#  errors in one file (e.g., exceptions.py/error.py). We
+#  could standardise this across the library
+
+
+class CounterfactualError(Exception):
+    pass
+
+
+CF_SUPPORTED_DISTANCE_FUNCTIONS = ['l1']
+CF_VALID_SCALING = ['MAD']
+CF_PARAMS = ['scale_loss', 'constrain_features', 'feature_whitelist']
+
+
+def counterfactual_loss(instance: tf.Tensor,
+                        cf: tf.Variable,
+                        lam: float,
+                        feature_scale: tf.Tensor,
+                        pred_probas: Optional[tf.Variable] = None,
+                        target_probas: Optional[tf.Tensor] = None,
+                        distance_fcn: str ='l1',
+                        ) -> tf.Tensor:
+
+    # TODO: ALEX: TBD: TO SUPPORT BLACK-BOX, PYTORCH WE NEED A SIMILAR FUNCTION AND A HIGHER LEVEL ROUTINE TO
+    #  "DISPATCH" AMONG THESE FUNCTIONS
+
+    # TODO: ALEX: DOCSTRING
+
+    distance = lam * distance_loss(instance, cf, feature_scale, distance_fcn=distance_fcn)
+    if pred_probas is None or target_probas is None:
+        return distance
+    pred = pred_loss(pred_probas, target_probas)
+    return distance + pred
+
+
+def distance_loss(instance: tf.Tensor,
+                  cf: tf.Variable,
+                  feature_scale: tf.Tensor,
+                  distance_fcn: str = 'l1') -> tf.Tensor:
+
+    # TODO: ALEX: DOCSTRING
+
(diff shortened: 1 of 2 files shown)


alibi/explainers/counterfactual_tf2.py
```diff
@@ -0,0 +1,870 @@
...
+        prefix = kwargs.get('prefix', '')
+        found = kwargs.get('found', 0)
+        not_found = kwargs.get('not_found', 0)
+        lr = self._get_learning_rate()
+
+        # TODO: ALEX: TBD: For image data, it would be nice if we displayed the counterfactuals as well?
+        # TODO: ALEX: TBD: I don't quite like how we take a random set of arguments here. Would be nice to have
+        #  a cleaner interface something that maybe takes step as pos, a dict with other things you might want to
+        #  pass it but mostly takes what is in a dictionary that saves the algo state (in this case, `instance_dict`)
+        #  and outputs that. Might be more hassle than its worth.
```
jklaise, line 717: I see your point, let's discuss.

alibi/explainers/counterfactual_tf2.py
```diff
@@ -0,0 +1,870 @@
...
+        TensorFlow event file for visualisation purposes.
+        """
+
+        # TODO: ALEX: DOCSTRING
+
+        instance = self.instance_numpy
+
+        # compute loss terms for current counterfactual
+        # TODO: ALEX: TBD: Depending on how we implement the loss of the black-box, we could leverage that here
+        #  as opposed to hardcoding the calculation
```
jklaise, line 677: Not sure I follow, let's discuss.

alibi/explainers/counterfactual_tf2.py
```diff
@@ -0,0 +1,885 @@
...
+
+    return tf.reduce_sum(tf.abs(cf - instance) / feature_scale, axis=ax_sum, name='l1')
+
+
+def pred_loss(pred_probas: tf.Variable, target_probas: tf.Tensor) -> tf.Tensor:
+
+    # TODO: ALEX: DOCSTRING
+
+    return tf.square(pred_probas - target_probas)
+
```
arnaudvl, line 103: I think that often improvements to the counterfactual method for specific problems will be made by modifying the loss functions (distance/pred) slightly to be more suitable for the problem at hand. Same goes for improvements of the CF method itself. 2 key parts are the optimisation function and process. Changes to the optimisation process (e.g. using a genetic algorithm) would usually involve bigger code changes anyway, but changes to the objective function should be very straightforward. Allowing users to pass custom loss functions for the distance and pred loss terms would be very convenient. These functions might need a specific signature but I think that's fine as users can always pass partially parameterised functions.

alibi/explainers/counterfactual_tf2.py
```diff
@@ -0,0 +1,870 @@
...
+        save_every
+            Controls the frequency at which data is written to Tensorboard.
+        kwargs
+            Keyword arguments can be passed to initialise the optimiser state and search parameters. The
+            parameters that are passed are listed in the documentation for `explain`. The user is encouraged
+            to pick adequate defaults for their problem and adjust the parameters if the search procedure fails
+            to converge.
+        """  # noqa W605
+
+        # TODO: ALEX: TBD: I think this
```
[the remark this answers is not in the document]
arnaudvl: Never seen a cliffhanger of this magnitude in a class init

alibi/explainers/counterfactual_tf2.py
```diff
@@ -0,0 +1,1028 @@
...
+
+def range_constraint(X: Union[tf.Variable, tf.Tensor], low: Union[float, np.ndarray], high: Union[float, np.ndarray]):
+
+    # TODO: ALEX: TBD: SHOULD WE DO ERROR HANDLING FOR THIS FUNCTION (E.G., LOW < HIGH)
+
+    # TODO: ALEX: ADD DOCSTRING,
+    # TODO: ALEX: CROSS REFERENCE INIT DOC FOR MIN AND MAX, MAKE SURE THEY DEFINE SHAPES FOR MIN/MAX
+
+    return tf.clip_by_value(X, clip_value_min=low, clip_value_max=high)
+
```
arnaudvl, line 115: To allow easy constraint of specific parameter values (e.g. can't change the age of a person but it's possible to change any other feature) clipping fn should be done in a user-friendly way by feature.

alibi/explainers/counterfactual_tf2.py
```diff
@@ -0,0 +1,870 @@
...
+                       lam: float,
+                       lam_lb: float,
+                       lam_ub: float,
+                       lam_cf_threshold: int = 5) -> Tuple[float, float, float]:
+        """
+        Runs a bisection algorithm on lambda in order to find a better counterfactual ( # TODO: ADD DETAILS)
+        """
+
+        # TODO: ALEX: DOCSTRING
+        # TODO: ALEX: TBD: SHOULD WE EXPOSE lam_cf_threshold as parameter to user?
```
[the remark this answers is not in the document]
arnaudvl: Same goes then for the scaling of lambda (/10).

alibi/explainers/counterfactual_tf2.py
```diff
@@ -0,0 +1,1028 @@
...
+        return self.build_explanation(X, result, instance_class, instance_proba)
+
+    def make_prediction(self, X: Union[np.ndarray, tf.Variable, tf.Tensor]) -> tf.Tensor:
+        """
+        Makes a prediction for data points in `X`
+        """
+        # TODO: ALEX: DOCSTRING
+
+        return self.predictor(X, training=False)
+
```
arnaudvl, line 439: Don't think you can make `training=False` keyword assumption in prediction signature.

alibi/explainers/counterfactual_tf2.py
```diff
@@ -0,0 +1,1028 @@
...
+        """
+
+        # TODO: ALEX: TBD: Parametrise `threhold` in explain or do we assume people always assign labels on this rule?
+        # TODO: ALEX: ADD DOCSTRING
+
+        if Y.shape[1] == 1:
+            return int(Y > threshold)
+        else:
+            return np.argmax(Y)
+
```
arnaudvl, line 454: Might be missing edge case where `Y` has shape `(n,)` for custom models.

alibi/explainers/counterfactual_tf2.py
```diff
@@ -0,0 +1,1028 @@
...
+                instance=self.instance,
+                cf=self.cf,
+                lam=lam,
+                pred_probas=prediction,
+                target_probas=self.target_proba,
+            )
+        gradients = tape.gradient(loss, [self.cf])
+
+        return gradients
+
```
arnaudvl, line 736: Not urgent but need to keep in mind that potentially numerical gradients could be added as well.

alibi/explainers/counterfactual_tf2.py
alexcoca: So if you give a tensor of dimension [B, N] to a predictor in TF, do you not get [B, X] where X is output dimension? The docs clearly state there should be a leading singleton dimension and an error is raised if that's not the case.

PR #249, opened 2020-06-23
Make TensorFlow an optional dependency

jklaise:
First attempt at removing TF strict dependency, to be discussed.

New module `alibi.utils.imports.py` defines `has_tensorflow`, `tf_version`, `tf_required` and `tf_upgrade` variables. A function `assert_tensorflow_installed()` raises an error if TF is missing or issues a warning that functionality may not work if TF2.x is installed and a warning that in the future TF2.x will be required if TF1.x is installed. `assert_tensorflow_installed` is meant to be called whenever user instantiates a class that requires TensorFlow (so currently `CEM`, `CounterFactual`, `CounterFactualProto` and `IntegratedGradients`).

Checked that everything works as expected in environments with and without TF.

The main issue is the multiple imports of TF within the bodies of functions. Whilst an import at `__init__` of a class is no big deal, due do the modularity of some current (e.g. `integrated_gradients`) and upcoming modules the import needs to be called in every function and method that uses TF. It is unclear what kind of performance hit this would result in. I would expect that on subsequent import calls this would be cheap.

Points to discuss:

- Is this a scalable way of managing optional dependencies (in terms on increasing number of modules/functions, also with `torch` being treated the same way in the future)
- Alternatives - e.g. separate packages for `tensorflow` and `torch` building blocks was proposed, but I'm skeptical that this would alleviate the issue as we would now have to manage multiple packages and it's unclear how we could e.g. write a generic `Counterfactual` implementation that works with either `torch` or `tensorflow` without doing similar import checks and not duplicating work across the proposed packages.


jklaise:
Closing as will need rethinking.

PR #254, opened 2020-06-26
Update expected value after explain call

### LINKED ISSUE #253: Incorrect expected_value from alibi.explainers.TreeShap.explain function
Expected_value of shap.TreeExplainer.expected_value not matching with the one provided from alibi.explainers.TreeShap.explain
Have used xgboost multiclass classification model for generating the shap values on a modified IRIS dataset.

**At stage** **explainer.fit()**
The shap.TreeExplainer returns a value (in my case [0.41060104 0.39954053 0.37059403]) as the “expected_value” which is directly assigned to “self.expected_value” inside fit function of alibi.explainers.TreeShap (_line number 1109 at https://github.com/SeldonIO/alibi/blob/master/alibi/explainers/shap_wrappers.py_)

**At Stage** **explainer.explain(df)**
The shap.TreeExplainer refreshes the expected value with the bias value of phi (which is [-0.100694835, 0.5807366, 0.700694] in my case). But in explain function of shap_wrappers.py (_line 1249 at https://github.com/SeldonIO/alibi/blob/master/alibi/explainers/shap_wrappers.py_) the value of expected_value is assigned with previously set value and not refreshed with the new calculated value of shap.TreeExplainer expected_value

jklaise:
Fixes #253 .

Also updated `KernelShap` for consistency even though the current implementation in `shap.kernel` doesn't change the `expected_value` after the call to `shap_values`.

Initially was going to remove `expected_value` as an attribute, then realized we use it in tests. Was rewriting tests when realized we also use it in examples, so in the end kept it as an attribute that is present after `fit`.

### DIFF
--- a/alibi/explainers/shap_wrappers.py
+++ b/alibi/explainers/shap_wrappers.py
@@ -770,12 +770,13 @@ class KernelShap(Explainer, FitMixin):
             X = X.toarray()
 
         shap_values = self._explainer.shap_values(X, **kwargs)
+        self.expected_value = self._explainer.expected_value
         expected_value = self.expected_value
         # for scalar model outputs a single numpy array is returned
         if isinstance(shap_values, np.ndarray):
             shap_values = [shap_values]
         if isinstance(expected_value, float):
-            expected_value = [self.expected_value]
+            expected_value = [expected_value]
 
         explanation = self.build_explanation(
             X,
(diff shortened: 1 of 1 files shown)


alexcoca:
> 
> 
> Fixes #253 .
> 
> Also updated `KernelShap` for consistency even though the current implementation in `shap.kernel` doesn't change the `expected_value` after the call to `shap_values`.
> 
> Initially was going to remove `expected_value` as an attribute, then realized we use it in tests. Was rewriting tests when realized we also use it in examples, so in the end kept it as an attribute that is present after `fit`.

We need that attribute because people need the expected value and they should not care about retrieving it from objects they aren't supposed to interact with (e.g., `_explainer`).

PR #255, opened 2020-06-26
Enable correct AnchorText behaviour with spacy>=2.3.0

### LINKED ISSUE #251: AnchorText tests fail with spacy=2.3.0
New version of spacy made our AnchorText tests fail, an example failed build is here: https://travis-ci.com/github/SeldonIO/alibi/jobs/352674078

jklaise:
Fixes #251 . This commit makes the code functional with either spacy version but does no attempts at improving runtimes or trying to conditionally navigate the need for the extra lookup package dependency.

Spacy 2.3.0 moved probability tables outside into a separate package and
the way to access these was slightly changed. This commit enables one to
use either spacy version but crucially relies on installing
spacy_lookups_data even though this is only necessary for spacy>=2.3.0.
Importantly, this data package includes a lot of data files we don't
care about and may grow in the future. We should investigate if there's
a more sensible way of getting the probability scores.


jklaise:
Thanks. I will rename #255 to focus on performance issues for calculating similarities.

Issue #242, opened 2020-06-09
Testing different TensorFlow model versions


jklaise:
Currently all our test that use TF model just train a simple model during test runtime with whatever TF version is available. This is not robust to changes in TF versions.

We should use pre-trained models with different TF versions (to start with, TF1.15, TF2.x and old Keras 2.3.1) which would allow us to check the library compatibility with older models.

We can store the pre-trained models on a bucket and fetch them at test-time, the challenges would be to figure out the right fixtures and parametrization to enable succinct testing across different model versions whilst ensuring models are downloaded and cached once.

jklaise:
Closed by #260, white-box explainers are now tested by pre-trained TF1.x and TF2.x models hosted at https://github.com/SeldonIO/alibi-testing .

PR #260, opened 2020-06-30
Switch to TensorFlow 2.x

jklaise:
This PR adds the minimal changes required to switch from TF1.x to TF2.x mainly by importing `import tensorflow.compat.v1 as tf` in modules and using `tf.disable_v2_behavior()` in example notebooks. Some adjustments have been made to the tests as well.

One potentially contentious point is the continued support of old `Keras` models. I had to get rid of `keras.backend.get_session()` call in `_check_keras_or_tf` because it simply doesn't work with TF2.x. Whilst the tests currently pass, it is not clear whether full support for both TF2.x models and Keras models using TF2.x is there due to potential interactions with global state that could come with disabling v2 behaviour. Furthermore, it is unclear if TF1.x will work out-of-the-box, although I suspect they should. Follow up work to scope and test various combos of TF and Keras versions should be done as part of #242 .


jklaise:
We should check if TF2.x behaviour is disabled before initializing these components, if not then raise an error and ask the user to disable it. The reason for not doing this inside the library is that it would mess up eager components and it's not possible to enable eager mode in the middle of execution so the user would have to restart the Python kernel.

ssameerr:
Hi,

I would like know know if this branch is pushed to pip already? I am running TF2+ and would be helpful to know any roadmap of integrating TF2 in future.

ERROR: Could not find a version that satisfies the requirement tensorflow<2.0 (from alibi) (from versions: 2.2.0rc1, 2.2.0rc2, 2.2.0rc3, 2.2.0rc4, 2.2.0, 2.3.0rc0, 2.3.0rc1)
ERROR: No matching distribution found for tensorflow<2.0 (from alibi)

Thanks
Sameer

jklaise:
This branch is currently only a proof of concept and in fact will likely be superseeded by #266 to make both TensorFlow and PyTorch optional plugins.

Currently TensorFlow is really only used for `CounterFactual, CEM, CFProto, IntegratedGradients` explainers. If you are only using other components then the version of TF doesn't really matter, although it needs to be present to be able to import the library.

jklaise:
Tests which use trained Tensorflow models now use pre-trained models from `alibi-test-models` helper package which contains model binaries for TF1.x and TF2.x.

Additionally, `alibi-test-models` contains a few data loading functions which are used in some tests. TBD: there is some overlap with the functionality in `alibi.explainers.tests.utils`.

alibi/explainers/tests/utils.py
```diff
@@ -96,6 +96,25 @@ def fashion_mnist_dataset():
     }
 
 
+def mnist_dataset():
```
alexcoca, line 99: should these functions move to the `alibi_models` repo? When we load the data, we want to ensure that it is compatible with the model trained.

alibi/explainers/tests/test_cfproto.py
```diff
@@ -94,23 +25,30 @@ def tf_keras_iris_explainer(request, tf_keras_iris):
...
-    ((True, 0., 1), True, 2),
-    ((True, 1., 3), True, 2)
-], indirect=['tf_keras_iris_explainer'])
-@pytest.mark.parametrize('tf_keras_iris_model,tf_keras_iris_ae', [('tf', 'tf'), ('keras', 'keras')],
+    yield model, cf_explainer
+    keras.backend.clear_session()
+    tf.keras.backend.clear_session()
+
+
+# TODO: old Keras model missing
```
alexcoca, line 33: Are we going to support this?

alibi/explainers/tests/test_cfproto.py
```diff
@@ -1,89 +1,20 @@
...
-
-    return X_train, model, ae, enc
+from alibi_test_models.data import adult_data, iris_data
 
 
 @pytest.fixture
-def tf_keras_iris_explainer(request, tf_keras_iris):
-    X_train, model, ae, enc = tf_keras_iris
-
+def tf_keras_iris_explainer(request, models, get_iris_dataset):
```
alexcoca, line 15: I'm not sure I follow where `get_iris_dataset` is used here? It seems that `iris_data()` returns the data instead?

alibi/explainers/tests/conftest.py
```diff
@@ -8,18 +8,46 @@
...
+@pytest.fixture
+def models(request):
+    """
+    This fixture loads multiple pre-trained test-models by name from the
+    alibi-test-models helper package.
+    """
+    models = []
+    for name in request.param:
+        models.append(alibi_test_models.load(name))
+    return models
```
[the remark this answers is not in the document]
jklaise: Possibly, but that would make the single-model fixture dependency slightly annoying (having to index into [0]). Another alternative is to validate request and return a list if it a list, otherwise a model if it is a string.

alibi/explainers/tests/test_anchor_image.py
```diff
@@ -4,18 +4,19 @@
 import numpy as np
 from alibi.api.defaults import DEFAULT_META_ANCHOR, DEFAULT_DATA_ANCHOR_IMG
 from alibi.explainers import AnchorImage
-from alibi.explainers.tests.utils import fashion_mnist_dataset
-
+from alibi.explainers.tests.utils import mnist_dataset
 
 # Data preparation
-data = fashion_mnist_dataset()
+data = mnist_dataset()
```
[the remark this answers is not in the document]
jklaise: Because the pre-trained model used is on MNIST.

alibi/explainers/tests/test_cfproto.py
```diff
@@ -212,29 +101,35 @@ def tf_keras_adult(tf_keras_adult_model):
...
 
-    return X_train, model, cat_vars_ohe
+    return cat_vars_ohe
 
 
 @pytest.fixture
-def tf_keras_adult_explainer(request, tf_keras_adult):
-    X_train, model, cat_vars_ohe = tf_keras_adult
-
+def tf_keras_adult_explainer(request, model, adult_cat_vars_ohe):
```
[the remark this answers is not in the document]
jklaise: Yep, could be either way.

alibi/explainers/tests/test_cfproto.py
```diff
@@ -212,29 +101,35 @@ def tf_keras_adult(tf_keras_adult_model):
...
-], indirect=['tf_keras_adult_explainer'])
-@pytest.mark.parametrize('tf_keras_adult_model', ['tf', 'keras'], indirect=True)
-def test_tf_keras_adult_explainer(tf_keras_adult_explainer, use_kdtree, k, d_type):
-    X_train, model, cf = tf_keras_adult_explainer
+    yield model, cf_explainer
+    keras.backend.clear_session()
+    tf.keras.backend.clear_session()
+
+
+# TODO: old Keras model missing
```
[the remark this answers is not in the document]
jklaise: No

alibi/utils/tf.py
```diff
@@ -32,7 +32,8 @@ def _check_keras_or_tf(predict_fn: Union[Callable, tf.keras.Model, 'keras.Model'
         import keras  # noqa
         sys.stderr = stderr
 
-        ksess = keras.backend.get_session()
+        # ksess = keras.backend.get_session() # had to remove this because this doesn't exist with TF2.x (error raised)
+        ksess = tf.keras.backend.get_session()  # TODO: does this actually get the old Keras session?
```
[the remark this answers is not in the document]
jklaise: The answer is that I don't know, the tests seemed to run in a previous iteration with old Keras models, but we're not planning to support this so this should not be a concern anymore.

alibi/explainers/tests/utils.py
jklaise: Possibly, although I want to avoid too much fragmentation of test utility code between the two repos. We need to think about this a bit more.

alibi/explainers/tests/test_cfproto.py
jklaise: No

alibi/explainers/tests/test_cfproto.py
jklaise: Well spotted

PR #262, opened 2020-07-03
Pin spacy version to be <2.3.0

jklaise:
The newest spacy version does not function well with inferring POS tags
for single lexemes leading to no similar words found and crashing the
  AnchorText algorithm. Putting an upper bound on the spacy version
  until we find a suitable work-around.


jklaise:
Closing as the issue was to do with a bug in the code base when tokens with no similar ones (e.g. `,`) were present.

PR #263, opened 2020-07-03
Bug fixes:

alexcoca:
- Handle empty arrays in set_data_type
- Remove unused kwargs in perturb_sentence
- Rename properties to avoid confusion between objects

### DIFF
--- a/alibi/explainers/anchor_text.py
+++ b/alibi/explainers/anchor_text.py
@@ -206,7 +206,7 @@ class AnchorText(Explainer):
 
         return self.predictor(samples.tolist()) == self.instance_label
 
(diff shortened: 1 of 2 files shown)


jklaise:
The issue was using the defunc `use_proba` instead of `use_similarity_proba` in the tests. A bit worrying as the tests where this should be True never ran as a result.

jklaise:
Added another example. I opted for removing a few helper functions for string processing used to assert the correctness of the processed tokens and opted for hard-coded numbers instead. Because the helper functions process the strings in a different way than the explainer it doesn't always make sense for these two to coincide (and in fact they don't) making testing unreliable.

PR #270, opened 2020-07-21
Prototype implementation for distributed KernelSHAP

alexcoca:
The purpose of this PR is to prototype integration of distributed explanation for the KernelSHAP algorithm in Alibi.

**Aims:**
- Speed up computations for KernelSHAP
- Provide users with a simple interface for using the distributed implementation 
- Ideally, make the distributed backend re-usable 

**Summary of changes:**

1. `KernelExplainerWrapper`
- Implements a wrapper for `shap.KernelExplainer` that allows calling the explainer with a `(idx, batch)` input in a distributed context
- The wrapper allows the main process to read state that is common across the worker process through the `_return_attribute` method

2. `DistributedExplainer`
- This class creates a pool of processes to which batches of explanations can be submitted, using ray. It needs to know the explainer, and the `__init__` args and kwargs to do so
- Splits an input array into batches and submits the batches to the pool (which returns results as they are finished)
- Collect the results from the pool, re-order and concatenate them 
- Tries to be model agnostic, so the pool target function and the concatenation are separate functions, implemented in the global scope 

**Pros:**
- At least 2x speedup on Adult dataset (with equal sized batches)
- Easy to use, no massive API changes
- Should be able to reuse and distribute other algorithms

**Cons:**
- Implementation entangled in the wrapper. Preferred this design because:
       - using a "n_jobs" argument is familiar to the community (e.g., from sklearn, xgboost)
        - Avoided overcomplicating the object hierarchy or getting the user to instantiate a separate object if they want distributed capability. 
- Currently results are returned as they finished but the preprocessing waits until everyone is finished before concatenation. I did not overcomplicate this, because the postprocessing carries v little overhead, but in theory we could combine the results as they finish and not wait.


alibi/explainers/shap_wrappers.py
```diff
@@ -207,6 +209,57 @@ def _reduction(arr, axis, indices=None):
...
+    """
+    A wrapper around `shap.KernelExplainer` that supports:
+
+        - fixing the seed when instantiating the KernelExplainer in a separate process
+        - passing a batch index to the explainer so that a parallel explainer pool can return batches in arbitrary order
+    """
+    def __init__(self, *args, **kwargs):
+        seed = None
+        if 'seed' in kwargs:
+            seed = kwargs.pop('seed')
```
jklaise, line 228: Why not simply `seed=kwargs.get('seed', None)` ?

alibi/explainers/shap_wrappers.py
```diff
@@ -255,6 +309,30 @@ def __init__(self,
...
+                
+            The distributed explanation works only the `ray`_ library is installed. 
+            
+            .. _ray:
+               https://docs.ray.io/en/master/
+            
+        Raises
+        ------
+        ModuleNotFoundError
+            If the `ray` library is not installed.
```
jklaise, line 334: ... and `n_cpus` is set in `distributed_opts`.

alibi/utils/distributed.py
```diff
@@ -217,3 +224,203 @@ def _chunk(values, chunksize):
...
+    """
+
+    s = np.empty_like(p)
+    s[p] = np.arange(len(p))
+    return s
+
+
+class DistributedExplainer:
+    """
+    A class that orchestrates the execution of the execution of a batch of explanations in parallel.
```
jklaise, line 305: repeated typo

alibi/explainers/shap_wrappers.py
alexcoca: `seed` is part of the kwargs only in a distributed context, and it is used to control the randomness inside the `shap` library from every process. Two things I overlooked:

- Does setting the seed before invoking the distributed context "carry over" to each process (eg, is the call inside `fit` sufficient to set the random stream across all processes?) 
- L229 has an undesired side-effect of resetting the seed set in `fit` (breaks non-distributed version) - to be fixed.

alibi/explainers/shap_wrappers.py
alexcoca: Good shout ...

alibi/utils/distributed.py
```diff
@@ -217,3 +224,203 @@ def _chunk(values, chunksize):
...
+    """
+    if RAY_INSTALLED:
+        import ray
+        ray = ray
+
+    def __init__(self, distributed_opts, cls, init_args, init_kwargs):
+        if not RAY_INSTALLED:
+            raise ModuleNotFoundError("Module requires ray to be installed. pip install alibi[ray] ")
```
jklaise, line 313: We need to put this in `setup.py`

alibi/utils/distributed.py
```diff
@@ -217,3 +224,203 @@ def _chunk(values, chunksize):
...
+        return actor.return_attribute.remote(item)
+
+    def create_parallel_pool(self):
+        """
+        Creates a pool of actors that can execute explanations in parallel.
+        """
+
+        handles = [DistributedExplainer.ray.remote(self.explainer) for _ in range(self.n_jobs)]
+        workers = [handle.remote(*self.explainer_args, **self.explainer_kwargs) for handle in handles]
+        return DistributedExplainer.ray.util.ActorPool(workers)
```
jklaise, line 347: Is `ActorPool` now part of `ray` itself?

alibi/utils/distributed.py
```diff
@@ -217,3 +224,203 @@ def _chunk(values, chunksize):
...
+        ----------
+        X
+            Array to be split.
+        Returns
+        ------
+            A list of sub-arrays of X.
+        """
+
+        n_records = X.shape[0]
+        if isinstance(X, sparse.spmatrix):
```
jklaise, line 368: This would likely never happen as the error is raised already in `KernelExplainer`?

alibi/utils/distributed.py
alexcoca: Yes, same code as in our library.

alibi/utils/distributed.py
alexcoca: I guess that it's probably ok to just cast so we can split the batch and remove the `TypeError` and support the sparse matrix?

What are your thoughts on pandas? I was thinking to extract the values and just return numpy arrays with the explanations.

alexcoca:
Closing - to be merged to a development branch in alibi.

alexcoca:
Superseded by #303

PR #271, opened 2020-07-24
Fixed a small bug in the trustscore notebook

abs428:
I believe that the correct function that should be called is `run_model` rather than `run_lr`. This is my first time trying to contribute to a GitHub project. Please let me know if something is wrong.


jklaise:
Good catch! The results won't change as the `run_lr` function was in global scope, but it's good to have cleaner example!

abs428:
> Good catch! The results won't change as the `run_lr` function was in global scope, but it's good to have cleaner example!
> 
> Actually, I just checked that the `trustscore_mnist` notebook has the same issue, would you be up for correcting that too? If you make another commit and push it this PR will update automatically to include the new commit.

Sure. I can do that.

jklaise:
Thanks!

PR #290, opened 2020-08-18
Update metadata

jklaise:
This PR updates the base `Explainer` with a private method `_update_metadata` used for updating metadata during the code execution in explainers. This was previously done on `shap_wrappers` explainers but it is generic enough to include in the base class.

I've removed previously used pattern of an `allowed` set of metadata keys as we should be checking the keys are correct at test time not at runtime.

I've updated the `shap_wrappers` tests accordingly. I have removed the tests that tested only the `_update_metadata` method as now that is done by new tests of the `Explainer` base class. I have added a check that the set of `meta['params']` keys is a subset of `{KERNEL/TREE}_SHAP_PARAMS`. In doing this I found that subset equality was not preserved as in the TreeSHAP case there are more defined `PARAMS` than actually set and in the KernelSHAP case there are more parameters set than defined in `PARAMS`. Consequently, I've added the previously undefined `PARAMS` to KernelSHAP (`grouped`, and `transpose`). However, we need to decide if we want this to be a strict subset equality or an inclusion in the future.

Finally, I've removed `DEFAULT_META` and turned it into a factory (we can but the name back in and have the factory return a deep copy if we want to maintain semantics). This led to a bug where initializing an `Explanation` with no metadata (defaulting to  deep copying `DEFAULT_META`) would not actually create  a brand new dictionary! This bug was not relevant as we never initialize any `Explanation` with empty metadata.

### DIFF
--- a/alibi/api/interfaces.py
+++ b/alibi/api/interfaces.py
@@ -1,7 +1,7 @@
 import abc
 import json
 from collections import ChainMap
-from typing import Any, Optional, List
+from typing import Any
 import logging
 
 import attr
@@ -42,11 +42,10 @@ class Explainer(abc.ABC):
     def explain(self, X: Any) -> "Explanation":
         pass
 
-    def _update_metadata(self, data_dict: dict, params: bool = False, allowed: Optional[List[str]] = None) -> None:
+    def _update_metadata(self, data_dict: dict, params: bool = False) -> None:
         """
         This method updates the metadata of the explainer using the data from the `data_dict`. If the params option
-        is specified, then each key-value pair is added to the metadata `'params'` dictionary only if the key is
-        specified in the `allowed` dictionary
+        is specified, then each key-value pair is added to the metadata `'params'` dictionary.
 
         Parameters
         ----------
@@ -54,19 +53,11 @@ class Explainer(abc.ABC):
             Dictionary containing the data to be stored in the metadata.
         params
             If True, the method updates the `'params'` attribute of the metatadata.
-        allowed
-            Set containing the parameters allowed in the update.
         """
 
         if params:
             for key in data_dict.keys():
-                if key not in allowed:
-                    logger.warning(
-                        f"Parameter {key} not recognised, ignoring."
-                    )
-                    continue
-                else:
-                    self.meta['params'].update([(key, data_dict[key])])
+                self.meta['params'].update([(key, data_dict[key])])
         else:
             self.meta.update(data_dict)
 
--- a/alibi/api/interfaces.py
+++ b/alibi/api/interfaces.py
@@ -1,8 +1,7 @@
 import abc
-import copy
 import json
 from collections import ChainMap
-from typing import Any
+from typing import Any, Optional
 import logging
 
 import attr
@@ -12,13 +11,15 @@ import numpy as np
 
 logger = logging.getLogger(__name__)
 
+
 # default metadata
-DEFAULT_META = {
-    "name": None,
-    "type": [],
-    "explanations": [],
-    "params": {},
-}  # type: dict
+def default_meta() -> dict:
+    return {
+        "name": None,
+        "type": [],
+        "explanations": [],
+        "params": {},
+    }
 
 
 @attr.s
@@ -27,7 +28,7 @@ class Explainer(abc.ABC):
     Base class for explainer algorithms
     """
 
-    meta = attr.ib(default=copy.deepcopy(DEFAULT_META), repr=pretty_repr)  # type: dict
+    meta = attr.ib(default=attr.Factory(default_meta), repr=pretty_repr)  # type: dict
 
     def __attrs_post_init__(self):
         # add a name to the metadata dictionary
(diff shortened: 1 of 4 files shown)


alibi/api/interfaces.py
```diff
@@ -41,6 +42,25 @@ def __attrs_post_init__(self):
     def explain(self, X: Any) -> "Explanation":
         pass
 
+    def _update_metadata(self, data_dict: dict, params: bool = False) -> None:
+        """
+        This method updates the metadata of the explainer using the data from the `data_dict`. If the params option
```
alexcoca, line 47: Should just start with a verb, like the other docstrings here: "Updates  the metatata..."

alibi/api/interfaces.py
```diff
@@ -41,6 +42,25 @@ def __attrs_post_init__(self):
...
+        is specified, then each key-value pair is added to the metadata `'params'` dictionary.
+
+        Parameters
+        ----------
+        data_dict
+            Dictionary containing the data to be stored in the metadata.
```
alexcoca, line 53: Can exclude "Dictionary containing" as that is obvious from type hint.

alibi/api/tests/test_interfaces.py
```diff
@@ -8,6 +9,8 @@
 invalid_meta = []  # type: list
 invalid_data = {}  # type: dict
 
+data_dict = {'b': 2, 'c': 3, 'd': 4}
```
alexcoca, line 12: This parameterizes the test so we need to add marks on the tests and pass it that way rather than relying on scope? This is mostly a matter of consistency with the rest of the library but also makes the tests easier to read/understand fast.

alibi/explainers/tests/test_shap_wrappers.py
```diff
@@ -363,7 +363,6 @@ def get_column_ranks(X, ascending=False):
 ]
 
 
-# @pytest.mark.skip
```
alexcoca, line 366: Please leave these in as they make it very easy to run specific tests when things break down without messing with the CLI.

alibi/explainers/tests/test_shap_wrappers.py
```diff
@@ -999,6 +993,9 @@ def test_explain_kernel(monkeypatch, mock_kernel_shap_explainer, use_groups, sum
     assert explanation.meta.keys() == DEFAULT_META_KERNEL_SHAP.keys()
     assert explanation.data.keys() == DEFAULT_DATA_KERNEL_SHAP.keys()
 
+    # parameter keys
+    assert set(explanation.meta['params']) <= set(KERNEL_SHAP_PARAMS)
+
```
alexcoca, line 998: So I guess it would be wise to test also if there are things in `explanation.meta['params']` that should not be there? In other words the difference between said object and `KERNEL_SHAP_PARAMS` should be empty because technically every member of  `explanation.meta['params']` should be specified in `KERNEL_SHAP_PARAMS`?

alibi/explainers/tests/test_shap_wrappers.py
```diff
@@ -1079,7 +1058,6 @@ def test_build_explanation_kernel(mock_kernel_shap_explainer, task):
 data_types = ['array', 'frame']
 
 
-# @pytest.mark.skip
```
alexcoca, line 1082: As above for marks.

alibi/explainers/tests/test_shap_wrappers.py
```diff
@@ -1578,6 +1530,9 @@ def setter(obj: object, attr: str, val: Any):
         assert explanation.meta.keys() == DEFAULT_META_TREE_SHAP.keys()
         assert explanation.data.keys() == DEFAULT_DATA_TREE_SHAP.keys()
 
+        # parameter keys
+        assert set(explanation.meta['params']) <= set(TREE_SHAP_PARAMS)
```
alexcoca, line 1534: See comment in KernelShap tests.

alexcoca (changes requested):
Hi Janis,

Thanks for this great improvement to the code base. Some very minor comments, then we're good to merge.

alibi/explainers/tests/test_shap_wrappers.py
jklaise: This relates to my point whether we want strict subset equality? I'm sensing you are in favour so I'm happy to make the changes.

jklaise: `link` is never set for `KernelSHAP`

alibi/explainers/tests/test_shap_wrappers.py
jklaise: `TreeSHAP` does not set the following: `approximate`, `explain_loss`, `interactions`, `model_output`.

alibi/explainers/tests/test_shap_wrappers.py
alexcoca: Ok, that's an issue I'll correct in the patch. For some reason it has ended up in `data`.

alexcoca: Regarding the inequality, I would guess we should check all the information we expect to be in `KERNEL_SHAP_PARAM` has been updated? So in a way I would check for equality, because what we want is to set the parameters we said we would in the schema. There is no point in putting a parameter in the schema if we're not going to set it and there is no point of adding data that is not defined in the schema.

alibi/explainers/tests/test_shap_wrappers.py
alexcoca: I'm not sure why, the following lines of code set the first 3.

```python
self._update_metadata(
    {'interactions': self.interactions,
     'explain_loss': True if y is not None else False,
     'approximate': self.approximate,
},
     params=True,
)
```

alexcoca: `model_output` seems to have ended up in `data` (just as `link`), so I'll move it to params in the patch. Not sure why we made a decision to put them in `data` originally (and didn't change `params`) but looking at it now, it seems like this fits better in `params`.

alexcoca: I just realised that the `_update_metadata` is actually performed in `explain` which does not get called.

PR #292, opened 2020-08-19
Run linter on the interop branch

jklaise:
This should fix flake8 and mypy issues. Also a few bugs with docs fixed. Merge when checks pass.


alibi/explainers/cf_example_wb.py
```diff
@@ -1,10 +1,12 @@
+# flake8 noqa
 import tensorflow as tf
```
alexcoca, line 2: This file and its `*bb.py` counterpart were used by myself to do some sanity checks when making big changes. Do we keep them in this branch until we finish the development or shall we remove them?

alibi/explainers/experimental/counterfactuals.py
```diff
@@ -110,32 +109,32 @@
...
-    `lam_exploration_steps` consecutive iteration. The computational time is reduced as a result.
-    - `instance_proba_delta`: the model prediction on the counterfactual for the class with the highest probability for \
-    the original instance should decrease by at least this value in `lam_exploration_steps` to warrant continuation of \
-    optimisation for a given :math:`\lambda` in the initial search loop
-    - `lam_cf_threshold`, `lam_multiplier`, `lam_divider`: see `bisect_lam` docstring in `Watcher counterfactual`
-    - `common_ration`, `decay_steps`: role in determining the original optimisation schedule for :math:`\lambda` as  \
-    described above
-    - 'max_lam_steps': maximum number of times to adjust the regularization constant before terminating the search \
-    (number of outer loops).
+ which depends on the hyperparameter :math:`\lambda`. The explainer will first try to determine a suitable range for
```
alexcoca, line 112: Have we checked this renders fine?

alibi/utils/frameworks.py
```diff
@@ -69,9 +70,9 @@ def infer_device(predictor, predictor_type: str, framework: str) -> Union[None,
     """
 
     if framework == 'tensorflow':
-        return  # type: ignore
+        return None
     if predictor_type == 'blackbox':
-        return
+        return None
```
alexcoca, line 75: I disagree with this. `return` means `return None` in `python`, this is a language convention. If the typing library doesn't know about this, then this is a problem with it so I say we should stick with the `# type:  ignore` directive.

alibi/explainers/experimental/counterfactuals.py
```diff
@@ -1062,8 +1061,8 @@ def __init__(self,
             'method_opts': method_opts,
             'feature_range': feature_range,
             'framework': framework,
-        }
-        self._explainer = self._explainer_type(*explainer_args, **explainer_kwargs, **kwargs)
+        }  # type: Dict
+        self._explainer = self._explainer_type(*explainer_args, **explainer_kwargs, **kwargs)  # type: ignore
```
alexcoca, line 1065: Why do we ingnore the type here? Isn't this a `dict`?

alibi/utils/frameworks.py
```diff
@@ -36,12 +37,12 @@ def tensorflow_installed() -> bool:
     if not has_tensorflow:
         raise ImportError(template.format(pkg=tf_required))
     if int(tf_version[0]) > 1:
```
alexcoca, line 39: We need to introduce here the change we discussed privately, otherwise the docs do not build.

alexcoca (changes requested):
Hi Janis,

Thanks for cleaning this up. We need to add appropriate casting in `alibi.utils.frameworks.py` as we discussed, otherwise the docs won't build. Other than that just one or two comments about typing.

alibi/explainers/cf_example_wb.py
jklaise: Can keep them as long as we remember to delete before merge to master.

alibi/explainers/experimental/counterfactuals.py
jklaise: I checked it.

alibi/utils/frameworks.py
jklaise: Ok fair enough, I just thought this is cleaner/shorter than doing the extra type declaration.

alibi/explainers/experimental/counterfactuals.py
jklaise: There was a reason for this...

jklaise: With no type info I get the following error: `alibi/explainers/experimental/counterfactuals.py:1065: error: Function gets multiple values for keyword argument "predictor"`

jklaise: It's because `predictor` is somehow present in both `explainer_args` and `kwargs`.

jklaise: The correct type (correct me if I'm wrong) would be `#type: _WachterCounterfactual`, but the problem persists, so I just originally ignored the type.

jklaise: This is a known bug so I would stick with `# type: ignore`: https://github.com/python/mypy/issues/6799

alexcoca: Agreed.

alibi/utils/frameworks.py
alexcoca: I still insist :)

alexcoca: Actually, I take back my original assertion. Having researched a bit, we can see that your modification makes sense in this case, according to [this](https://stackoverflow.com/questions/15300550/return-return-none-and-no-return-at-all) source.

alexcoca (approved):
All nice and tiy, I'll merge this.

PR #294, opened 2020-08-28
Pin shap version to 0.35.0 to prepare for upgrade for 0.36.0

alexcoca:
Pinning `shap` version until we uprade the wrappers and examples to support `0.36.0`


arnaudvl (changes requested):
Also pin version in setup.py?

arnaudvl (approved):

PR #295, opened 2020-09-03
Patch for alibi to support shap 0.36.0

alexcoca:
This PR updates our implementation for KernelSHAP and the tests so that our library works with shap `0.36`. 

Changes:

- `DenseData` and `DenseDataWithIndex` object can no longer be imported from `shap.common`. Instead we import `shap.utils._legacy as shap_utils` and use `shap_utils.DenseData`  or `shap_utils.DenseDataWithIndex` to refer to the objects. This change affects `shap_wrappers.py` and `test_shap_wrappers.py`. This change means that our library is no longer compatible with shap `0.35` (I did not see any reason to maintain that compatibility, especially because `0.35` is incompatible with the latest `xgboost` versions)
- Small updates to `kernel_shap_wine_intro.ipynb` (typos)
- Reduced number of runs and saved results every run for the KernelSHAP convergence experiment in `interventional_tree_shap_adult_xgb.ipynb` example.

Fixes:
- This upgrade fixes existing issues encountered when running `TreeShap` with `xgboost>=1.1`


jklaise:
I assume the `shap.utils._legacy` will eventually stop being supported and the custom data types as well? We should be vigilant about this, I wonder if it's worth dropping support for these data types soon.

alexcoca:
> 
> 
> I assume the `shap.utils._legacy` will eventually stop being supported and the custom data types as well? We should be vigilant about this, I wonder if it's worth dropping support for these data types soon.

We need `DenseData` to support categorical variables in the right way, we can't do away with it for this reason (although the workaround is not to encode the data, as one of the examples shows). So long `KernelExplainer` is still supported by `shap`, the `_legacy` module will continue to exist because they are used internally in their code. If they stop using it and upgrade `KernelExplainer`, then we can upgrade too.

jklaise (approved):

PR #299, opened 2020-09-25
Distributed backend for Alibi

alexcoca:
This pull request builds on the ideas introduced in #270 in order to build a distributed backend for `alibi` using the `ray` library. 

**New objects**

1. `DistributedExplainer` class in `alibi.utils.distributed`. The role of this class is to create a pool of explainer replicas and distribute a batch of instances to be explained across the pool.

2. `PoolCollection` class in `alibi.utils.distributed`. This class is used when a particular explainer involves solving multiple independent optimisation problems. It instantiates `DistributedExplainer` objects in different processes, which allows us to distributed batches across different optimisation processes, simultaneously.

**Things to note**

1. The `DistributedExplainer` object has a `return_generator` option, which allows the user to return a generator which yields the computed explanations as they are returned. In this way, the caller can use the results of the computations before the whole batch is completed.

2. The above cannot be used in conjunction with the `PoolCollection` class, since this would involve pickling a generator (not supported in CPython). This suggests that the mechanism for returning results as they complete could be achieved by passing an appropriate callback to the `DistributedExplainer` object directly. That is, the generator would be consumed inside the `DistributedExplainer` by a function that can return the results to a given context (e.g., an endpoint).


jklaise:
Should we use Ray's implementation of `ActorPool` going forward?

alibi/utils/distributed.py
```diff
@@ -217,3 +232,523 @@ def _chunk(values, chunksize):
...
+
+    def return_attribute(self, name: str) -> Any:
+        """
+        Returns an attribute specified by its name. Used in a distributed context where the properties cannot be
+        accessed using the dot syntax.
+        """
+
+        # if the attribute requested if of the PoolCollection object, return it
+        try:
+            return self.__getattribute__(name)
```
jklaise, line 523: I'm not sure I follow, how is `PoolCollection` related to this?

alibi/utils/tests/test_distributed.py
```diff
@@ -0,0 +1,366 @@
...
+    result = invert_permutation(permutation_generator)
+    np.testing.assert_allclose(np.array(permutation_generator)[result], np.array(np.arange(result.shape[0])))
+
+
+# MockExplainer args and kwargs
+explainer_init_args = [[(0.01,), (0.05, )], ]   # type: ignore
+explainer_init_kwargs = [[{'multiplier': 2}, {'multiplier': 3}], [{'multiplier': 2}, ]]  # type: ignore
+
+batch_size = [2]
+ncpus = [1, len(explainer_init_kwargs[0])]  # that's so that we test the
```
jklaise, line 260: Test the...?

alibi/utils/tests/test_distributed.py
```diff
@@ -0,0 +1,366 @@
...
+
+    ray.shutdown()
+
+
+# MockExplainer positional args and kwargs
+explainer_init_args = [[(0.01,), (0.05, )], ]  # type: ignore
+explainer_init_kwargs = [[{'multiplier': 2}, {'multiplier': 3}], ]  # type: ignore
+
+batch_size = [2]
+ncpus = [len(explainer_init_kwargs[0])]  # that's so that we test the
```
jklaise, line 322: Test the...?

alibi/utils/tests/test_distributed.py
```diff
@@ -0,0 +1,366 @@
...
+    )
+
+    X = data_generator
+
+    explanations = pool_collection.get_explanation(X)
+    for this_expln_result, this_expln_args, this_expln_kwargs in zip(explanations, expln_args, expln_kwargs):
+        batched_result = np.concatenate(this_expln_result, axis=0).squeeze()
+        assert np.isclose(np.unique(batched_result / X), this_expln_kwargs['multiplier'], atol=atol)
+
+    # retrieve process state. This requires setting
```
jklaise, line 357: setting what?

jklaise (changes requested):
Nice work Alex! Let's set some time aside to go through the intended use cases as per one of my comments.

alexcoca:
> Should we use Ray's implementation of `ActorPool` going forward?

The short answer is yes. I took the `alibi.utils.distributed.py` version from a branch in the `ray` repo before it was merged and added batching capability to it in order to implemenent the distributed anchor tabular algorithm.

Things to note:

- Batching logic will have to be implemented outside the `ActorPool`  object if you decide to only use `ray.util.ActorPool` simply importing from `ray.util` as opposed to our code will break the achors code

alexcoca:
> Should we use Ray's implementation of `ActorPool` going forward?

alibi/utils/distributed.py
```diff
@@ -217,3 +232,523 @@ def _chunk(values, chunksize):
...
+        self.batch_size = distributed_opts['batch_size']
+        self.return_generator = return_generator
+        algorithm = distributed_opts.get('algorithm', 'default')
+        if 'algorithm' == 'default':
+            logger.warning(
+                "No algorithm specified in distributed option, default target function and postprocessing will be "
+                "selected."
+            )
+        self.target_fcn = default_target_fcn
+        # TODO: TBD: @JANIS - THE POSTPROCESSING COULD BE HANDLED BY CALLER? OR THE CALLER COULD PASS THE FUNCTION?
```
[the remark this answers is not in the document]
alexcoca: Differences:

- target function: used by ray to schedule a remote explainer call (see `submit` method in `ray.util.ActorPool`

- postprocessing function: used to re-assemble the results computed in parallel into a single object (think e.g., concatenation of numpy arrays in the simplest case). It needs to be custom because each explainer returns something different. We can have a generic one that assumes the input array is a list numpy arrays and concatenates them, and register different functions for algorithms with more complex output types (e.g., kernel shap). 

[NB: On second thoughts I am thinking that `singledispatch` might be another way to implement this]

alibi/utils/tests/test_distributed.py
```diff
@@ -0,0 +1,366 @@
...
+
+    atol = 1e-5  # abs tolerance for floating point comparisons
+    distributed_explainer = DistributedExplainer(distributed_opts, MockExplainer, expln_args, expln_kwargs)
+    assert distributed_explainer.target_fcn.__name__ == 'default_target_fcn'
+    assert not distributed_explainer.post_process_fcn
+    assert ray.is_initialized()
+    assert isinstance(distributed_explainer.pool, ray.util.ActorPool)
+    assert len(distributed_explainer.pool._idle_actors) == distributed_opts['n_cpus']
+
+    # TODO: RAY.GET SHOULD BE APPLIED WHEN RETURNING ATTRIBUTES? NOT HERE?
```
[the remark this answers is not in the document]
alexcoca: Sorry, this is an errant TODO. When I first wrote the test I was calling `ray.get` to obtain the attribute values for the assertion but subsequently realised that we should return the actual attributes, not futures.

alibi/utils/tests/test_distributed.py
```diff
@@ -0,0 +1,366 @@
...
+    # test remote process is initialised properly
+    for actor_idx in range(distributed_opts['n_cpus']):
+        distributed_explainer.actor_index = actor_idx
+        assert distributed_explainer.multiplier == expln_kwargs['multiplier']
+        assert np.isclose(distributed_explainer.sleep_time, expln_args[0], atol=atol)
+
+    ray.shutdown()
```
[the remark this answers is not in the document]
alexcoca: When you want to completely tear down ray (e.g., kill the underlying processes, terminate actors, etc). This should be done in between tests to ensure corectness. Also worth checking out the ray docs h[ere](https://docs.ray.io/en/master/package-ref.html)

alibi/utils/distributed.py
```diff
@@ -217,3 +232,523 @@ def _chunk(values, chunksize):
...
+        Parameters
+        ----------
+        See constructor documentation.
+        """
+
+        handles = [DistributedExplainer.ray.remote(explainer_type) for _ in range(self.n_processes)]
+        workers = [handle.remote(*explainer_init_args, **explainer_init_kwargs) for handle in handles]
+        return DistributedExplainer.ray.util.ActorPool(workers)
+
+    def get_explanation(self, X: np.ndarray, **kwargs) -> \
```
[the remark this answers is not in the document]
alexcoca: In brief:

- The underlying explainer doesn't need to work with minibatches if this does not make sense - simply set the `batch_size` to `1` in the default distributed options for the algorithms that don't do minibatches and raise an error if user overrides it. However, if in the future we have batched implementations that leverage vectorisation, we don't need to change the code at all.

- we have a fixed resource pool specifed by the user in the `distributed_opts` kwarg. If we have many more instances to explain, we queue them and feed them to the pool as some of the actors there finish the job. In this way we keep the number of processes under control

- if we want to play around with different `kwargs` for `explain` (aka, esentially run experiments in parallel), the most appropriate way would be to use the `PoolCollection` object - suggest a simple config file approach where the user can specify the algorithm and its configuration in a yaml and then run distributed experiments. Foundation work has been done in [this ](https://github.com/alexcoca/alibi/blob/Benchmarking/benchmark) branch.

alibi/utils/distributed.py
alexcoca: Sorry, this is garbled. You can do `my_pool_collection.prop` where `prop` is a either an attribute of the `PoolCollection` or of the explainer wrapped by the `DistributedExplainer`. `__geattr__` handles the latter case.

jklaise (approved):
Looks great, happy to merge when tests pass.

Issue #289, opened 2020-08-18
Investigate extending IntegratedGradients for text to work on raw text


jklaise:
Whilst `AnchorText` works directly on raw text, `IntegratedGradients` works on the token level. One reason for this is that `IntegratedGradients` is use case agnostic - tabular data, images and text are handled in the same way at the cost of the user having to do pre- and post-processing for text to display explanations in the raw text space.

This would amount to having a wrapper like `IGText` which is a text-specific `IntegratedGradients` that works on arrays of strings. Fundamentally this means coupling the tokenizer together with the model. What needs to be explored if the proposal would work with all kinds of tokenizers and model backends (TensorFlow, PyTorch when we have IG for PyTorch).

abhishek-niranjan:
@jklaise just my thought, instead of the wrapper (`IGText`) dealing with a tokenizer in it's `__init__()`, you can just pass the `word_to_index dictionary` since most of the python based tokenizers implement a dictionary to convert tokens to token_ids. This would avoid the problem of dealing with multiple Frameworks (TensorFlow, PyTorch, etc. )

jklaise:
I think this is a good idea so given the `word_to_index` dictionary we could use it to output meaningful explanations. However, unless I'm missing something, this is not enough for the method to run on raw text, the tokenizer still would need to do its job to produce the tokens.

jklaise:
@abhishek-niranjan yes that's exactly the issue, the tokenizer is model-specific so we shouldn't assume that a simple heuristic will work as we may be generating invalid tokens with respect to the model.

abhishek-niranjan:
@jklaise right. So the idea initially suggested by you is the only ideal way to go ahead. 
>  What needs to be explored if the proposal would work with all kinds of tokenizers and model backends (TensorFlow, PyTorch when we have IG for PyTorch).

And this will need to be thought over then. Are you looking to add this functionality once alibi is migrated to TF2.0?

jklaise:
@abhishek-niranjan this is likely to be a high priority task especially if people are interested in the functionality. Alibi does actually work on TF2.x now, although we are still working on converting the code-base to idiomatic TF2.x code together with the necessary abstractions to enable development of PyTorch methods in the future (interop-refactor branch).

abhishek-niranjan:
@jklaise what do you think about the following:

```
class IGText(IntegradedGradients):
  def __init__(model, layer, method, word_to_index, preprocessing_callable, n_steps, internal_batch_size):
[...]

PR #300, opened 2020-10-06
Integrated gradients on raw text PoC

jklaise:
This is a WIP PR addressing #289.

For applying integrated gradients on raw text, I've written a new class `IntegratedGradientsText` which takes an additional object of type `TextTransformer`. The `TextTransformer` object must define two methods:

- `texts_to_tokens` - given a list of strings return a list of lists of string tokens (i.e. a ragged array). This is necessary to output attributions for each token.
- `texts_to_array` - given a list of strings return a homogenous numpy array ready for model consumption. This is necessary to prepare the raw text for the model.

Thus, the user has to inherit from `TextTransformer` and implement these two methods for a custom use case. I have also provided an implementation `KerasTextTransformer` which takes a Keras `Tokenizer` and a parameter `maxlen` (to define the array dimension) - see example.

Main things to discuss:

- In the implementation I assume that the attributions are calculated with respect to the embedding layer, hence I've hardcoded  a sum over the embedding axis to produce attributions on the token level. This may be brittle and fail if other layers are to be explained. As a minimum we should provide copious documentation of how to use IG for text (or even ask if we should only support attributions for the embedding layer?)
- The baselines are with respect to the layer that is being explained, so in the example if custom baselines are used, these should have the same shape as the layer explained.
- It is not clear how to name the returned attributions. Currently `attributions` refer to the raw attributions for the layer explained and have the same shape. I have named the summed attributions `attrs` which is a list of arrays of floats (i.e. a ragged array, one for each instance) of the same size as `tokens` which are also returned (list of lists of string tokens, i.e. a ragged array). We could also apply the same reasoning to `baselines` and return a summed version as a ragged array or even allow a ragged array to be taken as an input baseline to `explain`.


alibi/explainers/integrated_gradients.py
```diff
@@ -428,6 +429,7 @@ def explain(self,
             else:
                 grads_b = _gradients_input(self.model,
                                            tf.dtypes.cast(paths_b, self.input_dtype), target_b)
+                # TODO: is casting the right thing to do? Will loose information if input types differ
```
arnaudvl, line 432: Think that's fine, input types for a given model / layer should be consistent anyway.

alibi/explainers/integrated_gradients.py
```diff
@@ -481,3 +483,42 @@ def build_explanation(self,
...
+                                              internal_batch_size=internal_batch_size)
+        self.text_transformer = text_transformer
+
+    def explain(self,
+                X: List[str],
+                baselines: Union[None, int, float, np.ndarray] = None,
+                target: Union[None, int, list, np.ndarray] = None) -> Explanation:
+        tokens = self.text_transformer.texts_to_tokens(X)
+        X = self.text_transformer.texts_to_array(X)
+
```
arnaudvl, line 511: Tiny comment but I would stick with `text_to_tokens` and `text_to_array`

alibi/explainers/integrated_gradients.py
```diff
@@ -481,3 +483,42 @@ def build_explanation(self,
...
+        # TODO: baselines assumed to be wrt to the pre-processed X
+        explanation = self._explainer.explain(X=X, baselines=baselines, target=target)
+
+        # exctract data, augment, and create a new explanation object
+        # NB: we have to create a new object for attribute access to work for the new attributes
+        meta, data = explanation.meta, explanation.data
+        attrs = data['attributions'].sum(axis=-1)  # TODO: dangerous to assume these are for embeddings!?
```
arnaudvl, line 518: Seems like a good default for the vast majority of the cases. Maybe add an optional kwarg if the layer is not an embedding layer, then can pass a Callable which applies the correct operation to get your attributions?

alibi/explainers/integrated_gradients.py
```diff
@@ -481,3 +483,42 @@ def build_explanation(self,
...
+                                              method=method,
+                                              n_steps=n_steps,
+                                              internal_batch_size=internal_batch_size)
+        self.text_transformer = text_transformer
+
+    def explain(self,
+                X: List[str],
+                baselines: Union[None, int, float, np.ndarray] = None,
+                target: Union[None, int, list, np.ndarray] = None) -> Explanation:
+        tokens = self.text_transformer.texts_to_tokens(X)
```
arnaudvl, line 509: One slightly confusing aspect of *IntegratedGradientsText* is that it cannot be applied to tokenized text. Why not allow X to be of type `Union[List[str], List[List[int]]]` (latter type for tokens) and make tokenization step optional?

alibi/explainers/integrated_gradients.py
```diff
@@ -481,3 +483,42 @@ def build_explanation(self,
...
+        X = self.text_transformer.texts_to_array(X)
+
+        # TODO: baselines assumed to be wrt to the pre-processed X
+        explanation = self._explainer.explain(X=X, baselines=baselines, target=target)
+
+        # exctract data, augment, and create a new explanation object
+        # NB: we have to create a new object for attribute access to work for the new attributes
+        meta, data = explanation.meta, explanation.data
+        attrs = data['attributions'].sum(axis=-1)  # TODO: dangerous to assume these are for embeddings!?
+        attrs = self.text_transformer.array_to_ragged_array(attrs)
```
arnaudvl, line 519: I assume here the array with IG attributions is mapped back to the ragged original input text version (e.g. 7 words or tokens in 1st instance and 19 in the next one etc). But wouldn't the tokens need to follow the same mapping again? Otherwise the padding etc tokens are not cleaned up.

PR #304, opened 2020-10-19
Upgrade deprecated pre-commit flake8, add nbqa-pyupgrade

MarcoGorelli:
Hi again :wave:

I noticed that in your pre-commit file you reference a deprecated hook, see https://github.com/pre-commit/pre-commit-hooks

> flake8: instead use upstream flake8

So, I've replaced that with the repo homepage (`pre-commit run flake8 --all` still passes). Also, since the minimum Python version you support is Python3.6, I thought I'd add a hook to upgrade the syntax to make sure of Python3.6+'s features

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,8 +1,13 @@
 repos:
-  - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.4.0
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: 3.8.4
     hooks:
       - id: flake8
+  - repo: https://github.com/nbQA-dev/nbqa
+    rev: 0.3.2
+    hooks:
+      - id: nbqa-pyupgrade
+        args: [--nbqa-mutate, --py36-plus]
   - repo: https://github.com/pre-commit/mirrors-mypy
     rev: v0.740
     hooks:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,7 @@ one of the requirements for running the test suite.
 We use `flake8` for linting adhering to PEP8 with exceptions defined in `setup.cfg`.
 
 ## Syntax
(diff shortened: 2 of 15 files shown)


jklaise:
Looks great, thanks!

Would be good to add a line in https://github.com/SeldonIO/alibi/blob/master/CONTRIBUTING.md describing what the new hook does (I assume `--nbqa-mutate`, `--py36-plus` work together to format notebooks for Python 3.6+ ?).

MarcoGorelli:
> Would be good to add a line in https://github.com/SeldonIO/alibi/blob/master/CONTRIBUTING.md describing what the new hook does (I assume `--nbqa-mutate`, `--py36-plus` work together to format notebooks for Python 3.6+ ?)

Sure, done - `--py36-plus` just specifies the minimum Python version, `--nbqa-mutate` just tells `nbqa` to let `pyupgrade` modify notebooks. You could add just `pyupgrade` by itself as a separate hook to upgrade the syntax of your source code.

> Btw is there any noticeable performance hit using it on all our notebooks?

This one's very fast: timing from my laptop:
```console
$ time pre-commit run nbqa-pyupgrade --all
nbqa-pyupgrade...........................................................Passed

real    0m2.070s
user    0m1.844s
sys     0m0.238s
```

jklaise (approved):

jklaise:
Thanks @MarcoGorelli !

MarcoGorelli:
@jklaise just out of interest, when you run `mypy` locally, do you get lots of errors? At least,
```
pre-commit run mypy --all-files
```
throws quite a few, though `mypy alibi` doesn't (I think this is likely because pre-commit runs `mypy` in an isolated environment)

jklaise:
@MarcoGorelli not sure why that happens, which files are affected? Tbf I haven't used pre-commit for a while (had to switch it off as it was hampering prototyping new features) so I just run `mypy alibi` which is the same as the command run on Travis. When I did use it though it worked as intended.

MarcoGorelli:
Lots of them...this was just in case you wanted to run pre-commit during CI, as currently mypy would fail

PR #305, opened 2020-10-19
Remove prettyprinter dependency.

jklaise:
Replaced prettyprinter with a slightly modified standard library version
of PrettyPrinter. This should allow us to publish alibi on conda as all
dependencies are also required to be published on conda.


jklaise:
Looks like `sort_dicts` was only introduced in Python 3.8.

PR #306, opened 2020-10-20
Anchor exp return additions

jklaise:
* Introduce a `data.raw['instances']` field which is a numpy array (or
  List[str] in the case of AnchorText) which adds an extra dimension to
  the front for singleton explanations (currently all anchor methods).
* Breaking change: make `data.raw['prediction']` be an array of labels
  even in the case of singleton explanations (currently all anchor
  methods).

@SachinVarghese please have a look if this is what you have in mind and if there are other things that should be changed or added. In particular, anchor methods currently only take single instances instead of arrays (or lists in the case of AnchorText), whilst the SHAP methods take arrays of instances.


SachinVarghese:
This looks good. @jklaise.

PR #321, opened 2020-11-25
Intgrad multiple inputs

gipster:
Extends the integrated gradients method to model with multiple inputs. For each input of the model, attributions are calculated and returned in a list. Also extends the method allowing to calculate attributions for multiple internal layers. If a list of layers is passed, a list of attributions is returned.


alibi/explainers/integrated_gradients.py
```diff
@@ -388,84 +468,120 @@ def explain(self,
...
 
         # defining integral method
         step_sizes_func, alphas_func = approximation_parameters(self.method)
         step_sizes, alphas = step_sizes_func(self.n_steps), alphas_func(self.n_steps)
-
-        # construct paths and prepare batches
-        paths = np.concatenate([baselines + alphas[i] * (X - baselines) for i in range(self.n_steps)], axis=0)
+        target = _format_target(target, nb_samples)
         if target is not None:
             target_paths = np.concatenate([target for _ in range(self.n_steps)], axis=0)
```
jklaise, line 487: `target_paths` is not defined for regression models which currently fails the tests as it's a required argument for `_calculate_sum_int`

alibi/explainers/tests/test_integrated_gradients.py
```diff
@@ -48,6 +55,95 @@ def ffn_model(request):
     return model
 
 
+@pytest.fixture()
+def ffn_model_multi_inputs(request):
+    """
+    Simple feed-forward model with configurable data, loss funciton, output activation and dimension
```
jklaise, line 61: funiction --> function

alibi/explainers/tests/test_integrated_gradients.py
```diff
@@ -48,6 +55,95 @@ def ffn_model(request):
...
+    assert explanations['data']['predictions'].shape[0] == N_TEST
+
+
+@pytest.mark.parametrize('ffn_model_multi_inputs', [({'output_dim': 1,
+                                                      'activation': 'sigmoid',
+                                                      'loss': 'binary_crossentropy',
+                                                      'X_train_multi_inputs': X_train_multi_inputs,
+                                                      'y_train': y_train_classification_ordinal})], indirect=True)
+@pytest.mark.parametrize('method', INTEGRAL_METHODS)
+def test_integrated_gradients_binary_classification_single_output_mi(ffn_model_multi_inputs, method):
```
jklaise, line 121: mi --> multi_inputs, best to be explicit in test function names

alibi/explainers/tests/test_integrated_gradients.py
```diff
@@ -48,6 +55,95 @@ def ffn_model(request):
...
+    return model
+
+
+@pytest.mark.parametrize('ffn_model_multi_inputs', [({'output_dim': 2,
+                                                      'activation': 'softmax',
+                                                      'loss': 'categorical_crossentropy',
+                                                      'X_train_multi_inputs': X_train_multi_inputs,
+                                                      'y_train': y_train_classification_categorical})], indirect=True)
+@pytest.mark.parametrize('method', INTEGRAL_METHODS, ids='method={}'.format)
+def test_integrated_gradients_binary_classification_mi(ffn_model_multi_inputs, method):
```
jklaise, line 89: mi --> multi_inputs, best to be explicit in test function names

jklaise (changes requested):
Thanks for the PR Giovanni, mostly minor things to fix.

alibi/explainers/integrated_gradients.py
```diff
@@ -477,7 +593,7 @@ def build_explanation(self,
         data.update(predictions=predictions)
 
         # calculate convergence deltas
-        deltas = _compute_convergence_delta(self.model, attributions, baselines, X, target)
-        data.update(deltas=deltas)
+        delta = _compute_convergence_delta(self.model, self.input_dtypes, attributions, baselines, X, target)
+        data.update(deltas=delta)
```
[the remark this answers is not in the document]
gipster: It's not a big deal but maybe the history of the changed name could be interesting :) : 
I think originally it was deltas because there's one delta for each data point in the batch. 
For multiple inputs, at the beginning I thought there should be also one delta for each input, so I changed the old deltas into delta to free deltas and assign it to the final list.
Then I realized that even with multiple inputs there's only one delta for each input, so I got rid of a loop over the inputs.
In this process deltas became delta. I changed it back to deltas

jklaise (approved):

PR #325, opened 2020-12-02
Resolve bug caused by constant features and improve binning.

jklaise:
Resolves #319.

- For constant features, ALE is now calculated in a small interval
  around the single value +-10% of the feature value or at least 0.1.
- For low resolution features the default is now to calculate the number
  of unique values for each feature and to use these values instead of a
  grid returns by the quantile function.

PR #326, opened 2020-12-03
Ale select features

### LINKED ISSUE #324: Allow specifying features for which to calculate ALE
This was raised in #319, opening a separate issue to keep track.

The main decision for this is whether to specify the features at `__init__` or at `explain` and how this functionality would interact with the `feature_names` argument (i.e. should we always expect `feature_names` for all features or just for the ones being explained?).

jklaise:
Resolves #324.

This PR also includes code from #325 which is independent of the functionality implemented here, this is to avoid merge conflicts.

### DIFF
--- a/alibi/explainers/ale.py
+++ b/alibi/explainers/ale.py
@@ -21,7 +21,10 @@ class ALE(Explainer):
                  feature_names: Optional[List[str]] = None,
                  target_names: Optional[List[str]] = None,
                  check_feature_resolution: bool = True,
-                 low_resolution_threshold: int = 10) -> None:
+                 low_resolution_threshold: int = 10,
+                 extrapolate_constant: bool = True,
+                 extrapolate_constant_perc: float = 10.,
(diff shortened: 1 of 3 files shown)


alibi/explainers/ale.py
```diff
@@ -60,12 +72,15 @@ def explain(self, X: np.ndarray, min_bin_points: int = 4) -> Explanation:
             An `Explanation` object containing the data and the metadata of the calculated ALE curves.
 
         """
+        # TODO: need to check dtype of X, if int then quantiles will be rounded to nearest int
```
arnaudvl, line 75: Still relevant?

jklaise: No, I don't think so.

jklaise:
@arnaudvl changes made:
- exposed parameters to control the handling of constant features
- added a test
- removed the TODO comment regarding `int` and `float` arrays
- added missing docstring on the `features` parameter for `explain`

arnaudvl (approved):

PR #334, opened 2021-01-22
Avoid shap 0.38.1

### LINKED ISSUE #333: Shap 0.38.1 breaks imports
Most recent release breaks importing which affects us. Need to specify avoiding this shap version in `setup.py` and make a new release. https://github.com/slundberg/shap/pull/1749

jklaise:
Resolves #333.


requirements/requirements_all.txt
```diff
@@ -27,7 +27,7 @@ ipykernel>=5.1.0
 ipython>=7.2.0
 codecov>=2.0.15
 pre-commit>=1.20.0
-shap>=0.36.0
+shap>=0.36.0,!=0.38.1
```
RafalSkolasinski, line 30: Is it only 0.38.1, do we know if 0.38.2 will be ok?

jklaise: Well I was hoping to catch any new breaking releases in the daily cronjob.

PR #337, opened 2021-01-28
Fix incorrect coverage calculation in anchor tabular

### LINKED ISSUE #328: Unclear coverage calculation in AnchorTabular
Hi,

i applied the AnchorTabular module to the standard Titanic dataset and got weird coverage values.

Here is an example Anchor result for a non-surviving instance:
image
The Anchor appears to make sense but coverage seems high. As far as I understood, coverage gives the fraction of other sampled instances that this Anchor can be applied to. Since the titanic dataset is small, I guess this fraction can be seen as the fraction of the overall data.

Second example for a surving instance:
image
Here, I notices that the coverage of example 1 and 2 sum up to 1 so I checked the dataset distribution of gender and found that the dataset consists of 78% males and 22% females. 

Am I missing something in the definition of coverage or is the coverage calculation only considering the first rule of the anchor?

Thanks for any insight that you can provide.

jklaise:
Resolves #328 . The bug was caused by a call to `reversed` on a list holding the number of instances from the training set satisfying each partial anchor. As a result, for every anchor of length >=2, the coverage value was being overwritten by the coverage value of the partial anchor of length 1. I'm not sure why the `reversed` call was there in the first place, I don't think it serves any other purpose.

Tested manually that this gives the correct coverage calculations on the Adult example. Ideally we would have a test for this, but it's non-trivial to check.

### DIFF
--- a/alibi/explainers/anchor_tabular.py
+++ b/alibi/explainers/anchor_tabular.py
@@ -276,7 +276,7 @@ class TabularSampler:
             [allowed_rows[feat] for feat in uniq_feat_ids],
             np.intersect1d),
         )
-        nb_partial_anchors = np.array([len(n_records) for n_records in reversed(partial_anchor_rows)])
+        nb_partial_anchors = np.array([len(n_records) for n_records in partial_anchor_rows])
         coverage = nb_partial_anchors[-1] / self.n_records
 
         # if there are enough train records containing the anchor, replace the original records and return...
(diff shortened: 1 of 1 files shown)


jklaise:
On further inspection the `reversed` call was actually used as a means of sorting the number of partial anchors list in increasing order to call `np.searchsorted` later on which ensures that enough samples are returned. I've reinstated the `reversed` call but changed the coverage calculation to index `0` instead of `-1` as after sorting the first entry will be the coverage of the full anchor (rather than the last).

arnaudvl (approved):

PR #342, opened 2021-02-05
Pydantic schemas for explainer outputs

jklaise:
This PR is a draft proposal for specifying the return schemas of `Explanation` objects resulting from calls to `explain` methods on `Explainer` objects.

# Motivation
Currently finding out the exact return schema of of any explainer is cumbersome as you either need to rely on documentation (which may be incomplete) or read the source code directly.
Moreover, in some instances the return type (an instance of `Explanation`) may have a slightly different schema for the same algorithm depending on which code branch was taken at runtime.
There is currently no method of enforcing the return type to adhere to a well-defined schema.

This proposal introduces schemas (`api/schemas.py`) for every explainer which codifies what to expect from each explainer.
This is primarily intended to make life easier for developers in the following ways:

1. For application developers building on top of `alibi` - a clear, single source of truth of what to expect from every explainer method.
2. For Alibi developers - a single place to define the return schema whilst developing a new explainer.
   Also, a single place to clear up inconsistencies in the current schema definitions (e.g. different attribute names for the same type of data across different explanations).

# Features
The return schemas are defined in `api/schemas.py` and are defined as `pydantic` data models.
The models are often hierarchical to reflect the rather complex return types of some methods.

The data types are defined in terms of Python native types, `pydantic` constrained types (e.g. `confloat` for a `float` with value constraints) and a custom `Array` type defined in `api/types.py` to model a `numpy` array.
The custom `Array` type is used to model more granular array types, e.g:
- `Array` - arbitrary array
- `Array[float]` - array of floats of arbitrary dimension and size
- `Array[int, (-1, 3)]` - a 2D array of ints with 3 columns and an arbitrary number of rows.

These schemas are used at runtime to ensure the return types adhere to the schemas, see `api.interfaces.Explanation`.
When a new `Explanation` model is initialized (which occurs before returning from `explain`), a private attribute `_explanation_model` is created which takes the `meta` and `data` dictionaries and validates the entries against the appropriate schema.
This is effectively a runtime check that ensures that what is returned adheres to the schema definition.
(The reason to not return `_explanation_model` directly to the user is to provide a form of encapsulation - we want to use our custom `Explanation` object as the return type with potentially custom methods to manipulate/display it etc.).

`Explanation` objects can be encoded to `json` (e.g. for persistence) by using the `to_json` method. This will use the custom `numpy` encoder in `_explanation_model` to convert arrays to nested lists.

When loading a persisted `Explanation` using the `from_json` class method, the persisted `Explanation` is again validated and coerced to the expected schema type.
A particulary useful consequence of this is that `numpy` arrays, which would be serialized to nested lists in `json` would be coerced to become `numpy` arrays when deserialized again so we can ensure that in Python you always have arrays where you expect them.

# Discussion points
1. Since we recognize that schemas will be iterated upon and `Explanation` objects can be persisted, it would be wise to include the version of `alibi` in the metadata of every `Explanation`.
   When reading a persisted `Explanation` we can read this field and ensure that the version of `alibi` being run is the same as the version used to produce the `Explanation`.
2. The usage of `Union` on the `data` field in `api.schemas.ExplanationModel` is not ideal.
   At runtime `pydantic` will attempt to match the returned `Explanation` to any of these schemas and pick the first one that matches.
   If there isn't a match (due to developer error), the error raised may be unrelated to the actual expcted schema (e.g. expect to satisfy the `ALEData` schema, but it fails, so `pydantic` tries all the others, which also fail with error messages).
   Even more worrying is the default behaviour of `pydantic` (perhaps this can be altered?) of matching an object to a partially satisfied schema and dropping the missing fields.
   This actually happened when a `TreeSHAP` explanation was matched to `KernelSHAP` as the fields of the latter are a subset of the former.
   I had to reverse the order of `TreeSHAP` and `KernelSHAP` declarations in the union to ensure that the more specific type comes first so it's matched first.
   A solution would be to instead have a separate `ExplanationModel` for each type of `Explainer` and do dynamic dispatch at runtime in `Explanation` based on the `meta.name` attribute.
   More generally, maybe this is another reason for thinking of moving away from an all-encompassing `Explanation` object towards more granularly defined objects, e.g. `AnchorExplanation` etc. as then there would be one class and one data model definition per explainer/explanation type.
   This would also be helpful in writing explanation specific methods in the future.
3. The validation of the return type according to the schema is enforced at runtime (Python's type checking is not powerful enough to validate the correct types with a static typechecker). This means that if we haven't modelled all possible return types of an `explain` call appropriately and haven't exercised all code branches in tests, the user may find the corner cases where the validation fails and throws an error.
   The solution is probably for developers of `alibi` to avoid introducing different return types in different branches of the code and make sure all code branches constructing `Explanation` objects are tested.
4. Exposing the `meta` and `data` attributes of `Explanation` via calls to `self._explanation_model.meta.dict()` duplicates the data due to converting the `pydantic` model to a `dict`.
   There should be a better way of doing this. Maybe turn `meta` and `data` into `property` attributes and call the `dict()` methods on demand?
5. The dictionaries in `api/defaults.py` so far have been used as a loose "schema" and as a factory of generating default dictionaries to be populated by data.
   With the schema definition moving to `api/schemas.py` I would propose moving the contents of `api/defaults.py` to each explainer specific module and turning them into factory functions.
   The idea here would be to use these as convenient factories for `alibi` developers when initializing default return data types.
   Note: it may be possible to re-use the `api/schemas.py` for this purpose, but it needs more investigating. Also it's trickier with some types (e.g. you may want to initialize `None` for an `Array` type as it's easier to create a new array than fill in an existing one, but the schema may not allow a `None` to be returned in the end.)
6. Whilst the schemas are primarily a developer tool, it would be nice if we could leverage this single source of truth as documentation for users as well. The alternative is to maintain separate documentation which may go out of sync.
7. The `Array` custom type allows us to refine the constraints on `numpy` arrays in the schemas, however the current implementation is a bit of a hack. The right implementation would be to inherit from `typing.Generic` and use two type variables (`typing.TypeVar`) - one for the `dtype` one for the `shape`. However, note that `Array[float, (2,3)]` has one type variable (`float`) and one concrete variable (the shape) which would fail in such an approach. So instead we opt for the `__class_getitem__` approach which is more flexible (although `api/schemas.py` doesn't pass the static typecheck as a result...).
8. We may need some more type aliases for `Array`-like types in the future, e.g. there is no way to define an `Array` of arbitrary dimensionality but with a fixed size of the first dimension.
9. We may need a type aliases for numeric types, e.g. `Numeric=Union[int, float]` to avoid some of the `Array[Any]` declarations.
   However, for consistency we need to decide if we should coerce any `float` type into the `Python` `float`. For example, there are `numpy.float64` etc. types, and perhaps more importantly, returns from `tensorflow` models will likely be `np.float32` - with a type declaration including only `float` these will always be coerced to 64bit.
10. One nice feature of `pydantic` is that given a model it can generate a `jsonschema`. This is not possible out of the box with custom types like `Array`, but it seems there may be ways to define extensions for this.

### DIFF
--- a/alibi/api/defaults.py
+++ b/alibi/api/defaults.py
@@ -3,63 +3,6 @@ This module defines the default metadata and data dictionaries for each explanat
 Note that the "name" field is automatically populated upon initialization of the corresponding
 Explainer class.
 """
-from alibi.api.types import Array
-from pydantic import BaseModel, conlist, confloat
-from typing import Any, List, Optional, Union
-from typing_extensions import Literal
-
-import numpy as np
-
-ModelType = Literal['blackbox', 'whitebox']
-ExpType = Literal['local', 'global']
-
-
-class AlibiBaseModel(BaseModel):
-    class Config:
-        json_encoders = {
-            np.ndarray: lambda x: numpy_encoder(x)
-        }
-
-
-class DefaultMeta(AlibiBaseModel):
-    name: Optional[str] = None
-    type: List[ModelType] = []
-    explanations: List[ExpType] = []
-    params: dict = {}
-
-
-class AnchorDataRawExamples(AlibiBaseModel):
-    covered_true: Array[Any, (-1, -1)]
-    covered_false: Array[Any, (-1, -1)]
-    uncovered_true: Array[Any, (-1, -1)]
-    uncovered_false: Array[Any, (-1, -1)]
-
-
-class AnchorDataRaw(AlibiBaseModel):
-    feature: List[int]
-    mean: List[float]
-    precision: List[confloat(ge=0.0, le=1.0)]
-    coverage: List[confloat(ge=0.0, le=1.0)]
-    examples: conlist(AnchorDataRawExamples, min_items=2, max_items=2)
-    all_precision: float
-    num_preds: int
-    success: bool
-    names: List[str]
-    prediction: int
-    instance: Array[Any, (-1,)]
-
-
-class AnchorData(AlibiBaseModel):
-    anchor: List[str]
-    precision: float
-    coverage: float
-    raw: AnchorDataRaw
-
-
-class ExplanationModel(AlibiBaseModel):
-    meta: DefaultMeta
-    data: Union[AnchorData]
-
 
 # Anchors
 DEFAULT_META_ANCHOR = {"name": None,
@@ -280,27 +223,3 @@ DEFAULT_DATA_INTGRAD = {
 """
 Default IntegratedGradients data.
 """
-
-
-def numpy_encoder(obj: Any) -> Any:
-    if isinstance(
-            obj,
-            (
-                    np.int_,
-                    np.intc,
-                    np.intp,
-                    np.int8,
-                    np.int16,
-                    np.int32,
-                    np.int64,
-                    np.uint8,
-                    np.uint16,
-                    np.uint32,
-                    np.uint64,
-            ),
-    ):
-        return int(obj)
-    elif isinstance(obj, (np.float_, np.float16, np.float32, np.float64)):
-        return float(obj)
-    elif isinstance(obj, (np.ndarray,)):
-        return obj.tolist()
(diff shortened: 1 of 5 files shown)


jklaise:
Following a discussion it was decided that for the next approach we will try to move towards returning the `pydantic` objects directly to the user and dispense with any custom `Explanation` objects which would make many of the points above obsolete. This new approach will be explored in a separate PR.

PR #349, opened 2021-02-10
Upper bound dependencies and dependabot support

jklaise:
This PR introduces upper bounds on both `alibi` dependencies and dev dependencies. It also includes `dependabot` integration which would notify us of new versions of dependencies via PRs daily.

The idea is that we don't want newly released versions of our dependencies to break our library (which has happened in the past), but we also need a way to test that the library keeps working with new released of dependencies. The proposed workflow is to bound the current dependencies and utilize `dependabot` whenever new versions are available. `dependabot` PRs will trigger a CI run with the new versions and, assuming this passes, we can merge the PRs bumping the upper bound (and release a new `alibi` version soon after so installation from PyPI supports latest dependencies).

For specifying the current upper bounds, I chose somewhat arbitrarily whether to follow major or minor versions based on my experience on previous breakages with specific libraries.

I also took the opportunity to simplify our various requirements files to only two - `dev.txt` and `docs.txt`. I though about switching to `setup.cfg` or `pyproject.toml` to have all dependencies in one place, but because we use readthedocs.org it is still required to have a separate file listing those requirements (`docs.txt`). 

If we agree with this strategy I will make a similar PR for `alibi-detect`.


jklaise:
I just noticed one thing that is not working well with `dependabot` + CI - using upper bounds for new versions may not trigger the CI to use the actual latest version of a dependency as the previous one would still be cached. This could lead to a dangerous scenario where CI tests pass but actually the new version was not used. This happened here where I know that the newer version breaks the documentation build: https://github.com/SeldonIO/alibi/pull/346 . We would need a solution to force no re-using cache on CI (but this could be unnecessarily wasteful for the vast majority of build) or switch to exact dependencies for `dev.txt` and `docs.txt` (but this still leaves the issue unsolved with library dependencies...).

RafalSkolasinski:
Few thing to consider:
- too strict upper bound may become inconvenient for users when later they'd want to use given alibi version with newer version of dependency (newer version that would work fine but at the point of release it wasn't nor released so upper bound is lower version)
- too strict lower bound (this usually should be put to the last LTS version or the version that introduced the feature one rely on - in kwant we used to put lowest bound to versions that ships with Stable Debian)

jklaise:
@RafalSkolasinski pointed out 2 things:

- lower bounds in `setup.py` currently just reflect the versions when we started development not the actual supported bounds, we should just remove them or find what actual good lower bounds are
- upper bounding too strictly may result in situations where people are happy with older `alibi` version but can't use them with newer versions of dependencies; this can be problematic and also we expect the vast majority of new releases to not break `alibi`, so we could be artifically restricting the upper bound too much.

jklaise:
@RafalSkolasinski I relaxed the dependency bounds, basing lower bounds in `setup.py` on Debian Latest (where available), and upper bounds on major version changes, unless the deps are on 0.x in which case it's on minor version changes. Same story for the `dev.txt` and `docs.txt` dependencies (less important as not related to what users can run with the library, but we still want to monitor new versions of these to take advantage of new features but not break our builds/docs).

.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+updates:
+- package-ecosystem: pip
+  directory: "/"
```
RafalSkolasinski, line 4: Wonder if wouldn't need to create another entry for dependabot to track requirements files inside the `requirements` directory.

.github/workflows/ci.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         run: |
           sudo apt-get install pandoc
           pip install --upgrade pip setuptools
-          pip install -r requirements/requirements_all.txt
+          pip install -r requirements/dev.txt -r requirements/docs.txt
```
RafalSkolasinski, line 34: Note for later could be to consider having the `Build documentation` step in separate build / pipeline (whatever github actions naming convention is) so it uses its own dependency. 

Just conscious that providing more than one `-r` argument may lead to nondeterministic dependency resolution.

RafalSkolasinski (approved):
Looks good, I think we should merge it!

Things to consider in the future: if the `requirements.txt` files are used only for development and CI testing (99.9% chance that's the case) we could consider them being exact locks so that version installed both for CI and tests and development are more deterministic. 

Anyway, that shouldn't stop this from merging.

.github/dependabot.yml
jklaise: Based on my experiment it found the subdirectories automatically.

.github/workflows/ci.yml
jklaise: This makes sense, probably one for a future PR.

PR #353, opened 2021-02-15
Docs build improvements

jklaise:
This PR does 2 things:

- Split out the docs build step in CI as it doesn't have to be done for every Python version and also separates the requirements from the test environment more cleanly
- Add a config file for readthedocs so the builds are configured directly in the repo instead of via their console


jklaise:
Added a version guard around `typing_extensions` module (only necessary for Python < 3.8). This should've been done as part of #349 but it is actually pulled in by some 3rd party dependency on Python 3.8 hence we didn't see an error.

PR #373, opened 2021-03-25
Fix deselected tests

### LINKED ISSUE #371: AnchorTabular tests deselected
Just noticed that most tests are not being run, need to investigate what has changed to cause tests to be deselected.

jklaise:
Resolves #371 . Also dropped the minimum `coverage` expected from 0.05 to 0.01 as the coverage calculation fix in #337 leads to smaller coverages on the adult dataset when multiple features are in the anchor.


jklaise:
Updated the test to allow for 10% of runs not resulting in a valid anchor. This is needed as the algorithm is stochastic and unless the instance tested is very "well-behaved" some runs will fail to find a valid anchor.

jklaise:
Had to drop the threshold to 80%, also reduced number of runs from 10 to 5, so expect 4/5 to result in a valid anchor.

PR #376, opened 2021-03-29
Make shap an extra dependency

jklaise:
This PR is meant to (temporarily) make `shap` an optional dependency. The reason is because the build process of `shap` currently is not optimal and will result in `RuntimeError` when packages have a different runtime dependency on `numpy` than the build dependency of `numpy` used for `shap`.

Concretely, `shap` on linux is built with the latest available version of `numpy` (currently `1.20`), whilst `tensorflow` has a `numpy` pin of `1.19.5`. This means that `pip install alibi` results in a broken installation as `shap` is still built against `1.20`. As far as I've looked, it is not possible to specify a different version of `numpy` to build `shap` against in a "one-pass" install of `alibi` (e.g. if `numpy` is not present, shap will be built against `1.20` regardless, if `numpy` is present it will be built against that version).

As a result, this PR proposes installing `alibi` with `shap` support in two steps:
```bash
pip install alibi
pip install alibi[shap]
```
I have added the above instructions to the README as well as at the top of the shap-specific notebooks.

This workaround should be resolved once `shap` specifies `oldest-supported-numpy` as a build dependency instead. For more information see here: https://github.com/slundberg/shap/pull/1802

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,12 @@ To take advantage of distributed computation of explanations, install `alibi` wi
 pip install alibi[ray]
 ```
 
+For SHAP support, install `alibi` as follows:
+```bash
+pip install alibi
+pip install alibi[shap]
+```
+
 The alibi explanation API takes inspiration from `scikit-learn`, consisting of distinct initialize,
 fit and explain steps. We will use the [AnchorTabular](https://docs.seldon.io/projects/alibi/en/latest/methods/Anchors.html)
 explainer to illustrate the API:
--- a/README.md
+++ b/README.md
@@ -86,8 +86,7 @@ pip install alibi[ray]
 
 For SHAP support, install `alibi` as follows:
 ```bash
-pip install alibi
-pip install alibi[shap]
+pip install alibi && pip install alibi[shap]
 ```
 
 The alibi explanation API takes inspiration from `scikit-learn`, consisting of distinct initialize,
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -84,6 +84,12 @@ To take advantage of distributed computation of explanations, install `alibi` wi
 pip install alibi[ray]
 ```
 
+For SHAP support, install `alibi` as follows:
+```bash
+pip install alibi
+pip install alibi[shap]
```
RafalSkolasinski, line 90: May be wrong here but, isn't `pip install alibi[shap]` installing alibi + alibi with shap dependencies already?

RafalSkolasinski (commented):
Looks okay but what if someone tries to import something from shap-dependent parts of library without shap being installed?
`alibi/explainers/shap_wrappers.py` ?

I think it will fail in quite an ugly way -> maybe worth to have some logic that do `try ... except ` in `shap` imports that in case shap is not present raise a valid warning / information?

README.md
jklaise: Yes it would, but the one-command results in the broken installation because `shap` is built against the newest `numpy` if no `numpy` is present in the environment...

jklaise:
> Looks okay but what if someone tries to import something from shap-dependent parts of library without shap being installed?
> `alibi/explainers/shap_wrappers.py` ?
> 
> I think it will fail in quite an ugly way -> maybe worth to have some logic that do `try ... except ` in `shap` imports that in case shap is not present raise a valid warning / information?

Good point, forgot to handle this completely as it would fail immediately on `alibi import` since it imports the `shap` specific explainers too...

Issue #388, opened 2021-04-16
TensorFlow get_session deprecation


jimbudarz:
Hi, I've been seeing a warning message when running cfproto:

`WARNING:tensorflow:From <venv>/lib/python3.7/site-packages/alibi/utils/tf.py:26: The name tf.keras.backend.get_session is deprecated. Please use tf.compat.v1.keras.backend.get_session instead.`

I believe it has to do with TF2.0 compatibility and this `get_session` method being from TF1. 
Can it be resolved by just replacing all instances of `tf.keras.backend.get_session` with `tf.compat.v1.keras.backend.get_session`?

jklaise:
Good catch, I think indeed it should be enough to switch to the `compat.v1` package.

jimbudarz:
Great- I'll submit a PR once I can get a testing environment up and running for the tests.

jklaise:
Closed via #390 .

PR #390, opened 2021-04-16
Changed keras reference to v1 compatibility module

jimbudarz:
Quick fix for issue #388

### DIFF
--- a/alibi/utils/tf.py
+++ b/alibi/utils/tf.py
@@ -23,7 +23,7 @@ def _check_keras_or_tf(predict_fn: Union[Callable, tf.keras.Model, 'keras.Model'
         Tuple of boolean values indicating whether the prediction function is a model or a black-box function and if
         it is a tf.keras or keras model, also returns the associated session (or a new one if the model is black-box).
     """
-    tfsess = tf.keras.backend.get_session()
+    tfsess = tf.compat.v1.keras.backend.get_session()
 
     try:
         # workaround to suppress keras backend message, see https://github.com/keras-team/keras/issues/1406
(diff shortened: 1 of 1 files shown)


jimbudarz:
I saw errors in testing around Integrated Gradients, but they already existed on the branch without these changes.

jklaise:
Thanks! What errors did you see locally?

jimbudarz:
@jklaise They were all like this:
```
_______________________________________ test_integrated_gradients_regression[1-riemann_middle-ffn_model0] _______________________________________
  File "/home/username/github/alibi/alibi/explainers/tests/test_integrated_gradients.py", line 452, in test_integrated_gradients_regression
    target=None)
  File "/home/username/github/alibi/alibi/explainers/integrated_gradients.py", line 508, in explain
    nb_samples)
  File "/home/username/github/alibi/alibi/explainers/integrated_gradients.py", line 627, in _compute_attributions_list_input
    for path in paths_ds:
  File "/home/username/venvs/alibi/lib/python3.6/site-packages/tensorflow/python/data/ops/dataset_ops.py", line 2776, in __iter__
    return iter(self._dataset)
  File "/home/username/venvs/alibi/lib/python3.6/site-packages/tensorflow/python/data/ops/dataset_ops.py", line 424, in __iter__
    raise RuntimeError("__iter__() is only supported inside of tf.function "
RuntimeError: __iter__() is only supported inside of tf.function or when eager execution is enabled.
--------------------------------------------------------------- Captured log call ---------------------------------------------------------------
WARNING  alibi.explainers.integrated_gradients:integrated_gradients.py:466 It looks like you are passing a model with a scalar output and target is set to `None`.If your model is a regression model this will produce correct attributions. If your model is a classification model, targets for each datapoint must be defined. Not defining the target may lead to incorrect values for the attributions.Targets can be either the true classes or the classes predicted by the model.
```

jklaise:
Ahh, this is because we actually need to run pytest in two passes due to having some tensorflow 1.x code in the codebase which needs eager mode switched off, but once you do that there is no way back. It's in our CI script here: https://github.com/SeldonIO/alibi/blob/658d7772d63685ec9634a0dbc50f39fcc89b093f/.github/workflows/ci.yml#L50-L51 I'll update the contributing.md file to reflect this.

Issue #377, opened 2021-03-31
Issue with the shape input to ig.explain


sbecon:
I am using the Integrated gradients for text classification on the IMDB dataset [notebook]( https://docs.seldon.io/projects/alibi/en/latest/examples/integrated_gradients_imdb.html), and on the ig.explain command, for my own model, I have submitted an object in the first argument that is TensorShape([128,512]). Because I am using a regression model, I have typed in None for baselines. However, I receive this error:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-115-af7b05840e2d> in <module>
      1 explanation = ig.explain(onebatch[0],
      2                          baselines=None,
----> 3                          target=None)
      4 
      5 attributions = explanation.attributions

/opt/conda/lib/python3.7/site-packages/alibi/explainers/integrated_gradients.py in explain(self, X, baselines, target)
    507             x, baseline = X[i], baselines[i]
    508             # format and check baselines
--> 509             baseline = _format_input_baseline(x, baseline)
    510             baselines[i] = baseline
    511 

/opt/conda/lib/python3.7/site-packages/alibi/explainers/integrated_gradients.py in _format_input_baseline(X, baselines)
    284     """
    285     if baselines is None:
--> 286         bls = np.zeros(X.shape).astype(X.dtype)
    287     elif isinstance(baselines, int) or isinstance(baselines, float):
    288         bls = np.full(X.shape, baselines).astype(X.dtype)

TypeError: data type not understood
```

If, instead, I input baselines to be baselines=tf.zeros([128, 512], tf.int32), the error is 

```ValueError: baselines must be `int`, `float`, `np.ndarray` or `None`. Found <class 'tensorflow.python.framework.ops.EagerTensor'>```

If I return baselines=None, and instead of sending in a TensorShape, I send in a np.array, I receive a different error: 

```---------------------------------------------------------------------------
InvalidArgumentError                      Traceback (most recent call last)
/opt/conda/lib/python3.7/site-packages/tensorflow_core/python/eager/context.py in execution_mode(mode)
   1896     ctx.executor = executor_new
-> 1897     yield
   1898   finally:

/opt/conda/lib/python3.7/site-packages/tensorflow_core/python/data/ops/iterator_ops.py in _next_internal(self)
    658             output_types=self._flat_output_types,
--> 659             output_shapes=self._flat_output_shapes)
    660 

[...]

PR #362, opened 2021-03-12
Intgrads subclassed models

gipster:
Support for tensorflow and keras models with no explicit inputs (subclassed models).

In the present code base, tensor inputs are converted to lists at explain time in order to handle models with multiple inputs. If a model has a single input, the tensor input will be converted to a list with one element. 

List input are handled well by models with an explicit input layer (functional and sequential models). When no explicit input layer is present (in subclassed models), the input must be in the format expected by the the first layer.

In order to support subclassed models, in this pull request
* The attributions are calculate with two different functions `_calculate_attributions_list_input` and `_calculate_attributions_tensor_input` for list inputs and tensor inputs, respectively. 
* For subclassed models with no explicit input, the inputs types and the output shapes of the models, which are needed to calculate the attributions, are inferred at explain time with a single mock call of the model.

Issue #343, opened 2021-02-06
Integrated Gradient for Multi-Task Models


senjed:
How can the integrated gradient explainer be used for a model with multiple outputs (e.g. a multi-task classification model where each output can have few possible values)? Can you please provide a working example? How can one choose to get attributions with respect to each of the outputs?

gipster:
At the moment there's no option to select an output within the IntegratedGradient class, but If you have a keras or tf.keras model with multiple outputs, you can do something like that

```
from tensorflow.keras.models import Model 

explanations = []
for out in model.output:
    model_2 = Model(inputs=model.input, outputs=out)
    ig  = IntegratedGradients(model_2)
    explanation = ig.explain(X)
    explanations.append(explanation)
```

It's good to keep this issue open to understand whether it is convenient to add an option that allow to pass an output as a parameter.

senjed:
I am getting a Graph disconnected error when I try to do this.

jklaise:
@senjed could you provide a summary view of what your model looks like (i.e. using the `summary()` method) for the original and the new model?

jklaise:
We should have an example of this.

nouna99:
Hello;
I have the same issue: I need to apply the Integrated gradients to a multi-task model with two inputs and two outputs. but I'm getting a Graph disconnected error when I try to create a loop .:
explanations = []
out= new_model.output[0]:
inp=new_model.inputs[0]:
    model_task = Model(inp, out)
    ig  = IntegratedGradients(model_task)
    explanation = ig.explain(Z)
    explanations.append(explanation)

jklaise:
@nouna99 could you provide a short, self-contained example of the multi-task model with the error? This will help us debug the issue.

nouna99:
new_model=tf.keras.models.load_model(my_wd, custom_objects={'f1_score': f1_score, 'precision' : precision, 'recall' : recall, 'AdamWeightDecay' : optimizer})
new_model.summary()
![image](https://user-images.githubusercontent.com/35919412/134311880-42ad3ce8-13e2-499c-a23f-d70328a0c860.png)

from tensorflow.keras.models import Model
predictions = pred_class
explanations = []
for out in new_model.output:
  for inp in new_model.inputs:
    model_task = Model(inputs=inp, outputs=out)
    ig  = IntegratedGradients(model_task)
    explanation = ig.explain(X_test_sample)
    explanations.append(explanation)

---------------------------------------------------------------------------
[...]

PR #393, opened 2021-04-22
Integrated gradients for transformers models

gipster:
# Integrated gradients for transformers models
Transformers models are widely used for tasks such as NLP. The aim of this pull request is to extend the integrated gradients method in [alibi](https://github.com/SeldonIO/alibi/blob/master/alibi/explainers/integrated_gradients.py) to the pretrained tensorflow models provided by the [transformers](https://github.com/huggingface/transformers) library.

## Usage of alibi with transformers library
With a minimal change in the present alibi code, it is possible to obtain the integrated gradients attributions for the input embeddings of a transformer model. A minimal example is provided below  (see issues #362 and #377 )

```python
import tensorflow as tf
import numpy as np
from transformers import TFBertModel
from alibi.explainers import IntegratedGradients

# load the pretrained bert model
modelBert = TFBertModel.from_pretrained('bert-base-uncased')

# define the custom model using bert as a first layer
max_len=512
input_ids_in = tf.keras.layers.Input(shape=(max_len,), 
                                     name='input_tokens', 
                                     dtype=tf.int32)
attention_masks_in = tf.keras.layers.Input(shape=(max_len,), 
                                           name='attention_mask_tokens', 
                                           dtype=tf.int32)
X = modelBert([input_ids_in, attention_masks_in])[0]
# adding a flatten layer as a mock layer 
X = tf.keras.layers.Flatten()(X)
X = tf.keras.layers.Dense(1)(X)
frozenModelOut = tf.keras.Model(inputs=[input_ids_in, attention_masks_in], 
                                outputs=X)
frozenModelOut.layers[2].trainable = False

# extract the input embedding layer from bert model
iel = frozenModelOut.layers[2].bert.embeddings

# define random inputs for testing
x = [np.random.randint(1000, size=(2, max_len)).astype('int32'),
    np.ones((2, max_len)).astype('int32')]

# calculate integrated gradients
ig  = IntegratedGradients(frozenModelOut, layer=iel, internal_batch_size=5)
expl = ig.explain(x)                          
```

The only change required for the examples above to run is to allow the assignment of `layer_num = None` in the metadata output. This is necessary because the extracted layer doesn't have an explicit index in the model. This change is included in the present version of the PR.

## Inner and output transformers layer
In order to use the method above to calculate the attributions for the inner and the output layer of the transformer, it is necessary to have a method to extract those layers from the pretrained bert model, in a way that is analogous to the one described above for the input embeddings. This is currently under investigation. Preliminary tests show that

* The pretrained models in [transformers](https://github.com/huggingface/transformers) do not seem to have a method to extract the inner layers analogous to `get_input_embeddings()`.

* The method `get_output_embeddings()` returns None. 

The goal is to find a systematic way to extract the inner layers from a tensorflow [transformers](https://github.com/huggingface/transformers) model. 

## Work around: wrapping bert in a tf.keras layer
A work around to the problem above is to wrap the bert model into a keras or tf.keras layer when part of a bigger model. 
In this case, the use of the `IntegratedGradients` class is straightforward.
Let's assume we want to use the output of the bert model or the output of one of its internal layers as a starting point for a custom model such as the one considered in the example above. The integrated gradients attributions with respect to the output of the bert layer can be calculated as

```python
import tensorflow as tf
import numpy as np
from transformers import TFBertModel, BertConfig
from alibi.explainers import IntegratedGradients

# load the pretrained bert model and configure it to return all hidden states
config = BertConfig.from_pretrained("bert-base-uncased", output_hidden_states=True)
modelBert = TFBertModel.from_pretrained('bert-base-uncased', config=config)

# wrap the model into a tf.keras layer
class BertLayer(tf.keras.layers.Layer):
    def __init__(self, modelBert, **kwargs):
        super(BertLayer, self).__init__()
        self.modelBert = modelBert

    def call(self, inputs, output_nb=5):
        return self.modelBert(inputs)[2][output_nb]
bertLayer = BertLayer(modelBert, trainable=False)

# define custom model
max_len=512
input_ids_in = tf.keras.layers.Input(shape=(max_len,), 
                                     name='input_tokens', 
                                     dtype=tf.int32)
attention_masks_in = tf.keras.layers.Input(shape=(max_len,), 
                                           name='attention_mask_tokens', 
                                           dtype=tf.int32)
X = bertLayer([input_ids_in, attention_masks_in])
# adding a flatten layer as a mock layer 
X = tf.keras.layers.Flatten()(X)
X = tf.keras.layers.Dense(1)(X)
frozenModelOut = tf.keras.Model(inputs=[input_ids_in, attention_masks_in], 
                                outputs=X)

# extract the input embedding layer from bert model
el = frozenModelOut.layers[2]

# define random inputs for testing
x = [np.random.randint(1000, size=(2, max_len)).astype('int32'),
    np.ones((2, max_len)).astype('int32')]

# calculate integrated gradients
ig  = IntegratedGradients(frozenModelOut, layer=el, internal_batch_size=5)
expl = ig.explain(x)                          
```

Note that the pretrained bert model must be configured in order to return all hidden states, and the layer must specify in advance which hidden state must be return. This work around is not ideal, but it allows to calculate the attributions of the inner and output embeddings in a limited number of use cases. 

A more permanent solution is under investigation.


jklaise:
I think this is how we can proceed.

1. Work on a real-world example with a transformer model explaining some real text data and showcase the 2 ways described above how this can be achieved.
2. The proposed handling of `layer_num=None` would make it much harder to handle serialization. I am currently using `layer_num` to extract the correct layer from `model` when deserializing, of course with this change it won't be possible. It also seems unlikely to be able to specify some layer identifier to automatically extract the layer from the model because the Bert example shows that you need to call a custom function (`get_input_embeddings`). We also don't want to save the whole model, it may be possible to save just the layer, but this then would be a separate copy of the layer unrelated to the layer in re-loaded model. One (not ideal) option is to extend the load API from `load_explainer('path', predictor)` to `load_explainer('path', predictor, **kwargs)` and require `layers` to be a `kwarg` specified by the user. Another option may be to provide a new `layer_extractor: Callable` `kwarg` so the user specifies the function used to extract the layer which we then can attempt to serialize, but this is also messy.
3. On a separate note, work on a PR proposal for automatically handling multi-output models (see https://github.com/SeldonIO/alibi/issues/343). The idea here is to come up with a reasonable solution for handling multi-output models natively. This would involve, as a minimum:
   - Figure out the `explain` API, e.g. do we add outputs to be explained as a list of `int`s, i.e. `outputs=[0,1]`, do we allow just one output to be explained at a time or multiple?
   - Figure out how to implement this internally, do we need to make an auxiliary model for every output to be explained or is there some other way?
   - Figure out how to change the returned attributions, currently `exp.attributions` is `List[np.ndarray]` supporting multi-input models (but not extending to multi-layer). Will need to figure out how this changes with multi-outputs.

jklaise:
From a discussion:

2. The best scenario would be to find if there is an API to get layers by name, this would also solve the issue for explaining intermediate layers.
4. Related to multi-output models, the current implementation also assumes the outputs are 1 to 2-dimensional (for doing classification/regression), so currently the methods wouldn't work on higher-dimensional outputs. We would need to decide if we want to support this.

PR #394, opened 2021-04-23
WIP: Explainer serialization v1

jklaise:
Proposed implementation of initial support for explainer persistence.

# API
Given an explainer, it can be saved by calling `explainer.save(path)`, where `path` is a directory which is created if it doesn't exist. Alternatively:
```python
from alibi.saving import save_explainer
save_explainer(explainer, path)
```

To load an explainer:
```python
from alibi.saving import load_explainer
explainer = load_explainer('path', predictor=predictor)
```
Note that we require the user to pass the `predictor` specifically as we decided not to handle serialization of arbitrary white-box and black-box models.

Alternatively, one can call a `classmethod` as follows:
```python
from alibi.api.interfaces import Explainer
from alibi.explainers import AnchorTabular  # e.g.

explainer = Explainer.load('path', predictor=predictor)
explainer = AnchorTabular.load('path', predictor=predictor)
```

# Implementation
The saving and loading functionality is defined in a new module `alibi.saving`, tests are provided in `alibi.tests.test_saving`.

There is a save/load function per explainer type which is determined at runtime based on the `name` attribute in `meta`. The reason for not using `singledispatch` is because of circular references due to importing `save_explainer` into `alibi.api.interfaces` to create the bound method `save`.

The saved format is different depending on the explainer. All explainers save their `meta` attribute as `json` which is used to re-initialize a new (identical) explainer. Further, some explainers save additional binaries (e.g. `numpy` formats, `spacy` models, user supplied `Callable` functions).

Currently all explainers are loaded by simply creating identical ones with appropriate `init_kwargs` stored in `meta` and also calling `fit` if required (e.g. `AnchorTabular` stores the data as an attribute, so we can dump this and load back to call `fit` again, though this won't be possible with `KernelShap` and `TreeShap` (see below).

# Discussion
- Using `json` to save metadata is a bit tricky as we need to make sure types are deserialized properly, e.g. a list in `json` is sometimes a `tuple` in Python, a list of lists is probably a `np.ndarray` originally. Currently these are manual steps in each `_load_...` function. One option may be to use a Pydantic definition of the `meta` data which could attempt to parse it to the correct objects (though this may not always result in the expected object if there are several options). This suggests that using `dill` may still be preferable...
- Putting initialization kwargs into the `meta` field is messy both due to the polluted `__repr__` but also for bookeeping purposes. It may make sense to save these `init` and `fit` kwargs into separate private attributes used only for saving/loading. This is related but not identical to designing a "state" attribute which can be used to re-create an explainer without having to call `init` and/or `fit`.

# TODO
- `KernelShap` and `TreeShap` require some small refactorings to be provide state to be saved/loaded to be able to serialize these without access to training data. The `shap` library actually provides serialization although it is `pickle` based so unclear if beneficial.
- when saving, also save the `alibi` version used which can then be used to raise a warning if versions mismatch when loading
- `CEM`, `CounterFactual` and `CFProto` not supported yet as the implementation would change significantly after the Tensorflow 2.x rewrite.

### DIFF
--- a/alibi/api/interfaces.py
+++ b/alibi/api/interfaces.py
@@ -1,7 +1,7 @@
 import abc
 import json
-from collections import ChainMap
-from typing import Any
+from collections import ChainMap, defaultdict
+from typing import Any, List
 import logging
 from functools import partial
 import pprint
@@ -108,6 +108,30 @@ class Explainer(abc.ABC):
                 self.meta['params'].update([(key, data_dict[key])])
         else:
             self.meta.update(data_dict)
+
+    def _update_state(self, key: str, params: dict, skip: List[str] = None) -> None:
+        """
+        Create or update the internal `_state` dictionary with parameters.
+        Used for tracking user supplied arguments to `__init__` and `fit` as well as state after calling `fit`.
+        The `_state` is serialized and used for recreating explainers, either by calling the `__init__` and `fit`
+        methods with the kwargs taken from `_state['init_kwargs'], _state['fit_kwargs']` or by setting relevant
+        attributes to avoid calling e.g. a fit method (using `_state['post_fit']`).
+
+        Parameters
+        ----------
+        key
+            Dictionary key, e.g. `init_kwargs`, `fit_kwargs`, `post_fit`
+        params
+            Dictionary of parameters to track, usually a subset of `locals()` after a method call.
+        skip
+            List of strings of parameters to not save, e.g. `predictor`, `self`, `__class__` etc.
+        """
+        if not hasattr(self, '_state'):  # could be initialized as a class attribute to avoid check
+            self._state = defaultdict(dict)
+
+        for k, v in params.items():
+            if skip is None or k not in skip:
+                self._state[key][k] = v
 
 
 class FitMixin(abc.ABC):
--- a/alibi/api/interfaces.py
+++ b/alibi/api/interfaces.py
@@ -7,7 +7,8 @@ from functools import partial
 import pprint
 
 import attr
-import numpy as np
+
+from alibi.saving import load_explainer, save_explainer, PathLike, NumpyEncoder
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


jklaise:
Could investigate [jsonpickle](https://github.com/jsonpickle/jsonpickle) for transparently serializing e.g. numpy arrays that appear in metadata.

jklaise:
In the last commit added support for `KernelShap`, but this is a bit rough around the edges.
cu
`KernelShap` now has private `_state` dictionary which is a dict of dicts tracking state (which is updated by calling `self._update_state`). There are a few types of state tracked:
 - method kwargs - when calling `self._update_state` with user supplied kwargs at the beginning of a method
 - post-method attributes - when calling `self._update_state` with a selection of attributes at the end of a method
 - others - mostly kwargs of complex attributes (objects from 3rd party libraries) which are hard to serialize in full, so we re-create them

The logic in `_load_KernelShap` first uses `init_kwargs` to re-initialize an equivalent explainer, then loads `state.json` and `background_data.dill` to set the state as if `fit` was called, this requires custom logic to re-create a complex object similar to what `fit` does (could be factored out into a method to avoid duplication).

The current `_load_KernelShap` is rather brittle (custom logic, need to handle unfitted explainer case) and also the peppering of `_update_state` in the implementation is not ideal. Finally, use of `json` for state again makes deserializaiton tricky if `numpy` arrays are present.

jklaise:
When saving an explainer we should also save the library version currently used. Upon loading, we should check that the current version agrees with that of the saved one. The question is, should we just issue a warning or raise an error and refuse to load until the user installs the correct version? Which is likely to lead to less headache?

If we're handling state transparently then this is similar to using pickle - we don't serialize the class definition itself so when loading we rely on the one that's available at runtime from whatever version of alibi is being used and hope for the best. In contrast, using non-transparent serialization with e.g. dill which also saves the class definition allows potentially outdated class definitions to be loaded in newer versions of the library, can this cause issues down the line?

jklaise:
Suggestion from @cliveseldon: instead of dealing with state and re-initializing explainers, can try to use `dill` on the whole object whilst saving the other non-dillable components in a different way and setting the attributes to `None`, e.g.:

```python
def _save_AnchorText(explainer, path):
    explainer.nlp.to_disk(Path(path, 'nlp'))
    explainer.nlp = None
    explainer.predictor = None
    with open(Path(path, 'explainer.dill'), 'wb) as f:
        dill.dump(explainer, f, recurse=True)

def _load_AnchorText(path, predictor):
   nlp = spacy.load(Path(path, 'nlp'))
   with open(Path(path, 'explainer.dill'), 'rb') as f
        explainer = dill.load(f)
    explainer.nlp = nlp
    explainer.reset_predictor(predictor)  # call reset_predictor instead of directly setting the attribute as at deployment there may be more actions needed to be performed than just setting an attribute

This would make the saving/loading logic much cleaner and with less code and potential for errors.

This falls into the option of serializing the whole class definition, it remains to investigate how these "old" objects with separate class definitions would play with new alibi versions. We would still need to save a piece of metadata that tracks the `__version__` and emit a warning on mistmatch.
```

jklaise:
Closing in favour of #397.

PR #397, opened 2021-04-27
Serialization v1.1

jklaise:
Alternative to #394.

Instead of saving input kwargs and dealing with state for cases when we cannot call `fit`, this directly saves the explainer using `dill` whilst setting attributes we don't want to save (e.g. `predictor`) to `None`. The outputs are as follows:
```
save_dir/meta.dill   # explainer.meta in a dill format to avoid json validation on loading back in
save_dir/explainer.dill  # pickled explainer
save_dir/other_artefact_1.dill   # e.g. spacy model; some custom Callable passed into explainer etc.
```

Note that due to the usage of `dill` even with `recurse=True` the class code is not serialized (https://github.com/uqfoundation/dill/issues/128) so it's assumed that `alibi` will be present on whatever remote environment this is deserialized (e.g. this is true in an container running alibi-explain server).

I also set a `_version` class variable for each explainer (set on the base interface `alibi.api.interfaces.Explainer` at import time) and validate against it when loading, raising a warning if there's a mismatch.

Pros:
 - No explicit state management or kwargs bookkeeping in the implementations
 - Use dill instead of json so no validation burden

Cons:
- No human-readable metadata due to using `dill` over `json` (but probably worth it due to no `json` validation overhead)
- Have to set attributes to `None`, save explainer, then reset the attributes back. Should be covered by tests and no trickier than the custom functionality in #394.

TODO: 
 - `AnchorText` modifies it's state after `explain` call with some unpicklable `spacy` stuff so the test currently doesn't pass


alibi/api/interfaces.py
```diff
@@ -64,7 +66,7 @@ class Explainer(abc.ABC):
     """
     Base class for explainer algorithms
     """
-
+    _version: ClassVar = __version__
```
cliveseldon, line 69: Is one meant to set the underling Type of the ClassVar? e.g. ClassVar[str] ?

cliveseldon (approved):

alibi/api/interfaces.py
jklaise: Good catch, it should be `ClassVar[str]`

PR #417, opened 2021-05-19
Changed integral path definition for layer integrated gradients - WIP

gipster:
The present version of the `IntegratedGradients` class calculates the [integrated gradients](https://arxiv.org/abs/1703.01365) respect to an internal layer by defining the path along which the integral is calculated in input space. A more correct approach consists of defining the path in the layer input or output space. This PR implements the latter version. Note that the two definitions give the same results for input attributions.

Main Changes:
* ~The classes `ModelToLayer` and `ModelFromLayer` are introduced. The purpose of these classes is to split the model in two parts at the layer of interest.  `ModelToLayer` forwards the inputs on the layer space. The 'ModelToLayer' output serves as input to  `ModelFromLayer`. The gradients are calculated along a integral path defined in the input space of  `ModelFromLayer`.~ 
The functions `_run_forward_to_layer` and `_run_forward_from_layer` are introduced. Those functions contain decorators that allow to modify the layer's inputs and outputs during the forward pass. The layer's gradients are calculated in a similar way using the function `_gradients_layer`.  
* ~The code do not use the `_gradients_layer` function anymore. All gradients are calculated using the `_gradients_input` function.~
* It is possible now to choose whether to calculate the attributions for the layer's inputs or the layer's output. This choice is controlled by the `compute_layer_inputs_gradients` flag added to the `explain` arguments. When `False`, attributions are calculated respect to the layer's output, when `True` attributions are calculated respect to the layer's inputs.
* The code has been restructured to fit these changes.

Todo
* ~Implement layer's gradients for [huggingface transformers](https://github.com/huggingface/transformers) models.~
* ~Adding tests for multi-input layer's gradients.~
* Code cleaning.


gipster:
The layer gradients are calculated using forward functions with decorators that modify the original `call` method of the layer. 
The example below is a minimalistic self contained piece of code that demonstrate the behaviour of the decorator. The function `_run_forward_from_layer` below is essentially the same as the one defined in the `integrated_gradients` module. The layer's gradients are calculated  using essentially the same trick.

```python
import tensorflow as tf
import numpy as np

def _run_forward_from_layer(model,
                            layer,
                            orig_call,
                            orig_dummy_input,
                            x,
                            run_from_layer_inputs = False):
    """
    Executes a forward call from an internal layer of the model to the model output.
    Parameters
    ----------
    model
        Tensorflow or keras model.
    layer
        Starting layer for the forward call.
    orig_call
        Original `call` method of the layer.
    orig_dummy_input
        Dummy input needed to initiate the model forward call.
        The  layer's status is overwritten during the forward call.
    x
        Layer's inputs. The layer's status is overwritten with `x` during the forward call.
    run_from_layer_inputs
        If True, the forward pass starts from the layer's inputs, if False it starts from the layer's outputs.

    Returns
    -------
        Model's predictions for the given target.

    """
    def feed_layer(layer):
        """
        Overwrites the intermediate layer status with the precomputed values `x`.

        """
        def decorator(func):
            def wrapper(*args, **kwargs):
                # Store the result and the inputs of `layer.call` internally.
                if run_from_layer_inputs:
                    layer.inp = x
                    layer.result = func(*x, **kwargs)
                else:
                    layer.inp = args
                    layer.result = x
                # Return the result to continue with the forward pass.
                return layer.result

            return wrapper

        layer.call = decorator(layer.call)
        return layer

    feed_layer(layer)
    preds = model(orig_dummy_input)

    delattr(layer, 'inp')
    delattr(layer, 'result')
    layer.call = orig_call

    return preds

# Define model with all weights = 1
inputs = tf.keras.Input(shape=(16,))
out = tf.keras.layers.Dense(8, 
                            kernel_initializer=tf.keras.initializers.Ones(),
                            name='linear1')(inputs)
out = tf.keras.layers.Dense(1, 
                            kernel_initializer=tf.keras.initializers.Ones(), 
                            name='linear3')(out)
model = tf.keras.Model(inputs=inputs, outputs=out)

# We will consider the first layers which has input_shape = (16, ) and output_shape = (8, )
layer = model.layers[1]
orig_call = layer.call

# Define dummy model's inputs, layer's input and layer's outputs
dummy_input = np.zeros((1, 16))
x_input_layer_1 = np.ones((2, 16))
x_output_layer_1 = np.ones((3, 8))

# Forward to model's predictions from layer's inputs
preds_from_layer_inputs = _run_forward_from_layer(model, 
                                                  layer, 
                                                  orig_call,
                                                  dummy_input,
                                                  [tf.convert_to_tensor(x_input_layer_1)],
                                                  run_from_layer_inputs=True)

# Forward to model's predictions from layer's outputs
preds_from_layer_output = _run_forward_from_layer(model,
                                                  layer,
                                                  orig_call,
                                                  dummy_input,
                                                  x_output_layer_1,
                                                  run_from_layer_inputs=False)
```

alibi/explainers/integrated_gradients.py
```diff
@@ -197,9 +401,15 @@ def watch_layer(layer, tape):
...
-                layer.result = func(*args, **kwargs)
-                # From this point onwards, watch this tensor.
+                # Store the result and the input of `layer.call` internally.
+                if compute_layer_inputs_gradients:
+                    layer.inp = x
+                    layer.result = func(*x, **kwargs)
+                else:
+                    layer.inp = args
+                    layer.result = x
+                # From this point onwards, watch this tensors.
```
jklaise, line 411: typo `this --> these`

alibi/explainers/integrated_gradients.py
```diff
@@ -578,25 +852,30 @@ def _compute_attributions_list_input(self,
...
+        input_dtypes = [xx.dtype for xx in X]
+
+        if (len(self.model.output_shape) == 1
+            or self.model.output_shape[-1] == 1) \
+                and target is None:
+            logger.warning("It looks like you are passing a model with a scalar output and target is set to `None`."
+                           "If your model is a regression model this will produce correct attributions. If your model "
+                           "is a classification model, targets for each datapoint must be defined. "
+                           "Not defining the target may lead to incorrect values for the attributions."
+                           "Targets can be either the true classes or the classes predicted by the model.")
```
jklaise, line 873: Can this check be factored out and performed earlier (ditto for the other tensor attribution function)? Two reasons:
- it's repeated in both `tensor` and `list` attribution functions
- we can do the validation before computing attributions as the model is called and `output_shape` is know, so this check should not be inside the attribution calculation code

alibi/explainers/integrated_gradients.py
```diff
@@ -108,24 +147,184 @@ def _run_forward(model: Union[tf.keras.models.Model],
...
-                raise NotImplementedError
-        else:
-            raise ValueError("target cannot be `None` if `model` output dimensions > 1")
-        return ps
+    return preds
+
+
+def _run_forward_from_layer(model: tf.keras.models.Model,
+                            layer: tf.keras.layers.Layer,
+                            orig_call: Callable,
```
jklaise, line 159: btw, is there a reason to define `orig_call` outside this function instead of just getting it inside via `layer.call` ? Is there a risk of `layer.call` already being overwritten somehow?

alibi/explainers/tests/test_integrated_gradients.py
```diff
@@ -420,13 +425,14 @@ def test_integrated_gradients_binary_classification_layer(ffn_model, method, lay
...
+                              compute_layer_inputs_gradients=layer_inputs_attributions)
 
     assert isinstance(explanations, Explanation)
-    if layer is not None:
+    if layer is not None and not layer_inputs_attributions:
         layer_out = layer(X_test).numpy()
         assert explanations['data']['attributions'][0].shape == layer_out.shape
-    else:
+    elif layer is None:
         assert explanations['data']['attributions'][0].shape == X_test.shape
```
jklaise, line 436: Looks like there are no assertions performed when `layer_input_attributions=True`, is there something we can add or is that not possible? E.g. call `layer(X_test)` and then check `layer.input_shape`?

jklaise (commented):
Nice work @gipster, looks good. Left mostly minor comments for improvement and a couple of questions (also check the notebook comments). Two extra things I'm keen to do in this PR:
- make the self-contained example in your comment a test - would be nice to test at least the `_run_forward_from_layer` function with both `run_from_layer_inputs=True/False` and check the dummy output corresponds to hard-coded model weights (all ones and bias zero)
- as I noted before, we should make sure our code cannot leave the model in a broken state, e.g. currently if an unsupported model is passed in (e.g. one that outputs tuples instead of arrays/tensors), then IG breaks but also the model layer is modified irrevocably. Ideally we validate that the model outputs are as expected and then bypass this problem in this scenario, but we may later on need to raise an exception and undo the layer patching.

examples/integrated_gradients_imdb.ipynb
```diff
@@ -19,7 +19,7 @@
   },
```
jklaise, line 19: <p>Can you upgrade TF to 2.5.0, rerun the notebook and check if you get <code>tf.int32</code>  warnings when calling <code>explain</code>  please?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/417/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

jklaise, line 19: <p>Can leave out <code>compute_layer_input_gradients=False</code>  as that is the default unless there's another reason to keep it explicit here?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/417/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

alibi/explainers/integrated_gradients.py
gipster: I think it's safer to fix the original layer call method at the very beginning, and stick to that to reset it. The functions may be called many times and in this way we avoid funny things to happen.

alibi/explainers/integrated_gradients.py
gipster: If the model doesn't have `inputs`, you have to call the model on a dummy input to the output shape. So the warning has to be repeated twice anyway in explain for list inputs and tensor inputs. I think it's cleaner to leave it inside the functions.

alibi/explainers/integrated_gradients.py
```diff
@@ -421,10 +632,15 @@ def __init__(self,
         self.method = method
         self.internal_batch_size = internal_batch_size
 
+        self._is_list: Optional[bool] = None
+        self._is_np: Optional[bool] = None
+        self.orig_dummy_input: Optional[Union[list, np.ndarray]] = None
```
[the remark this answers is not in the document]
gipster: They are used in 3 functions: `_compute_attributions_list_inputs`, `_compute_attributions_tensor_inputs` and `_compute_convergence_delta`. They have to be passed to each function if they are not class attributes.

alibi/explainers/integrated_gradients.py
```diff
@@ -108,24 +147,184 @@ def _run_forward(model: Union[tf.keras.models.Model],
...
+                    layer.result = x
+                # Return the result to continue with the forward pass.
+                return layer.result
+
+            return wrapper
+
+        layer.call = decorator(layer.call)
+        return layer
+
+    feed_layer(layer)
```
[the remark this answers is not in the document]
gipster: Ya, probably not

gipster:
> Nice work @gipster, looks good. Left mostly minor comments for improvement and a couple of questions (also check the notebook comments). Two extra things I'm keen to do in this PR:
> 
> * make the self-contained example in your comment a test - would be nice to test at least the `_run_forward_from_layer` function with both `run_from_layer_inputs=True/False` and check the dummy output corresponds to hard-coded model weights (all ones and bias zero)
> * as I noted before, we should make sure our code cannot leave the model in a broken state, e.g. currently if an unsupported model is passed in (e.g. one that outputs tuples instead of arrays/tensors), then IG breaks but also the model layer is modified irrevocably. Ideally we validate that the model outputs are as expected and then bypass this problem in this scenario, but we may later on need to raise an exception and undo the layer patching.

* I added the test for `_run_forward_from_layer` .
* I also added a `_validate_output` function and a global variable `_valid_outputs`, which is a list containing all the valid outputs types (at the moment only `tf.Tensor`). I'm not sure if that's the best way, but in this way if we decide to support other output types in the future should be sufficient to add it to the list.

examples/integrated_gradients_imdb.ipynb
gipster: I think it's better to keep it explicitly since it's an embedding layer and you cannot run it for inputs

alibi/explainers/integrated_gradients.py
```diff
@@ -116,15 +118,29 @@ def _sum_rows(inp):
...
 
-def _select_target(ps, ts):
-    if ts is not None:
-        if isinstance(ps, tf.Tensor):
-            ps = tf.linalg.diag_part(tf.gather(ps, ts, axis=1))
+def _select_target(preds: tf.Tensor,
+                   targets: Union[None, tf.Tensor, np.ndarray, list]) -> tf.Tensor:
+    """
+    Select the predictions corresponding to the targets if targets is not None.
+    Parameters
```
jklaise, line 125: Please put a blank line before `Parameters`

alibi/explainers/integrated_gradients.py
```diff
@@ -572,6 +589,35 @@ def _calculate_sum_int(batches: List[List[tf.Tensor]],
     return sum_int
 
 
+def _validate_output(model: tf.keras.Model,
+                     target: Optional[List[int]]) -> None:
+    """
+    Validates the model's output type and rises an error if the output type is not supported.
+    Parameters
```
jklaise, line 596: - Please put a blank line before `Parameters`
 - typo rises --> raises

alibi/explainers/integrated_gradients.py
```diff
@@ -572,6 +589,35 @@ def _calculate_sum_int(batches: List[List[tf.Tensor]],
...
+    target
+        Targets for which gradients are calculated
+
+    Returns
+    -------
+
+    """
+    if not model.output_shape or not any(isinstance(model.output_shape, t) for t in _valid_output_shape_type):
+        raise NotImplementedError(f"The model output_shape attribute must be in {_valid_output_shape_type}. "
+                                  f"Founded model.output_shape: {model.output_shape}")
```
jklaise, line 609: Typo founded --> found

alibi/explainers/integrated_gradients.py
jklaise: Just leaving a note that this has been refactored out as part of `_validate_output`.

jklaise (changes requested):
Just a couple of typos (see comments) and can you remove the captum notebook? Can merge afterwards.

jklaise (approved):

Issue #420, opened 2021-06-04
Validate categorical_names map for AnchorTabular


jklaise:
Incorrect entries can result in processing bugs, we should validate that the passed mapping is of type `Dict[int, List[str]]` and if not either raise an error or coerce + warn user. See also https://github.com/SeldonIO/alibi/issues/252#issuecomment-854898086

Also, update the type hints with more specific types.

jklaise:
A type hint has been added in #422 which should clear up any confusion. We're not doing extra validation currently due to overhead.

Issue #336, opened 2021-01-27
Support One Hot Encoding for AnchorTabular Explainers


FarrandTom:
Currently the AnchorTabular Explainer does not support the use of OHE variables. AnchorTabular does support the use of categorical variables encoded ordinally, but does not support OHE. 

Given how common an encoding scheme OHE is, and the fact that ordinal/label encoding is not suitable in all cases (there are some ML libraries where you cannot specify that you have ordinally/label encoded variables such as XGBoost and so they are simply treated as continuous numerical features which can introduce correlations that are not present in the true data) it would be great to extend the AnchorTabular explainer to support this case.

jklaise:
Implemented in #422.

PR #422, opened 2021-06-08
Extended AnchorTabular to one-hot encoding input.

RobertSamoilescu:
Extends AnchorTabular to one-hot encoding input. This solves the issue https://github.com/SeldonIO/alibi/issues/336

### Description
The categorical variable can be passed as one-hot encoded by setting the introduced flag `ohe=True`. The rest of the arguments (`feature_names: List[str], categorical_names: Dict[int, List[str]]` -- typing was add which solves the issue https://github.com/SeldonIO/alibi/issues/420) remain consistent with the previous version. For example, if the data contains the following `feature_names=['age', 'gender', 'race']` then the `categorical_names={1: ['female', 'male'], 2: ['asian', 'black', 'hispanic', 'white']}`. Note that the keys in the categorical_names are not offseted by the numbers of categories required for the one-hot encoding.

### Code changes

1.  `ohe` flag added to the constructor: `def __init__(self, predictor: Callable, feature_names: List[str], categorical_names: Dict[int, List[str]] = None, ohe: bool = False, seed: int = None) -> None:`
2. if the `ohe=True`, the predictor is expecting the categorical input features to be one-hot encoded, but the AnchorTabular works with label encoding. Thus we define a proxy for the predictor `ord_predictor = lambda x: predictor(ord_to_ohe(x, self.cat_vars_ord)[0])` which takes as input labels and transforms them to one-hot encodings before passing them to the original prediction function passed as argument in the constructor. Thus the new predictor used in AnchorTabular is defined by `self._predictor = self._transform_predictor(ord_predictor)`. Note that there has been a renaming from `self.predictor` to `self._predictor` as will be further explained. See the predictor.setter
3.  `train_data` passed as an argument to the `fit` method is transformed from one-hot encoding to labels to be compatible with AnchorTabular: ` train_data = ohe_to_ord(X_ohe=train_data, cat_vars_ohe=self.cat_vars_ohe)[0]` if `self.ohe == True`. Similar in the `expain` method, `X = ohe_to_ord(X_ohe=X.reshape(1, -1), cat_vars_ohe=self.cat_vars_ohe)[0].reshape(-1)`.
4. As the `predictor` was exposed to the user (see example [here](https://docs.seldon.io/projects/alibi/en/stable/examples/anchor_tabular_adult.html)), the newly defined self._predictor can not be used since it expects categorical data labeled encoded. Thus, we define another internal predictor `self._ohe_predictor = self._transform_ohe_predictor(predictor)`, which expects one-hot encoded features. Finally, the `predictor` is exposed to the user through the `@property predictor` which returns the appropriate `self._ohe_predictor` if `self.ohe == True` and `self._predictor` otherwise.
5. Similar changes to the constructor were introduced in reset_predictor.
6. All changes were added to DistributedAnchorTabular.

### DIFF
--- a/alibi/explainers/anchor_tabular.py
+++ b/alibi/explainers/anchor_tabular.py
@@ -3,7 +3,7 @@ import logging
 import numpy as np
 from collections import OrderedDict, defaultdict
 from itertools import accumulate
-from typing import Any, Callable, DefaultDict, Dict, List, Set, Tuple, Union
+from typing import Any, Callable, DefaultDict, Dict, List, Set, Tuple, Union, Optional
 
 from alibi.api.interfaces import Explainer, Explanation, FitMixin
 from alibi.api.defaults import DEFAULT_META_ANCHOR, DEFAULT_DATA_ANCHOR
@@ -12,6 +12,7 @@ from .anchor_explanation import AnchorExplanation
 from alibi.utils.wrappers import ArgmaxTransformer
 from alibi.utils.discretizer import Discretizer
 from alibi.utils.distributed import RAY_INSTALLED
+from alibi.utils.mapping import ohe_to_ord, ord_to_ohe
 
 
 class TabularSampler:
@@ -657,8 +658,8 @@ class RemoteSampler:
 
 class AnchorTabular(Explainer, FitMixin):
 
-    def __init__(self, predictor: Callable, feature_names: list, categorical_names: dict = None,
-                 seed: int = None) -> None:
+    def __init__(self, predictor: Callable, feature_names: List[str], categorical_names: Dict[int, List[str]] = None,
+                 ohe: bool = False, seed: int = None) -> None:
         """
         Parameters
         ----------
@@ -668,15 +669,24 @@ class AnchorTabular(Explainer, FitMixin):
             List with feature names.
         categorical_names
             Dictionary where keys are feature columns and values are the categories for the feature.
+        ohe
+            Whether the categorical variables are one-hot encoded (OHE) or not. If not OHE, they are
+            assumed to have ordinal encodings.
         seed
             Used to set the random number generator for repeatability purposes.
         """
         super().__init__(meta=copy.deepcopy(DEFAULT_META_ANCHOR))
 
+        self.ohe = ohe
         self.feature_names = feature_names
 
-        # set the predictor
-        self.predictor = self._transform_predictor(predictor)
+        if ohe:
+            self.cat_vars_ord = {col: len(values) for col, values in categorical_names.items()}
+            self.cat_vars_ohe = ord_to_ohe(np.zeros((1, len(feature_names))), self.cat_vars_ord)[1]
+
+        # defines self._predictor which expect label categorical features, and if ohe == True,
+        # it defines self._ohe_predictor which expects one-hot encoded categorical features
+        self.predictor = predictor
 
         # define column indices of categorical and numerical (aka continuous) features
         if categorical_names:
@@ -690,6 +700,7 @@ class AnchorTabular(Explainer, FitMixin):
         self.numerical_features = [x for x in range(len(feature_names)) if x not in self.categorical_features]
 
         self.samplers = []  # type: list
+        self.ohe = ohe
         self.seed = seed
         self.instance_label = None
 
@@ -711,13 +722,16 @@ class AnchorTabular(Explainer, FitMixin):
             List with percentiles (int) used for discretization.
         """
 
+        # transform one-hot encodings to labels if ohe == True
+        train_data = ohe_to_ord(X_ohe=train_data, cat_vars_ohe=self.cat_vars_ohe)[0] if self.ohe else train_data
+
         # discretization of continuous features
         disc = Discretizer(train_data, self.numerical_features, self.feature_names, percentiles=disc_perc)
         d_train_data = disc.discretize(train_data)
         self.feature_values.update(disc.feature_intervals)
 
         sampler = TabularSampler(
-            self.predictor,
+            self._predictor,
             disc_perc,
             self.numerical_features,
             self.categorical_features,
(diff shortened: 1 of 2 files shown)


jklaise:
One thing to check is that the saving functionality is not impacted by changing `predictor` to be a property, i.e. see the following logic: https://github.com/SeldonIO/alibi/blob/73eabdacbdb08a75494955a9affbb800ac73abbb/alibi/saving.py#L101-L106

Because the `predictor` is set, we may need to write a setter method for it to work. I suspect the current save/load test (https://github.com/SeldonIO/alibi/blob/73eabdacbdb08a75494955a9affbb800ac73abbb/alibi/tests/test_saving.py#L261) won't pass right now.

jklaise:
We should add a few sentenced in the Anchors documentation under `Tabular Data` about the new `ohe` flag and that inputs can be either label/ordinal encoded or ohe encoded: https://github.com/SeldonIO/alibi/blob/master/doc/source/methods/Anchors.ipynb

jklaise:
We also need to check that after an explanation is computed, all the instance-related data fields in `explanation.data` are in the ohe-format if `ohe=True`, I don't think that is currently the case (i.e. check `explanation.raw['instance']`, `explanation.raw['instances']`, `explanation.raw['examples']`.

alibi/explainers/anchor_tabular.py
```diff
@@ -937,13 +964,13 @@ def add_names_to_exp(self, explanation: dict) -> None:
...
                         geq = geq - 1
-                    name = self.feature_values[feat_id][geq + 1]
+                    name = self.feature_values[feat_id][int(geq) + 1]
                     if '<' in name:
                         geq_val = name.split()[0]
                     elif '>' in name:
                         geq_val = name.split()[-1]
                 if leq < float('inf'):
-                    name = self.feature_values[feat_id][leq]
+                    name = self.feature_values[feat_id][int(leq)]
```
jklaise, line 973: Under what circumstances do these need to be cast to `int` explicitly? What change in this PR made this casting necessary?

alibi/explainers/anchor_tabular.py
```diff
@@ -979,9 +1037,9 @@ class DistributedAnchorTabular(AnchorTabular):
         ray = ray  # set module as class variable to used only in this context
 
     def __init__(self, predictor: Callable, feature_names: list, categorical_names: dict = None,
```
jklaise, line 1039: Add the same explicit type hints for `feature_names` and `categorical_names`.

alibi/explainers/anchor_tabular.py
```diff
@@ -958,19 +985,50 @@ def add_names_to_exp(self, explanation: dict) -> None:
...
     def _transform_predictor(self, predictor: Callable) -> Callable:
-        # check if predictor returns predicted class or prediction probabilities for each class
-        # if needed adjust predictor so it returns the predicted class
-        if np.argmax(predictor(np.zeros([1, len(self.feature_names)])).shape) == 0:
+        try:
+            # check if predictor returns predicted class or prediction probabilities for each class
+            # if needed adjust predictor so it returns the predicted class
+            prediction = predictor(np.zeros([1, len(self.feature_names)]))
+        except Exception:
+            raise ValueError("The classifier is expecting a different number of features.")
```
jklaise, line 1003: We should be more explicit in the error message, e.g. add "Check the values of `feature_names` and `categorical_names`", or even, if it's possible to keep track of the inferred number of columns, "Calling the classifier using a datapoint with {n_cols} failed, check the values of `feature_names` and `categorical_names`."

alibi/explainers/anchor_tabular.py
```diff
@@ -958,19 +985,50 @@ def add_names_to_exp(self, explanation: dict) -> None:
...
+
+        if np.argmax(prediction.shape) == 0:
             return predictor
         else:
             transformer = ArgmaxTransformer(predictor)
             return transformer
 
+    def _transform_ohe_predictor(self, predictor: Callable) -> Callable:
+        if type(self._predictor) == ArgmaxTransformer:
```
jklaise, line 1012: `isinstance(self._predictor, ArgmaxTransformer)` is more idiomatic

alibi/explainers/anchor_tabular.py
jklaise: This is due to `mypy` complaining as the `category_map` is now fully typed with integer keys.

alibi/explainers/anchor_tabular.py
```diff
@@ -958,30 +979,64 @@ def add_names_to_exp(self, explanation: dict) -> None:
...
+            # this predictor expects ordinal/labels encoded categorical variables
+            ord_predictor = lambda x: predictor(ord_to_ohe(x, self.cat_vars_ord)[0])  # noqa: E731
+            self._predictor = self._transform_predictor(ord_predictor) if predictor else None
+
+            # this predictor expects one-hot encoded categorical variable
+            self._ohe_predictor = self._transform_ohe_predictor(predictor) if predictor else None
+
+        else:
+            # set the predictor
+            self._predictor = self._transform_predictor(predictor) if predictor else None
```
jklaise, line 999: What's the reason for these `if predictor else None`? The `predictor` is typed as `Callable` so `None` should not be passed, we probably shouldn't silently set `self._predictor` etc. to `None` unless I'm missing something?

RobertSamoilescu: `explainer.predictor = None` in the tests.

https://github.com/SeldonIO/alibi/blob/af1b3513f000daf4061eb20510ad539e70edeebc/alibi/saving.py#L101-L106

jklaise: Great, specifying the type hint for the setter to `Optional[Callable]` should be then good to go.

alibi/explainers/anchor_tabular.py
```diff
@@ -958,30 +979,64 @@ def add_names_to_exp(self, explanation: dict) -> None:
...
+
+        try:
+            # check if predictor returns predicted class or prediction probabilities for each class
+            # if needed adjust predictor so it returns the predicted class
+            prediction = predictor(zeros)
+        except Exception:
+            ncols = ord_to_ohe(zeros, self.cat_vars_ord)[0].shape[1] if self.ohe else zeros.shape[1]
+            raise ValueError("The classifier is expecting a different number of features. "
+                             f"Calling a classifier using {ncols} failed, check the values"
+                             "of the `feature_names` and `categorical_names`.")
```
jklaise, line 1013: Please expand slightly to `using {ncols} features` and add a space after `check the values " just like in the first line.

jklaise (approved):

PR #457, opened 2021-07-27
CF-RL project

RobertSamoilescu:
This pull request contains a potential design for the CF-RL project. It supports tensorflow and pytorch backends and contains working examples for two data modalities: MNIST and ADULT. The API is designed to be easy to use by practitioners and flexible to allow further improvements/experiments for researchers.

The flexible base class is defined here:
https://github.com/RobertSamoilescu/alibi/blob/0da4cc886058fb40eba21eaa1fd321f86218dbbd/alibi/explainers/cfrl_base.py#L300-L302

The tabular class is defined here:
https://github.com/RobertSamoilescu/alibi/blob/0da4cc886058fb40eba21eaa1fd321f86218dbbd/alibi/explainers/cfrl_tabular.py#L39-L41

The `explainer` package contains a new `backends` package that contains the implementation for `tensorflow` and `pytorch`.

**This is just a preview**. The functionalities included are not properly tested.

### TODOs:

- pytests to be included.
- documentation
- clean the examples
- fix `mypy` & `flake8` issues


jklaise (commented):
This comment will be updated with a list of things to do before merging (can be done last).

- [x] Expose new explainers in `explainers.__init__`
- [x] Add module docstrings explaining the purpose of each new module.
- [ ] Check that the API docs render correctly
- [ ] ~Would be nice to have default callbacks that log everything to `wandb` and can be turned on by a flag by the user~ (Will discuss as part of a more general PR)
- [ ] Add `torch` to `requirements/dev.txt`

jklaise (changes requested):
Nice work Robert! Mostly small comments from me.

alibi/explainers/cfrl_base.py
```diff
@@ -0,0 +1,714 @@
...
+
+    - ``'critic_hidden_dim'``: int, critic hidden layer dimension.
+"""
+
+
+class CounterfactualRLBase(Explainer, FitMixin):
+    """ Counterfactual Reinforcement Learning Base. """
+
+    PYTORCH = "pytorch"
+    TENSORFLOW = "tensorflow"
```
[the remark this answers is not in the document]
RobertSamoilescu: To work, this needs to be:

```python
class Framework(str, Enum):
    PYTORCH = "pytorch"
    TENSORFLOW = "tensorflow"
```

alibi/explainers/cfrl_base.py
```diff
@@ -0,0 +1,714 @@
...
+
+    def __init__(self,
+                 predict_func: Callable,
```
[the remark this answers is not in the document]
RobertSamoilescu: - renamed `predict_func` to `predictor` .
- yes. The predictor can be on GPU as well the actor, critic, autoencoder.

alibi/explainers/cfrl_base.py
```diff
@@ -0,0 +1,714 @@
...
+                 ae,
+                 latent_dim: int,
+                 coeff_sparsity: float,
+                 coeff_consistency: float,
+                 num_classes: int,
```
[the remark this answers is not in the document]
RobertSamoilescu: I will address the possibility of extending it to regression. Might need just to change the data loader Don't think we can infer it from the predictor.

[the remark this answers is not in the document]
jklaise: Doesn't have to be part of this PR but something to keep in mind / open an issue.

BTW I think in `IntegratedGradients` we infer the task based on the number of things the predictor returns... Maybe not a robust solution but avoids (for now) extra user parameters defining the task and task-specific parameters (e.g. `num_classes`).

alibi/explainers/cfrl_base.py
```diff
@@ -0,0 +1,714 @@
...
+
+        # Allow only pytorch and tensorflow.
+        elif backend not in [CounterfactualRLBase.PYTORCH, CounterfactualRLBase.TENSORFLOW]:
+            raise NotImplementedError(f'{backend} not implemented. Use `tensorflow` or `pytorch` instead.')
+
+        # Select backend.
+        self.backend = self._select_backend(backend, **kwargs)
+
+        # Validate arguments.
+        self.params, all_params = self._validate_kwargs(predict_func=predict_func,
```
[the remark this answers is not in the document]
RobertSamoilescu: Modified them to be public in the following commit.
`all_params` are used to update the metadata. Should we update the metadata with all parameters passed (correct or incorrect). Some incorrect params might be problematic when saving?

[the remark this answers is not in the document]
jklaise: We shouldn't add anything that doesn't belong into the metadata (i.e. not defined by the method) so I would get rid of it.

alibi/explainers/cfrl_base.py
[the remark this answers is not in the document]
RobertSamoilescu: So in this case the predictor should return soft/hard labels which is a tensor having the second dimension at least two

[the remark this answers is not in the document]
jklaise: Yea that's one option, of course multi-output regression models exist too... I think there is a more general discussion on an interface defining "tasks supported by alibi methods" in an explicit way in the future.

alibi/explainers/cfrl_base.py
```diff
@@ -0,0 +1,714 @@
...
+        z_cf_tilde
+            Noised counterfactual embedding.
+        c
+            Conditional array.
+        r_tilde
+            Noised counterfactual reward array.
+        """
+        # Initialize the buffers.
+        if self.x is None:
+            self.batch_size = x.shape[0]
```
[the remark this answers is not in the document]
RobertSamoilescu: During training, the batch_size is fixed by dropping the last one. Allowing the buffer size to have a flexible size can also generate warnings from TensorFlow because of the `@tf.function` decorator. This can also slow down the training. I would keep the batch size fixed during training.

alibi/explainers/cfrl_base.py
```diff
@@ -0,0 +1,714 @@
...
+        params.update({key: kwargs[key] for key in common_keys})
+        all_params.update(kwargs)
+        return params, all_params
+
+    @classmethod
+    def load(cls, path: Union[str, os.PathLike], predictor: Any) -> "Explainer":
+        return super().load(path, predictor)
+
+    def reset_predictor(self, predictor: Any) -> None:
+        pass
```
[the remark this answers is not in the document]
RobertSamoilescu: Following PR contains an implementation of the function.

alibi/explainers/cfrl_base.py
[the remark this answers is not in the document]
jklaise: OK sounds good, I would just document the behaviour and the reasoning for the fixed size.

alibi/explainers/cfrl_base.py
```diff
@@ -0,0 +1,714 @@
...
+
+        # TODO construct explanation
+        return self.params["ae_inv_preprocessor"](x_cf)
+
+
+class Postprocessing(ABC):
+    @abstractmethod
+    def __call__(self, x_cf: List[np.ndarray], x: np.ndarray, c: np.ndarray) -> Any:
+        """
+        Post-processing function
```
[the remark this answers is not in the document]
RobertSamoilescu: I guess the interface was intended for the tabular case and moved later to the base class. For the tabular case, sometimes it is easier to operate on each column and have a clear separation. This is also the format of the output of a heterogeneous autoencoder. I guess we can go for either `Any` or `Union[np.ndarray, List[np.ndarray]]`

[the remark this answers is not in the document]
jklaise: I think it just needs a bit more documentation of what is meant by "columns" and that it relates to the list output requirement for HeAE.

alibi/explainers/cfrl_base.py
```diff
@@ -0,0 +1,871 @@
...
+        """
+        meta = dict()
+
+        for param, value in params.items():
+            if param in _PARAM_TYPES["primitives"]:
+                # primitive types are passed as they are
+                meta.update({param: value})
+
+            elif param in _PARAM_TYPES["complex"]:
+                if isinstance(value, list):
```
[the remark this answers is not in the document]
RobertSamoilescu: This implementation is specific to CF-RL. There are not dictionary parameters. But I agree that for a more generic functionality we may need to handed other collections.

alibi/explainers/cfrl_base.py
```diff
@@ -0,0 +1,871 @@
...
+        y_t
+            Target labels.
+        c
+            Condition
+        Returns
+        -------
+        `Explanation` object containing the inputs with the corresponding labels, the counterfactuals with the
+        corresponding labels, targets and additional metadata.
+        """
+        data = {"orig": {"X": X, "class": y_m}, "cf": {"X": X_cf, "class": y_m_cf}, "target": y_t, "condition": c}
```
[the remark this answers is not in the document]
RobertSamoilescu: solved in the following commit

alibi/explainers/cfrl_tabular.py
```diff
@@ -0,0 +1,310 @@
...
+
+    def explain(self,
+                X: np.ndarray,
+                y_t: np.ndarray = None,  # TODO remove default value (mypy error)
+                c: Optional[List[Dict[str, List[Union[str, float]]]]] = None,
+                diversity: bool = False,
+                num_samples: int = 1,
+                batch_size: int = 100,
+                patience: int = 1000,
+                tolerance: float = 1e-3) -> Explanation:
```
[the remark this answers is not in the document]
RobertSamoilescu: The CF-RL has normally a deterministic behavior. For the tabular case we are able to generate a diverse counterfactual set by generating conditions that are a subset of the original condition. In this way, we keep the feasibility of the counterfactual. To actually generate a subset of the original condition, you need to know some insights regarding the condition vector. It can probably be extended, but I think it might require some user-defined subset condition generator.  At the same time, it can be a bit difficult since the explain method in the base, expects a numerical conditional vector which is not really human-readable.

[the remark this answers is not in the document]
jklaise: OK, I wasn't really suggesting extending the base class to handle `diversity` more that e.g. `batch_size` seemed like a general parameter that's not present in `cfrl_base`  ?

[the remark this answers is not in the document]
RobertSamoilescu: I've already included the `batch_size` in the following PR.

alibi/explainers/backends/pytorch/cfrl_base.py
```diff
@@ -0,0 +1,555 @@
...
+                                                             batch_size=self.batch_size)
+
+        # Preprocess the input data.
+        self.x = self.preprocessor(self.x)
+
+    def __len__(self):
+        return self.x.shape[0]
+
+    def __getitem__(self, idx) -> Dict[str, np.ndarray]:
+        self.num_classes = np.clip(self.num_classes, a_min=0, a_max=2)  # TODO: remove this
```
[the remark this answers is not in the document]
RobertSamoilescu: This is just for testing MNIST. Convergence is faster for 2 classes than for 10 classes. This will be deleted in the final version.

alibi/explainers/backends/pytorch/cfrl_base.py
```diff
@@ -0,0 +1,555 @@
...
+
+        # Generate random target class.
+        y_t = np.random.randint(low=0, high=self.num_classes, size=1).item()
```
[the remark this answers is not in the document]
RobertSamoilescu: I do this in the following commits.

alibi/explainers/backends/tflow/cfrl_base.py
```diff
@@ -0,0 +1,523 @@
...
+        conditional_func
+            Conditional function generator. Given an preprocesed input array, the functions generates a conditional
+            array.
+        num_classes
+            Number of classes in the dataset.
+        batch_size
+            Dimension of the batch used during training. The same batch size is used to infer the classification
+            labels of the input dataset.
+        shuffle
+            Whether to shuffle the dataset each epoch. `True` by default.
```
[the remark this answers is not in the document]
RobertSamoilescu: It is necessary for both backends.

alibi/explainers/backends/tflow/cfrl_base.py
```diff
@@ -0,0 +1,523 @@
...
+        """
+        This method is called every epoch and performs dataset shuffling.
+        """
+        self.indexes = np.arange(self.x.shape[0])
+
+        if self.shuffle:
+            np.random.shuffle(self.indexes)
+
+    def __len__(self) -> int:
+        return self.x.shape[0] // self.batch_size
```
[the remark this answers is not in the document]
RobertSamoilescu: The data generator in `keras` operates in batches. In `torch` each worker processes one instance. The notion of epochs is well defined for both backends.

alibi/explainers/backends/pytorch/cfrl_base.py
```diff
@@ -0,0 +1,555 @@
...
+        Pre-trained autoencoder.
+    device
+        Device to send data to.
+
+    Returns
+    -------
+    Input encoding.
+    """
+    ae.eval()
+    return ae.encoder(x.float().to(device))
```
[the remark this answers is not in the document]
RobertSamoilescu: The `ae` should always be in `eval` mode. Sometimes `torch` is sensitive with datatypes. So to ensure the input is the same datatype as the weights I convert them to floats.

alibi/explainers/backends/pytorch/cfrl_base.py
```diff
@@ -0,0 +1,555 @@
...
+        Pre-trained autoencoder.
+    device
+        Device to sent data to.
+
+    Returns
+    -------
+    Embedding tensor decoding.
+    """
+    ae.eval()
+    return ae.decoder(z.float().to(device))
```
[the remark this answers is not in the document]
RobertSamoilescu: See the answer above.

alibi/explainers/backends/pytorch/cfrl_base.py
```diff
@@ -0,0 +1,555 @@
...
+        Device object to be used.
+
+    Returns
+    -------
+    z_cf
+        Counterfactual embedding.
+    """
+    # Set autoencoder and actor to evaluation mode.
+    ae.eval()
+    actor.eval()
```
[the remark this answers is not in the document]
RobertSamoilescu: There is a convention throughout the implementation that the mode is set just before the call, depending on whether I need `eval` or `train`. This is why I do not set them back to `train`.

alibi/models/tflow/cfrl_models.py
```diff
@@ -0,0 +1,140 @@
+from abc import ABC
+
+import tensorflow as tf
+import tensorflow.keras as keras
+from typing import List
+
+
+class MNISTClassifier(keras.Model, ABC):
```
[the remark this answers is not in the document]
RobertSamoilescu: pycharm complains ... I will remove it

alibi/models/pytorch/cfrl_models.py
```diff
@@ -0,0 +1,157 @@
...
+        self.dropout1 = nn.Dropout(0.3)
+        self.conv2 = nn.Conv2d(64, 32, kernel_size=(2, 2), padding=1)
+        self.maxpool2 = nn.MaxPool2d(kernel_size=2, stride=2)
+        self.dropout2 = nn.Dropout(0.3)
+        self.flatten = nn.Flatten()
+        self.fc1 = nn.Linear(1568, 256)
+        self.fc2 = nn.Linear(256, output_dim)
+
+        # send to device
+        self.to(self.device)
```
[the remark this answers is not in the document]
RobertSamoilescu: The MNIST classifier inherits from the `Model` class which sets the device. It just tries to avoid setting the device manually by the user and use the GPU automatically if available. There are not any drawbacks. Probably I was trying to have something more similar with tf.

alibi/explainers/backends/pytorch/cfrl_base.py
```diff
@@ -0,0 +1,555 @@
...
+
+        self.x = x
+        self.preprocessor = preprocessor
+        self.predict_func = predict_func
+        self.conditional_func = conditional_func
+        self.num_classes = num_classes
+        self.batch_size = batch_size
+
+        # Infer the classification labels of the input dataset. This is performed in batches.
+        self.y_m = PtCounterfactualRLDataset.predict_batches(x=self.x,
```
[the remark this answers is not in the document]
RobertSamoilescu: it works with static classes too...

alibi/explainers/backends/pytorch/cfrl_base.py
```diff
@@ -0,0 +1,555 @@
...
+    Constructs default Adam optimizer.
+
+    Returns
+    -------
+    Default optimizer.
+    """
+    return torch.optim.Adam(model.parameters(), lr=lr)
+
+
+def get_actor(hidden_dim: int, output_dim: int, **kwargs) -> nn.Module:
```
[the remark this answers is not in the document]
RobertSamoilescu: not necessary. will remove them.

alibi/explainers/backends/tflow/cfrl_base.py
```diff
@@ -0,0 +1,523 @@
...
+
+        # Select indices to be returned.
+        indexes = self.indexes[idx * self.batch_size:(idx + 1) * self.batch_size]
+
+        # Generate random target.
+        y_t = np.random.randint(low=0, high=self.num_classes, size=self.batch_size)
+
+        # compute conditional vector.
+        c = self.conditional_func(self.x[idx * self.batch_size: (idx + 1) * self.batch_size])
+
```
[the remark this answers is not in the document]
RobertSamoilescu: There is no need here. In PyTorch, each worker processes a single item, and then a concatenation of the tensors is performed. Thus it is necessary to check if `c is not None` as a concatenation of `None`'s will raise an error. Since in TensorFlow we operate on batches, no concatenation is performed and thus no error in `c is None`.

alibi/models/pytorch/model.py
```diff
@@ -0,0 +1,275 @@
...
+import torch.nn as nn
+import torch.optim as optim
+from torch.utils.data import DataLoader
+
+from tqdm import tqdm
+from typing import List, Dict, Callable, Union, Tuple, Any, Optional
+from alibi.models.pytorch.metrics import Metric, LossContainer
+
+
+class NamedModule(nn.Module):
```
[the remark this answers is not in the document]
RobertSamoilescu: I tried to use it at some point, but right now there is of no use. I will delete it.

alibi/explainers/cfrl_tabular.py
```diff
@@ -0,0 +1,310 @@
...
+        if len(X.shape) > 2:
+            raise ValueError("The input should be a 2D array.")
+
+        # Reshape the input vector to have be 2D.
+        X = np.atleast_2d(X)
+
+        # Check if the number of features matches the expected one.
+        if X.shape[1] != len(self.params["feature_names"]):
+            raise ValueError(f"Unexpected number of features. The expected number "
+                             f"is {len(self.params['feature_names'])}, but the input has {X.shape[1]} features.")
```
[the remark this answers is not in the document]
RobertSamoilescu: yes. I moved it in a method and perform the validation in both `fit` and `explain` as it can use different splits.

alibi/explainers/backends/cfrl_tabular.py
```diff
@@ -0,0 +1,685 @@
...
+    x_ohe = preprocessor(x)
+
+    for i, feature_id in enumerate(num_features_ids):
+        min, max = np.min(x_ohe[:, i]), np.max(x_ohe[:, i])
+        stats[feature_id] = {"min": min, "max": max}
+
+    return stats
+
+
+def conditional_vector(x: np.ndarray,
```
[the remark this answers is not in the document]
RobertSamoilescu: I agree that the function is a bit complex and I will split it in two (one for numerical and the other for categorical).

alibi/explainers/backends/tflow/cfrl_base.py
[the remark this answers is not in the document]
jklaise: I mentioned this because it was missing from the `pytorch` one

alibi/explainers/backends/pytorch/cfrl_base.py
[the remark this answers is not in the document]
jklaise: Does that mean for the `pytorch` backend we should just call `ae.eval()` once somewhere at the top level and avoid having these inside the encode/decode functions?

alibi/explainers/backends/pytorch/cfrl_base.py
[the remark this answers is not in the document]
jklaise: Might be slightly more clear using `self` or `CounterfactualRLDataset` ?

alibi/explainers/backends/pytorch/cfrl_base.py
```diff
@@ -510,46 +512,98 @@ def update_actor_critic(ae: AE,
...
+
+    # torch related
+    torch.manual_seed(seed)
+    torch.cuda.manual_seed_all(seed)
+    torch.backends.cudnn.deterministic = True
+    torch.backends.cudnn.benchmark = False
+
+    # others
+    np.random.seed(seed)
+    random.seed(seed)
```
jklaise, line 609: These three functions feel like they belong into a generic `pytorch` utility module so can be reused for other stuff in the future?

alibi/explainers/backends/tensorflow/cfrl_base.py
```diff
@@ -479,45 +481,93 @@ def update_actor_critic(ae: keras.Model,
...
+    seed
+        seed to be set
+    """
+    # others
+    os.environ['PYTHONHASHSEED'] = str(seed)
+    random.seed(seed)
+    np.random.seed(seed)
+
+    # tf random
+    tf.random.set_seed(seed)
```
jklaise, line 573: Same comment as for `pytorch`, + `to_numpy`, `to_numpy` functions can also be factored out into a generic `tensorflow` utility module.

That does raise the question how to manage it in the public classes as `self.backend.to_numpy` wouldn't work if factored out, unless we do the same trick and import it (`from alibi.explainers.backends.tflow.utils import to_numpy, to_tensor`)?

RobertSamoilescu: I would say so.

alibi/explainers/backends/pytorch/cfrl_base.py
RobertSamoilescu: Yes. The functions are generic.

alibi/explainers/cfrl_base.py
```diff
@@ -77,66 +81,69 @@ def __init__(self, size=1000):
...
-        self.c = None                            # buffer for the conditional tensor
-        self.r_tilde = None                      # buffer for the noised counterfactual reward tensor
+        self.X: Optional[np.ndarray] = None              # buffer for the inputs
+        self.X_cf: Optional[np.ndarray] = None           # buffer for the counterfactuals
+        self.Y_m: Optional[np.ndarray] = None            # buffer for the model's prediction
+        self.Y_t: Optional[np.ndarray] = None            # buffer for the counterfactual targets
+        self.Z: Optional[np.ndarray] = None              # buffer for the input embedding
+        self.Z_cf_tilde: Optional[np.ndarray] = None     # buffer for the noised counterfactual embedding
+        self.C: Optional[np.ndarray] = None              # buffer for the conditional tensor
+        self.R_tilde: Optional[np.ndarray] = None        # buffer for the noised counterfactual reward tensor
```
jklaise, line 91: The `Optional` here are not really necessary as we have set "use strict optional" to `false` in mypy:
https://github.com/SeldonIO/alibi/blob/390a255403d61e8d7f87123f745b678b0a5e6753/setup.cfg#L24
https://mypy.readthedocs.io/en/stable/kinds_of_types.html#disabling-strict-optional-checking

alibi/explainers/cfrl_base.py
```diff
@@ -318,68 +331,72 @@ def sample(self) -> Dict[str, Optional[np.ndarray]]:
...
+    """
+     Counterfactual Reinforcement Learning Base.
+     TODO: check if all tensors are upperca
+    """
 
     def __init__(self,
-                 predict_func: Callable,
-                 ae,
+                 predictor: Callable,
+                 ae: Union['TensorflowAE', 'PytorchAE'],
```
jklaise, line 341: Just to confirm, we're going for an API that takes `encoder`, `decoder` as native `tensorflow` or `pytorch` types ?

RobertSamoilescu: Yes. I will modify it to encoder and decoder.

RobertSamoilescu: Would this require defining an encoder/decoder wrapper for each backend?

jklaise: The type hints can be native types, i.e. `tf.keras.Layer`, `nn.Module`, they just need to be imported within the `TYPE_CHECKING` guard.

alibi/models/tensorflow/actor_critic.py
```diff
@@ -1,9 +1,15 @@
+from abc import ABC
+
 import tensorflow as tf
 import tensorflow.keras as keras
 
 
-class Actor(keras.layers.Layer):
-    """ Actor network. """
+class Actor(keras.Model, ABC):
```
jklaise, line 7: Is the `ABC` here due to `mypy` complaints? Should be a way round it... (same for `Critic`).

alibi/explainers/backends/tflow/cfrl_base.py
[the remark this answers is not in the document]
RobertSamoilescu: For `pytorch`, the shuffling is performed by the `Dataloader`, not by the `Dataset` which I am defining. That's the reason why it is missing there.

alibi/explainers/cfrl_base.py
RobertSamoilescu: `Incompatible types in assignment (expression has type "ndarray", variable has type "None")`

alibi/models/tensorflow/actor_critic.py
RobertSamoilescu: forgot to delete it. The complaint was from pycharm to implement all methods.

alibi/explainers/backends/pytorch/cfrl_base.py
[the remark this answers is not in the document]
RobertSamoilescu: Could be done here:
https://github.com/SeldonIO/alibi/blob/d259921cee3e0b03f61d4953a8c987127dcea81d/alibi/explainers/cfrl_base.py#L399-L408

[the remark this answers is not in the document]
jklaise: Hmm, I don't have strong feelings either way, I see what you mean by using the convention where `eval` is called just before the forward pass (equivalent to `training=False` using `tensorflow`), so maybe it's actually more clear to leave as is?

alibi/explainers/cfrl_base.py
```diff
@@ -0,0 +1,1060 @@
...
+        coeff_sparsity
+            Sparsity loss coefficient.
+        coeff_consistency
+            Consistency loss coefficient.
+        backend
+            Deep learning backend: `tensorflow`|`pytorch`. Default `tensorflow`.
+        seed
+            Seed for reproducibility. The results are not reproducible for `tensorflow` backend.
+        kwargs
+            Used to replace any default parameter from :py:data:`alibi.expaliners.cfrl_base.DEFAULT_BASE_PARAMS`.
```
jklaise, line 370: Typo in the link (check that the link works in the built docs).

alibi/explainers/cfrl_base.py
```diff
@@ -0,0 +1,1060 @@
...
+    - ''`primitives`'': List[str], list of parameters having primitive data types.
+
+    - ''`complex`'': List[str], list of parameters having complex data types (e.g., functions, models, optimizers etc.)
+"""
+
+
+class CounterfactualRLBase(Explainer, FitMixin):
+    """
+     Counterfactual Reinforcement Learning Base.
+     TODO: check if all tensors are uppercase
```
jklaise, line 335: These TODO's should be outside the docstrings as otherwise they get shown on the API docs.

alibi/explainers/cfrl_base.py
```diff
@@ -0,0 +1,1060 @@
...
+    """
+
+    def __init__(self,
+                 predictor: Callable,
+                 encoder: Union['keras.Model', 'nn.Module'],
+                 decoder: Union['keras.Model', 'nn.Module'],
```
jklaise, line 341: If you check the API docs you will notice that these types are not rendered nicely, I think this is because `keras.Model` and `nn.Module` are only well defined if `TYPE_CHECKING` is true which does not happen when running `sphinx` to build docs. I would recommend using the fully qualified class names, i.e. `tf.keras.Model` and `torch.nn.Module` and check if the problem goes away (there may still be an issue due to `TYPE_CHECKING` being False...).

doc/source/methods/CFRL.ipynb
```diff
@@ -0,0 +1,475 @@
+{
```
jklaise, line 1: <p>Typo: X --&gt; x</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

jklaise, line 1: <p>Typo: X --&gt; x</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

jklaise, line 1: <p>Add a line to import <code>get_he_preproccesor</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

examples/cfrl_adult.ipynb
```diff
@@ -0,0 +1,1449 @@
+{
```
jklaise, line 1: <p>Emphasize at the start that this example is using <code>tensorflow</code> autoencoder and backend and <code>pytorch</code> is also supported.</p><p> </p><p>Typo: generatae --&gt; generate</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

jklaise, line 1: <p>Please import explainers from <code>alibi.explainers</code>  as that is the public package</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

jklaise, line 1: <p>"uses an extrac type checking"  ?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

examples/cfrl_mnist.ipynb
```diff
@@ -0,0 +1,820 @@
+{
```
jklaise, line 1: <p>Emphasize at the start that this example is using a <code>tensorflow</code> autoencoder and backend and <code>pytorch</code> is also supported.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

jklaise, line 1: <p>Please import explainers from <code>alibi.explainers</code>  as that is the public package.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

jklaise, line 1: <p>Typo: experince --&gt; experience</p><p> </p><p>In general, I recommend installing a spell-check extension to the Jupyter notebook, e.g. https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/spellchecker/README.html</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

alibi/explainers/cfrl_base.py
RobertSamoilescu: Still not working with `tf.keras.Model` and `torch.nn.Module`.

examples/cfrl_adult.ipynb
RobertSamoilescu: should have been `extra`. But maybe `additional` sounds better.

alibi/explainers/cfrl_base.py
RobertSamoilescu: The workaround:
Union['tf.keras.Model', 'torch.nn.Module'] -> 'Union[tf.keras.Model, torch.nn.Module']

README.md
```diff
@@ -123,6 +123,7 @@ The following tables summarize the possible use cases for each method.
 |[CEM](https://docs.seldon.io/projects/alibi/en/latest/methods/CEM.html)|BB* TF/Keras|local|✔| |✔| |✔| |Optional| |
 |[Counterfactuals](https://docs.seldon.io/projects/alibi/en/latest/methods/CF.html)|BB* TF/Keras|local|✔| |✔| |✔| |No| |
 |[Prototype Counterfactuals](https://docs.seldon.io/projects/alibi/en/latest/methods/CFProto.html)|BB* TF/Keras|local|✔| |✔| |✔|✔|Optional| |
+|[Counterfactuals with RL](https://docs.seldon.io/projects/alibi/en/latest/methods/CFRL.html)|BB|local|✔| |✔| |✔|✔|✔| |
```
jklaise, line 126: Could you add this row also to `doc/source/overview/algorithms`? Also note that the overview page has a paragraph for each method with links to examples which would be great to add.

alibi/explainers/backends/cfrl_base.py
```diff
@@ -51,11 +52,11 @@ def get_classification_reward(Y_pred: np.ndarray, Y_true: np.ndarray):
     Y_pred
         Prediction output as a distribution over the possible classes.
     Y_true
-        True lable as a distribution over the possible classes.
+        True label as a distribution over the possible classes.
 
     Returns
     -------
-    Classification reward per instance: 1 if th most likely classes match, 0 otherwise.
+        Classification reward per instance. 1 if th most likely classes match, 0 otherwise.
```
jklaise, line 59: th --> the

alibi/explainers/backends/cfrl_tabular.py
```diff
@@ -1,11 +1,11 @@
 """
-This module contains utility function for the Counterfactual with Reinforcement Learning tabular class (`cfrl_tabular`)
-that are common for both Tensorflow and Pytorch backends.
+This module contains utility function for the Counterfactual with Reinforcement Learning tabular class,
```
jklaise, line 2: function --> functions

alibi/explainers/backends/cfrl_base.py
```diff
@@ -1,6 +1,6 @@
 """
-This module contains utility function for the Counterfactual with Reinforcement Learning base class (`cfrl_base`)
-that are common for both Tensorflow and Pytorch backends.
+This module contains utility function for the Counterfactual with Reinforcement Learning base class,
```
jklaise, line 2: function --> functions

alibi/explainers/cfrl_base.py
jklaise: Check if `mypy` type-checking still works with these hints, e.g. feed a `numpy` array into a function of this input signature and see if an error is raised.

alibi/explainers/backends/cfrl_tabular.py
```diff
@@ -256,25 +255,32 @@ def generate_condition(X_ohe: np.ndarray,
...
+        C_cat = generate_categorical_condition(X_ohe=X_ohe,
+                                               feature_names=feature_names,
+                                               category_map=category_map,
+                                               immutable_features=immutable_features,
+                                               conditional=conditional)
+        C.append(C_cat)
 
     # Concatenate numerical and categorical conditional vectors.
-    return np.concatenate([C_num, C_cat], axis=1)
+    return np.concatenate(C, axis=1)
```
jklaise: Not sure I understand the reason for this change... Have the underlying `generate...` functions changed too?

jklaise: Also it seems that introducing `C=[]` and then appending to it twice just introduces unnecessary lines of code compared to the older version or am I missing something?

RobertSamoilescu: To cover the cases where the dataset contains only numerical or only categorical

RobertSamoilescu: No, the `generate...` function did not change

alibi/explainers/backends/cfrl_tabular.py
```diff
@@ -363,55 +371,55 @@ def sample_categorical(X_hat_cat_split: List[np.ndarray],
...
+           stats: Dict[int, Dict[str, float]]) -> List[np.ndarray]:
     """
     Samples an instance from the given reconstruction according to the conditional vector and
     the dictionary of statistics.
 
     Parameters
     ----------
     X_hat_split
-        List of one-hot encoded reconstructed columns form the auto-encoder. The categorical columns contain logits.
+        List of reconstructed columns form the auto-encoder. The categorical columns contain logits.
```
jklaise, line 384: form --> from

alibi/explainers/backends/cfrl_tabular.py
```diff
@@ -363,55 +371,55 @@ def sample_categorical(X_hat_cat_split: List[np.ndarray],
...
-        values for a feature.
 
     Returns
     -------
-    Most probable sample according to the auto-encoder reconstruction, sampled according to the conditional vector
-    and the dictionary of statistics. This method assumes that the input array, `x_ohe`, is encoded as follows:
-    first columns correspond to the numerical features, and the rest are one-hot encodings of the categorical
-    columns.
+    X_ohe_hat_split
+        Most probable reconstruction sample the autoencoder, sampled according to the conditional vector and the
```
jklaise, line 400: "according to the autoencoder" ?

alibi/explainers/backends/cfrl_tabular.py
```diff
@@ -491,7 +503,7 @@ def get_inv_preprocessor(x_ohe: np.ndarray):
 
         np_X_inv = np_X_inv[:, inv_perm].astype(object)
         for i, fn in enumerate(feature_names):
-            type = feature_types[fn] if fn in feature_types else int
+            type = feature_types[fn] if fn in feature_types else float
```
jklaise, line 506: What was the reason for the int -> float change?

alibi/explainers/backends/cfrl_tabular.py
```diff
@@ -363,55 +371,55 @@ def sample_categorical(X_hat_cat_split: List[np.ndarray],
...
-                                         C_cat_split=C_cat_split)
+        # Sample categorical columns
+        X_ohe_hat_split += sample_categorical(X_hat_cat_split=X_hat_split[-cat_feat:],
+                                              C_cat_split=C_cat_split)
 
-    return sampled_num + sampled_cat
+    return X_ohe_hat_split
 
 
 def get_he_preprocessor(X: np.ndarray,
```
jklaise, line 425: Just noticed that this function has a mutable default `dict()`, see https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

alibi/explainers/backends/cfrl_tabular.py
RobertSamoilescu: I considered `float` to be more generic than `int`

alibi/explainers/backends/pytorch/cfrl_base.py
```diff
@@ -1,6 +1,6 @@
 """
-This module contains utility function for the Counterfactual with Reinforcement Learning tabular class (`cfrl_tabular`)
-for the Pytorch backend.
+This module contains utility function for the Counterfactual with Reinforcement Learning base class,
```
jklaise, line 2: function --> functions

alibi/explainers/backends/pytorch/cfrl_tabular.py
```diff
@@ -1,6 +1,6 @@
 """
-This module contains utility function for the Counterfactual with Reinforcement Learning tabular class (`cfrl_tabular`)
-for the Pytorch backend.
+This module contains utility function for the Counterfactual with Reinforcement Learning tabular class,
```
jklaise, line 2: function --> functions

alibi/explainers/backends/pytorch/cfrl_tabular.py
```diff
@@ -76,9 +76,11 @@ def l0_ohe(input: torch.Tensor,
 
     Returns
     -------
-    L0 loss.
+        L0 loss.
     """
-    loss = torch.maximum(target - input, torch.zeros_like(input))
+    # Order matters as the gradient of zeros will still flow if reversed order. Maybe consider clipping a bit higher?
+    eps = 1e-7 / input.shape[1]
+    loss = torch.maximum(target - input, eps + torch.zeros_like(input)).sum(dim=1)
```
jklaise, line 83: What's the reason for having to add `eps`?

alibi/explainers/backends/tensorflow/cfrl_base.py
```diff
@@ -539,13 +538,13 @@ def update_actor_critic(encoder: keras.Model,
         # Compute counterfactual embedding.
         Z_cf = actor(state, training=True)
 
-        # Compute critic's output
+        # Compute critic's output.
         input_critic = tf.concat([state, Z_cf], axis=1)
-        output_critic = critic(input_critic, training=True)
+        output_critic = critic(input_critic, training=False)
```
jklaise, line 543: Was `training=True` a bug?

alibi/explainers/cfrl_base.py
```diff
@@ -209,7 +201,7 @@ def sample(self) -> Dict[str, Optional[np.ndarray]]:
     "act_low": -1.0,
     "act_high": 1.0,
     "replay_buffer_size": 1000,
-    "batch_size": 128,
+    "batch_size": 100,
```
jklaise, line 204: What's the common thing to do these days, no more powers of 2 batch size?

alibi/explainers/cfrl_base.py
```diff
@@ -645,6 +635,9 @@ def fit(self, X: np.ndarray, ) -> "Explainer":
             try:
                 data = next(data_iter)
             except StopIteration:
+                if hasattr(data_generator, "on_epoch_end"):
+                    data_generator.on_epoch_end()
```
jklaise, line 639: This is a `tensorflow` only thing right? Maybe worth leaving a comment here.

alibi/models/pytorch/cfrl_models.py
```diff
@@ -182,6 +182,6 @@ def __init__(self, hidden_dim: int, output_dims: List[int]):
         self.fcs = nn.ModuleList([nn.LazyLinear(dim) for dim in output_dims])
 
     def forward(self, x: torch.Tensor) -> List[torch.Tensor]:
-        x = self.fc1(x)
+        x = F.relu(self.fc1(x))
```
jklaise, line 185: I assume the missing `relu` was a bug?

alibi/explainers/cfrl_base.py
```diff
@@ -657,32 +650,22 @@ def fit(self, X: np.ndarray, ) -> "Explainer":
             data.update({"Z": Z})
 
             # Compute counterfactual embedding.
-            Z_cf = self.backend.generate_cf(step=step, **data, **self.params)
+            Z_cf = self.backend.generate_cf(**data, **self.params)
```
jklaise, line 653: What changed here so that `step` isn't required?

alibi/explainers/cfrl_base.py
```diff
@@ -740,6 +740,9 @@ def _validate_target(Y_t: np.ndarray):
         Y_t
             Targets to be checked.
         """
+        if Y_t is None:
+            raise ValueError("Target can not be `None`.")
```
jklaise, line 744: Slightly confused, if `Y_t` can't be `None`, why is it allowed to be `Optional`?

doc/source/methods/CFRL.ipynb
```diff
@@ -0,0 +1,565 @@
...
```
jklaise, line 1: <p>Can you double check the order of the arguments in the example as it's slightly different in the code (e.g. coef's are mandatory arguments whilst <code>latent_dim</code>  is optional, would be good to have these consistent in the docs. Also, I wonder if we should give the actual default values here for e.g. <code>batch_size</code>  and <code>train_steps</code>  ?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

jklaise, line 1: <p>BATH_SIZE --&gt; BATCH_SIZE</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

examples/cfrl_adult.ipynb
```diff
@@ -0,0 +1,1401 @@
...
```
jklaise, line 1: <p>CfRL --&gt; CFRL</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

jklaise, line 1: <p>Is this intended to be a markdown cell and not an executable one?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='43'/>

examples/cfrl_mnist.ipynb
```diff
@@ -0,0 +1,732 @@
...
```
jklaise, line 1: <p>Unintentional "*" at the end of the bold sentence?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

jklaise, line 1: <p>Is this intended to be a markdown cell and not an executable one?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='34'/>

alibi/tests/test_saving.py
```diff
@@ -334,3 +404,34 @@ def test_save_TreeShap(tree_explainer, rf_classifier, iris_data):
...
+        cfrl_explainer1 = load_explainer(temp_dir, predictor=predictor)
+        assert isinstance(cfrl_explainer1, CounterfactualRLTabular)
+
+        # check metadata. loading can change the class, so we have to remove the encoder, decoder, actor and critic
+        # see: https://www.tensorflow.org/api_docs/python/tf/saved_model/load for more details
+        # also have to remove the optimizers since those are not saved.
+        keys_to_remove = ['encoder', 'decoder', 'actor', 'critic', 'optimizer_actor', 'optimizer_critic', 'callbacks']
+        for key in keys_to_remove:
+            cfrl_explainer.meta["params"].pop(key)
+            cfrl_explainer1.meta["params"].pop(key)
```
jklaise, line 429: Not quite sure I follow why all those are missing, is it just because we don't save them? But surely all those models are required at prediction time...?

jklaise (approved):

alibi/explainers/backends/pytorch/cfrl_tabular.py
jklaise: @RobertSamoilescu followed-up on this privately: essentially `torch` and `tensorflow` have a different behaviour on which element of `maximum` is picked if they happen to be equal. This has an effect on gradient flow which is a side-effect we depend on here. The solution here is to add a small `eps` so that the order does not play a role anymore and we can rely on stopping the gradient flow for either backend.

alibi/explainers/backends/pytorch/cfrl_base.py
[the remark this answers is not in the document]
jklaise: Should this be removed now?

Issue #453, opened 2021-07-15
Best practices on setting state during `explain` calls


jklaise:
Currently we don't have guidelines for how to manage the state of an explainer object, especially during `explain` calls. An `explain` method may be very complex internally, calling a bunch of different methods or functions, so state is often set using various attributes to keep track of calculation results. However, this can be problematic for several reasons:
 - The attribute names are not consistent, it is not clear what is meant to be public/private or what is set during initialization or fit vs `explain`.
 - Any state set (and not cleared) during `explain` will by definition be related to something we just finished explaining, so it doesn't feel right to hold onto this state after `explain` call finishes
 - Arbitrary state can introduce difficulties serializing the explainer.

On the other hand, it's not clear what the best patterns of managing state in explainer objects could be. A broader move towards functional programming patterns may help.

adriangonz:
Following the initial exploration done in PR #460, we identified that this task may need a explainer-by-explainer approach. Therefore, we've decided to split this into smaller tasks to assess how this can be done for each explainer. 

Based on the above, this task also includes now #462 and #463 as sub-tasks.

PR #460, opened 2021-08-02
Remove internal state from AnchorImage explainer

adriangonz:
Part of #453 (only covering one explainer)

## Changelog

- Refactor sampling methods to a new `AnchorImageSampler` object
- Move out the `_scale` static method into a utils module
- Add (dumb) test to make sure the `AnchorImage` explainer holds no internal state


alibi/explainers/anchor_image_sampler.py
```diff
@@ -0,0 +1,260 @@
...
+
+from typing import Tuple, Callable, List, Union
+
+from .anchor_image_utils import scale_image
+
+
+class AnchorImageSampler:
+    def __init__(
+        self,
+        # TODO: Should we call `predictor`, `prediction_fn` instead?
```
jklaise, line 12: The convention is to call everything `predictor` regardless if it's a white-box model or just a black-box prediction function, partially because some methods accept both types.

jklaise (changes requested):
Thanks @adriangonz, looking good!

The only two things I would suggest are:
 - Keep to one module for the public class + utility functions + sampler as the implementation is not that complex to warrant splitting out into several modules, same for keeping to one test module
 - We should check what the increase in test times is due to the extra test that calls `explain` as it's expensive (looks like adding ~3 minutes on top compared to master). Maybe we should combine the stateless test together with the main test to reduce runtimes (at the expense of tests not testing a single behaviour) ?

alibi/explainers/anchor_image_sampler.py
adriangonz: Makes sense. I'll remove the TODO comment. :+1:

jklaise (approved):

Issue #468, opened 2021-09-01
Rename counterfactual classes


jklaise:
With #457 merging soon, we want to standardize the naming of some of the explainer classes, i.e. `CounterFactual` --> `Counterfactual` and `CounterFactualProto` --> `CounterfactualProto`. This would of course be a breaking change, but we can deprecate the old names and issue a warning message that the old classes will be removed in an upcoming release.

aniketzz:
just updating the alibi 0.6.0 to alibi 0.6.1 fix this issue.

jklaise:
Ah this issue was for keeping track of #469 which is now merged, so closing this.

PR #469, opened 2021-09-01
Rename cf classes

jklaise:
Addresses #468.

Initially I tried to be clever and make the warnings trigger on import as well, but due to the Python import system this would trigger on *any* import of `alibi` not just targeted to the imports of the old names, so I scrapped it and the warning is printed at initialization.

I ran 2 notebooks to check they work with the new names (and trigger a warning if the old name is used).

Minor niggle: sphinx `apidoc` includes the old placeholder functions in the docs. We might look if there are options to mark certain functions/classes as not to be documented, although this may be hard due to us building API docs on-the-fly using [apidoc](https://github.com/sphinx-contrib/apidoc).


alibi/explainers/cfproto.py
```diff
@@ -17,7 +17,16 @@
 logger = logging.getLogger(__name__)
 
 
-class CounterFactualProto(Explainer, FitMixin):
+def CounterFactualProto(*args, **kwargs):
+    # TODO: remove this function in an upcoming release
```
ascillitoe, line 21: Annoyingly having an empty docstring here doesn't make `apidoc` ignore the `CounterFactualProto` method. It probably makes sense to add a deprecation warning to the docstring instead?

ascillitoe (approved):

PR #475, opened 2021-09-08
Use readthedocs docker image for GA docs builds

jklaise:
Adds the same behaviour as the recent work in https://github.com/SeldonIO/alibi-detect/pull/319


jklaise:
The latest commit replaces the rather generic build commands with those used by readthedocs. This is important for further consistency between the two docs builds and it was also necessary for the `latex` build to finish (as `readthedocs` uses the `-interaction=nonstopmode` option.

jklaise:
The latest commit changes the call `python -m sphinx` to `sphinx-build` as they are the same entry point into the builder, however `python` does not work on the Ubuntu 18.04 image as it refers to Python 2.7.

jklaise:
Two issues need to be resolved:
 - Running `sphinx-build` on Ubuntu 18.04 without using a virtual environment results in errors (autodoc can't import modules), this is probably something to do with `python` vs `python3`
 - For the latex build `-jobname` is incompatible with specifying a directory of files, `redthedocs` somehow manages to change build directories to the right ones without shell commands (probably built in their python scripts?)

jklaise:
Latest commit uses a `virtualenv`. Trying to be clever I aliased the new python binary to the system `python` command so that we don't have to change the `Makefile` commands, however it seems like bash `alias` doesn't work out-of-the-box in Github Actions.

Added a `-cd` option to `latexmk` which should change directory to the latex source before building it.

jklaise:
It appears the virtual environment on GA needs to be activated at any step where it is required which I have done.

I also tried https://github.com/marketplace/actions/debugging-with-tmate to help with debugging GA actions - it creates an ssh connection straight into the host machine. We could look to integrate this for manual debug triggers as described in the section in the docs for future debugging.

jklaise:
This setup works now, the only issue is that the latex build is broken.

jklaise:
The latex build errors seem to come from using the `\tag` macro in some notebooks to manually label equations (as Jupyter doesn't support automatic equation labelling via the `equation` environment. However, when the notebooks are parsed into `tex` sources, the result is `\tag` macros nested inside `equation` environments which is disallowed by `amsmath`.

jklaise:
Even after removing `\tag` macros, the latex exit code is 1 and the log file doesn't give too many clues about what else is wrong.

jklaise:
Found the error - a few rouge ampersands in a `split` environment.

Also reduced the number of warnings by moving from `\choose` to `\binom`, replacing `$$` with `equation` environment and redefining the unicode symbols output by `tqdm` as in https://github.com/SeldonIO/alibi-detect/pull/319

jklaise:
Spanner in the works - the `latexmk` version available on Ubuntu 18.04 is 4.41 which does not respect the `-cd` command resulting in files being dumped in the current directory (https://tex.stackexchange.com/questions/222338/latexmk-cd-option-not-working). I suspect this results in the error code 256 we're seeing that files cannot be found.

jklaise:
@ascillitoe 

- Updated `README` to standardize build process and commands
- Updated instructions for embedding images
- Did not specifically talk about `latexmk` and building latex docs as this would rarely need to be done locally.

PR #481, opened 2021-09-17
Fix bug with version saving being overwritten on subsequent saves

jklaise:
Fixes the issue found in https://github.com/SeldonIO/alibi-detect/issues/236#issuecomment-921719606.

### DIFF
--- a/alibi/api/interfaces.py
+++ b/alibi/api/interfaces.py
@@ -2,7 +2,7 @@ import abc
 import json
 import os
 from collections import ChainMap
-from typing import Any, ClassVar, Union
+from typing import Any, Union
 import logging
 from functools import partial
 import pprint
@@ -22,6 +22,7 @@ def default_meta() -> dict:
         "type": [],
         "explanations": [],
         "params": {},
+        "version": None,
     }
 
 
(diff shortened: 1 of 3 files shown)


alibi/api/interfaces.py
```diff
@@ -22,6 +22,7 @@ def default_meta() -> dict:
         "type": [],
         "explanations": [],
         "params": {},
+        "version": None,
```
jklaise, line 25: This is not strictly necessary here but added it as it serves as a kind of high-level "schema".

alibi/api/interfaces.py
```diff
@@ -67,12 +68,12 @@ class Explainer(abc.ABC):
...
     Base class for explainer algorithms
     """
-    _version: ClassVar[str] = __version__
     meta = attr.ib(default=attr.Factory(default_meta), repr=alibi_pformat)  # type: dict
 
     def __attrs_post_init__(self):
-        # add a name to the metadata dictionary
+        # add a name and version to the metadata dictionary
         self.meta["name"] = self.__class__.__name__
+        self.meta["version"] = __version__
```
jklaise, line 76: The version needs to be added here instead of in the `default_meta` because each explainer has it's own `DEFAULT_META` defined in `api.defaults`.

ascillitoe (approved):
@jklaise I just tested this and it seems to work nicely i.e. not overwriting with runtime version, but still raising a warning (which is preferable in my opinion) instead of an exception.

PR #506, opened 2021-10-14
Expose `dtype` for explainers making dummy predictions with `np.ndarray`

### LINKED ISSUE #499: `AnchorImage` init will fail with pytorch models with default fp32 precission
NN pytorch models would default to using `torch.float32` if no dtype is specified, e.g. check https://pytorch.org/docs/stable/generated/torch.nn.Linear.html

These models will fail with `AnchorImage` at initialisation.

When `AnchorImage` is initialised, there is logic that tries to detect the output of classification models (if it is class probabilities we then perform ArgMaxTransformation)

As part of the logic, we send a dummy infer request and the [data](https://github.com/SeldonIO/alibi/blob/ab7dbf9b2a153ff49f29a372d5bdeac56dc7284a/alibi/explainers/anchor_image.py#L605) is constructed using `np.zeros` (which [defaults](https://numpy.org/doc/stable/reference/generated/numpy.zeros.html) to `numpy.fp64`), failing the init for pytorch models with default fp32 precision. 

A couple of ideas to sort this out:

- Pass metadata about the inference model dtype during `AnchorImage` init, which will then be used  to construct data with correct dtype.
- Describe the output of the inference model (whether it is a class label or class probabilities) instead of letting `AnchorImage` trying to infer it (and we will not require a dummy call to the inference model during init, which sounds like a good idea). This is also related to https://github.com/SeldonIO/alibi/issues/492

jklaise:
Resolves #499.

What this does:
 - Expose an additional `dtype` argument for explainers calling black-box predictors with dummy `np.ndarray` objects (currently `AnchorImage`, `AnchorTabular`). The default is `np.float32` which should work for the vast majority of use cases.
 - Adds an extra test for `AnchorImage` using a pre-trained `torch` model with `torch.float32` types which tests the new behaviour

Out of scope:
 - There are some other explainers where this may become an issue, e.g. black-box predictors for the various types of counterfactual explainers. However, fixing this properly for all possible use cases may become quite complex. Also, given the discussion in https://github.com/SeldonIO/alibi/issues/499#issuecomment-937693639 for the vast majority of use cases it suffices to just define the `predictor` function appropriately, taking care of types.

To discuss:
 - Is the docstring for the new argument clear enough?
 - The default type is `np.float32` which is a `numpy` class, this is the most flexible way of specifying `numpy` types. However, string aliases exist (e.g. `np.dtype(np.float32)` and `np.dtype('float32')` are equivalent), but don't seem to be well documented. From the point of view of a client like `mlserver`, the types can be specified using `np.float32` in the runtime, but would probably have to resort to string aliases if running from a configuration. **EDIT**: a list of string aliases is available under `numpy.sctype`, see https://numpy.org/doc/stable/reference/arrays.dtypes.html (but this doesn't seem to be exhaustive, e.g. `float32` is missing!).
 - An even more flexible form would be asking directly for `np.dtype` instance, e.g.:
```python
def fun(dtype: np.dtype = np.dtype(np.float32))
```
this would allow for the most customization of exotic data types, but is probably overkill and it is unclear if internals of explainers could cope with arbitrary data types.

### DIFF
--- a/alibi/explainers/tests/test_anchor_image.py
+++ b/alibi/explainers/tests/test_anchor_image.py
@@ -157,3 +157,10 @@ def test_anchor_image(predict_fn, models, mnist_data):
     assert len(np.unique(explanation.segments)) == len(np.unique(sampler.segments))
     assert explanation.meta.keys() == DEFAULT_META_ANCHOR.keys()
     assert explanation.data.keys() == DEFAULT_DATA_ANCHOR_IMG.keys()
+
+
+@pytest.mark.parametrize('predict_fn', [lazy_fixture('models'), ], indirect=True)
+@pytest.mark.parametrize('models', [("mnist-cnn-pt1.9.1.pt",)], indirect=True)
+def test_anchor_image_fails_init_torch_float64(predict_fn, models):
+    with pytest.raises(RuntimeError):
+        explainer = AnchorImage(predict_fn, image_shape=(28, 28, 1), dtype=np.float64)  # noqa: F841


jklaise:
Added the shortest possible test to test failure to init `torch` model with wrong `dtype`.

Issue #516, opened 2021-10-28
Support more diverse input types for methods working on tabular data


jklaise:
## Current status

Currently tabular data in methods such as `AnchorTabular` and `CounterfactualProto` are expected to be in one of a set of restricted formats, e.g.
 - `np.ndarray` of homogeneous number types where numerical columns are `float` and categorical columns are integer-encoded  (to be pedantic, `float` that can be cast to `int` without loss of precision, e.g. `0.0` denoting the first category of a categorical feature).
 - `np.ndarray` of homogeneous number types where numerical columns are `float` and categorical features have been expanded into one-hot-encoded columns (i.e. each categorical feature now occupies `n_categories` columns which are populated with `0` and `1` entries)

## Problem

If a user model is not trained on a data representation that is one of the above then Alibi tabular explainers cannot be used out-of-the-box which is undesirable (as found out by @FarrandTom).

For concreteness, denote by `X` an input data point that is **non-compliant** with the Alibi API, e.g. it could be `np.ndarray` but with unsupported column types, for example `array([49.5, 'Male'], dtype=object)` representing a numerical feature and a string-encoded categorical variable.

Further, denote by `Z` an input data point that is **compliant** with the Alibi API, e.g. ` array([49.5,  0. ])` representing the same numerical feature and the same but integer-encoded categorical variable.

A client may have a model `M` that's trained on non-compliant data, i.e. it would be of type `Callable[[X], np.ndarray]`, whereas Alibi expects a model `M_hat` (prediction function) of type `Callable[[Z], np.ndarray]`. How can we go from a non-compliant model to a compliant one?

The key is being able to map back and forth between `X` and `Z`. Let `f: X->Z` be such an invertible mapping, for the example above it would be something like:
```python
def f(X: np.ndarray, **kwargs) -> np.ndarray:  # use **kwargs for any other information needed to do the conversion
    Z_num = extract_numeric(X, **kwargs)  # extract columns like 49.5,  Z_num is now a homogenous array of numbers
    Z_cat = extract_cat(X, **kwargs)  # take columns like 'Male' and convert to 0, Z_cat is now a homogenous array of numbers
    Z = combine(Z_num, Z_cat, **kwargs)  # concatenate columns in the right order
    return Z

def f_inverse(Z: np.ndarray, **kwargs) -> np.ndarray:
    ... # do similar operations as above
    return Z
```

[...]

PR #520, opened 2021-11-03
Add custom exceptions for calling and validating user models

jklaise:
This PR introduces a new public `alibi.exceptions` module defining a (small) hierarchy of general exceptions to be used across `alibi`. The motivation is to provide nicer messages when things go wrong and also well-defined exceptions which can be handled by the caller. This could be considered an adjacent building block towards #516.

The initial application of this is 2 fold:
 - Catching and raising errors when calling user provided models with dummy data (currently `AnchorTabular`, `AnchorText`, `AnchorImage`)
 - Validating that the output type of the call (if successful) is of the expected type

The first use case now clearly defines to the caller what error to expect and when it is raised. I've raised this when the call to a user provided model fails during `__init__`, but not in other methods as it could be argued that in `fit` and `explain` we are calling the model with user provided data (and perturbations thereof) so it should work fine (if there is a compelling reason to introduce `try/except` on calling the user model elsewhere in the code then that can be done later).

The second use case is, if a call succeeds, we can check whether the return is of the expected type so an error can be raised early if it is not. I opted for an explicit `isinstance` check rather than something of the EAFP style `try: prediction.shape`. This has the side-effect that we tighten the return type expected to be exactly `np.ndarray` as opposed to anything that implements `.shape` attribute (for example `tensorflow` tensors). I don't think this is a big deal as other tensor/array types that may implement `.shape` should not work with `alibi` currently anyway, so it's good to tighten up the requirements at runtime.

The PR should be ready to review as is with a few minor points outstanding:
 - [ ] There are a few other explainers which call the `predictor` at `__init__` with dummy data (e.g. `Countefactual`) which should be treated the same way
 - [ ] In docstrings we should use the `sphinx` directives to link to the custom exception definitions in the code.

### DIFF
--- a/alibi/explainers/anchor_image.py
+++ b/alibi/explainers/anchor_image.py
@@ -9,6 +9,7 @@ from typing import Any, Callable, List, Union, Tuple, Type
 from alibi.utils.wrappers import ArgmaxTransformer
 from alibi.api.interfaces import Explainer, Explanation
 from alibi.api.defaults import DEFAULT_META_ANCHOR, DEFAULT_DATA_ANCHOR_IMG
+from alibi.exceptions import AlibiPredictorCallException, AlibiPredictorReturnTypeError
 from .anchor_base import AnchorBaseBeam
 from .anchor_explanation import AnchorExplanation
 from skimage.segmentation import felzenszwalb, slic, quickshift
@@ -332,6 +333,11 @@ class AnchorImage(Explainer):
             Images to overlay superpixels on.
         seed
             If set, ensures different runs with the same input will yield same explanation.
+
+        Raises
+        ------
+        alibi.exceptions.AlibiPredictorCallException
+            If calling `predictor` fails at runtime.
         """
         super().__init__(meta=copy.deepcopy(DEFAULT_META_ANCHOR))
         np.random.seed(seed)
@@ -613,7 +619,19 @@ class AnchorImage(Explainer):
     def _transform_predictor(self, predictor: Callable) -> Callable:
         # check if predictor returns predicted class or prediction probabilities for each class
         # if needed adjust predictor so it returns the predicted class
-        if np.argmax(predictor(np.zeros((1,) + self.image_shape, dtype=self.dtype)).shape) == 0:
+        x = np.zeros((1,) + self.image_shape, dtype=self.dtype)
+        try:
+            prediction = predictor(x)
+        except Exception as e:
+            msg = f"Predictor failed to be called on {type(x)} of shape {x.shape} and dtype {x.dtype}. " \
+                  f"Check that the parameter `image_shape` is correctly specified."
+            raise AlibiPredictorCallException(msg) from e
+
+        if not isinstance(prediction, np.ndarray):
+            msg = f"Excepted predictor return type to be {np.ndarray} but got {type(prediction)}."
+            raise AlibiPredictorReturnTypeError(msg)
+
+        if np.argmax(prediction.shape) == 0:
             return predictor
         else:
             transformer = ArgmaxTransformer(predictor)
--- a/alibi/explainers/anchor_image.py
+++ b/alibi/explainers/anchor_image.py
@@ -338,6 +338,8 @@ class AnchorImage(Explainer):
         ------
         alibi.exceptions.AlibiPredictorCallException
             If calling `predictor` fails at runtime.
+        alibi.exceptions.AlibiPredictorReturnTypeError
+            If the return type of `predictor` is not `np.ndarray`.
         """
         super().__init__(meta=copy.deepcopy(DEFAULT_META_ANCHOR))
         np.random.seed(seed)
--- a/alibi/exceptions.py
+++ b/alibi/exceptions.py
@@ -1,0 +1,29 @@
+"""
+This module defines the Alibi exception hierarchy and common exceptions
+used across the library.
+"""
+from abc import ABC
+
+
+class AlibiException(Exception, ABC):
+    """
+    Abstract base class of all alibi exceptions.
+    """
+
+    def __init__(self, message: str) -> None:
+        super().__init__(message)
+
+
+class AlibiPredictorCallException(AlibiException):
+    """
+    This exception is raised whenever a call to a user supplied predictor fails at runtime.
+    """
+    pass
+
+
+class AlibiPredictorReturnTypeError(AlibiException):
+    """
+    This exception is raised whenever the return type of a user supplied predictor is of
+    an unexpected or unsupported type.
+    """
+    pass
--- a/alibi/explainers/anchor_tabular.py
+++ b/alibi/explainers/anchor_tabular.py
@@ -7,6 +7,7 @@ from typing import Any, Callable, DefaultDict, Dict, List, Set, Tuple, Type, Uni
 
 from alibi.api.interfaces import Explainer, Explanation, FitMixin
 from alibi.api.defaults import DEFAULT_META_ANCHOR, DEFAULT_DATA_ANCHOR
+from alibi.exceptions import AlibiPredictorCallException, AlibiPredictorReturnTypeError
 from .anchor_base import AnchorBaseBeam, DistributedAnchorBaseBeam
 from .anchor_explanation import AnchorExplanation
 from alibi.utils.wrappers import ArgmaxTransformer
@@ -684,6 +685,11 @@ class AnchorTabular(Explainer, FitMixin):
             assumed to have ordinal encodings.
         seed
             Used to set the random number generator for repeatability purposes.
+
+        Raises
+        ------
+        alibi.exceptions.AlibiPredictorCallException
+            If calling `predictor` fails at runtime.
         """
         super().__init__(meta=copy.deepcopy(DEFAULT_META_ANCHOR))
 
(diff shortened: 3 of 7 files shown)


jklaise:
Unrelated to the content of this PR, but looks like the `tensorflow` 2.6.1 release 2 days ago now results in some errors collecting tests (and potentially runtime errors...).

sakoush (commented):
nice. Is there a way to wire up exceptions as well for cases where we are not doing the dummy call? e.g. `KernelShap`? 

It feels we are sticky with the dummy call approach longer term, which is fine but just highlighting that if we want to get rid of it we might consider doing the checks differently?

jklaise:
@sakoush good point, but I don't think there is another way to validate the user callable without doing some kind of dummy call. Since we're doing it for `Anchor*` it makes sense to augment it with these exceptions. For `KernelShap` if we wanted to validate the callable in the same way we would have to make the dummy call... (One option which I don't particularly like is deferring validation until `fit/explain` where we use the user supplied data, but this introduces complexity in the `alibi` code).

alibi/explainers/anchor_image.py
```diff
@@ -613,7 +621,19 @@ def overlay_mask(self, image: np.ndarray, segments: np.ndarray, mask_features: l
     def _transform_predictor(self, predictor: Callable) -> Callable:
         # check if predictor returns predicted class or prediction probabilities for each class
         # if needed adjust predictor so it returns the predicted class
-        if np.argmax(predictor(np.zeros((1,) + self.image_shape, dtype=self.dtype)).shape) == 0:
+        x = np.zeros((1,) + self.image_shape, dtype=self.dtype)
+        try:
+            prediction = predictor(x)
+        except Exception as e:
```
sakoush, line 627: another point here is that in mlserver / alibi explainer server v1 we are wrapping a `[predict_fn`](https://github.com/SeldonIO/MLServer/blob/07a36f2cacf4eea6ed3ac28436792f31adb04d01/runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py#L57) around a REST call. So there might be other exceptions that are raised with the connection. I am not sure though how we will identify the differences here. 

interesting in the mlserver runtime it will raise this [exception](https://github.com/SeldonIO/MLServer/blob/07a36f2cacf4eea6ed3ac28436792f31adb04d01/runtimes/alibi-explain/mlserver_alibi_explain/common.py#L50).

not sure about the interop between libraries and how to deal with exceptions in this case.

jklaise: Good point, there must be a way around this... I guess the issue is that with these semantics `alibi` doesn't know what can go wrong calling a user model, so all exceptions are caught and re-raised with an error that assumes something about the model is misspecified. But for clients such as `mlserver` they would like to distinguish between these types of errors and network errors.

jklaise: Can we test the behaviour of this? I believe it could still work as is as the remote request shouldn't raise a network error but rather an error code, in which case if an error is raised it must have come from the model internals.

sakoush: I am raising `ValueError` and here is what we get.

```
args = ()
kwargs = {'predictor_url': 'dummy_call', 'v2_payload': InferenceRequest(id=None, parameters=Parameters(content_type='np'), inpu... 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]))], outputs=None)}
loop = <_UnixSelectorEventLoop running=False closed=False debug=False>
res = InferenceResponse(model_name='custom_tf_mnist_model', model_version=None, id=None, parameters=None, outputs=[ResponseO...274435043, 0.10133041441440582, 0.09029622375965118, 0.11116134375333786, 0.07727331668138504, 0.09395473450422287]))])

    def mock_predict(*args, **kwargs):
        # note: sometimes the event loop is not running and in this case
        # we create a new one otherwise
        # we use the existing one.
        # this mock implementation is required as we dont want to spin up a server,
        # we just use MLModel.predict
        try:
            loop = asyncio.get_event_loop()
            res = loop.run_until_complete(
                custom_runtime_tf.predict(kwargs["v2_payload"])
            )
            raise ValueError("break")
            return res
        except Exception:
            loop = asyncio.new_event_loop()
            asyncio.set_event_loop(loop)
            res = loop.run_until_complete(
                custom_runtime_tf.predict(kwargs["v2_payload"])
            )
>           raise ValueError("break")
E           ValueError: break

conftest.py:161: ValueError

The above exception was the direct cause of the following exception:

args = ()
kwargs = {'anchor_image_directory': PosixPath('/home/sa/code/MLServer/runtimes/alibi-explain/tests/data/mnist_anchor_image'), 'custom_runtime_tf': <helpers.tf_model.TFMNISTModel object at 0x7f6936454f50>}
loop = <_UnixSelectorEventLoop running=False closed=False debug=False>
setup = <function pytest_fixture_setup.<locals>.wrapper.<locals>.setup at 0x7f6a84016830>

    def wrapper(*args, **kwargs):
        loop = fixture_stripper.get_and_strip_from(
            FixtureStripper.EVENT_LOOP, kwargs
        )
    
        async def setup():
            res = await coro(*args, **kwargs)
            return res
    
>       return loop.run_until_complete(setup())

/home/sa/miniconda3/envs/python3.7-mlserver/lib/python3.7/site-packages/pytest_asyncio/plugin.py:160: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/sa/miniconda3/envs/python3.7-mlserver/lib/python3.7/site-packages/nest_asyncio.py:70: in run_until_complete
    return f.result()
/home/sa/miniconda3/envs/python3.7-mlserver/lib/python3.7/asyncio/futures.py:181: in result
    raise self._exception
/home/sa/miniconda3/envs/python3.7-mlserver/lib/python3.7/asyncio/tasks.py:249: in __step
    result = coro.send(None)
/home/sa/miniconda3/envs/python3.7-mlserver/lib/python3.7/site-packages/pytest_asyncio/plugin.py:157: in setup
    res = await coro(*args, **kwargs)
conftest.py:177: in anchor_image_runtime_with_remote_predict_patch
    await rt.load()
../mlserver_alibi_explain/runtime.py:165: in load
    return await self._rt.load()
../mlserver_alibi_explain/explainers/black_box_runtime.py:43: in load
    self._model = self._load_from_uri(self._infer_impl)
../mlserver_alibi_explain/runtime.py:91: in _load_from_uri
    return load_explainer(uri, predictor=predictor)
/home/sa/miniconda3/envs/python3.7-mlserver/lib/python3.7/site-packages/alibi/saving.py:68: in load_explainer
    return load_fn(path, predictor, meta)
/home/sa/miniconda3/envs/python3.7-mlserver/lib/python3.7/site-packages/alibi/saving.py:155: in _load_AnchorImage
    explainer.reset_predictor(predictor)
/home/sa/miniconda3/envs/python3.7-mlserver/lib/python3.7/site-packages/alibi/explainers/anchor_image.py:643: in reset_predictor
    self.predictor = self._transform_predictor(predictor)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = AnchorImage(meta={
  'name': 'AnchorImage',
  'type': ['blackbox'],
  'explanations': ['local'],
  'params': {
       ...           'images_background': None,
              'segmentation_fn': 'slic'}
            ,
  'version': '0.6.2dev'}
)
predictor = <bound method AlibiExplainBlackBoxRuntime._infer_impl of <mlserver_alibi_explain.explainers.black_box_runtime.AlibiExplainBlackBoxRuntime object at 0x7f6935b20ad0>>

    def _transform_predictor(self, predictor: Callable) -> Callable:
        # check if predictor returns predicted class or prediction probabilities for each class
        # if needed adjust predictor so it returns the predicted class
        x = np.zeros((1,) + self.image_shape, dtype=self.dtype)
        try:
            prediction = predictor(x)
        except Exception as e:
            msg = f"Predictor failed to be called on {type(x)} of shape {x.shape} and dtype {x.dtype}. " \
                  f"Check that the parameter `image_shape` is correctly specified."
>           raise AlibiPredictorCallException(msg) from e
E           alibi.exceptions.AlibiPredictorCallException: Predictor failed to be called on <class 'numpy.ndarray'> of shape (1, 28, 28, 1) and dtype float32. Check that the parameter `image_shape` is correctly specified.

/home/sa/miniconda3/envs/python3.7-mlserver/lib/python3.7/site-packages/alibi/explainers/anchor_image.py:630: AlibiPredictorCallException
```

ascillitoe: @sakoush unless I'm misunderstanding, the above result looks good? As you are successfully raising the new `alibi` exception?

ascillitoe:
@jklaise these look nice to me. I'm assuming we'd add more exceptions as we go? `AlibiFitCallException` etc?

alibi/explainers/anchor_image.py
jklaise: I believe the point is that `mlserver` would like to catch specific errors e.g. `ValueError` from the scope calling the predictor, but because inside `alibi` all `Exception` are caught and re-raised with `AlibiPredictorCallException` this is not possible? Would be good to clarify again what the goals here are.

jklaise:
> @jklaise these look nice to me. I'm assuming we'd add more exceptions as we go? `AlibiFitCallException` etc?

Yes, we can also create a more granular hierarchy if needed in the future so callers can catch more general `alibi` errors.

alibi/explainers/anchor_image.py
jklaise: @sakoush wouldn't this be solved by doing `except ValueError` instead? `mlserver` shouldn't really be catching bare exceptions and if has error handling for both `ValueError` and `AlibiPredictorCallException` we should be good to go.

sakoush: We are not actually capturing exceptions, what we do in `mlserver` when we get an error from calling the endpoint for `predict_fn` is 
```
    if response_raw.status_code != 200:
        raise RemoteInferenceError(response_raw.status_code, response_raw.reason)
```
we probably can be more specific for the error codes but I dont think that the statement `alibi.exceptions.AlibiPredictorCallException: Predictor failed to be called on <class 'numpy.ndarray'> of shape (1, 28, 28, 1) and dtype float32. Check that the parameter `image_shape` is correctly specified.` is valid generally. I need to think more about this from my side, basically how we can propagate exceptions interprocess?

jklaise:
I will go ahead with merging this unless any further comments? @ascillitoe @sakoush

ascillitoe (approved):
Looks good!

Issue #492, opened 2021-10-04
Allow `IntegratedGradients.target` to be inferred for classification models with class score


sakoush:
As part of ongoing work to integrate `alibi` runtime in `mlserver`, we realised that for `IntegratedGradients` if the inference model outputs a class score (e.g mnist model handwritten digit recognition), a target index is required to be passed to `IntegratedGradients.explain` 

i.e.

`target = model(X_test_sample).numpy().argmax(axis=1)`

I see why it is required but at the same time I am not sure how we can easily handle this kind of custom code in production / deployment. 

`seldon-core` v1 does have it fixed in the IG explainer model wrapper but it is restrictive i.e. we are not sure it will generalises to all use cases.

Having a discussion with @jklaise  we could consider some changes to alibi to make it easier in this case.

jklaise:
Having thought more about it, it seems to me the cleanest and most Pythonic way to solve this is via callbacks. Instead of extending the definition of `target` to take a restricted set of strings such as `argmax` that map to internal functions, allow the user passing the function in question directly (and maybe disambiguate the semantics by introducing another kwarg `target_fn` with the expectation that only one of `target` or `target_fn` can be not `None`). In this example `target_fn=np.argmax` (module some `partial`s with axis), but the only requirement on `target_fn` is that it maps (batched) vectors to scalars.

 (There is a lot of precedent for callbacks as it leads to powerful Pythonic code, but in particular, [captum](https://captum.ai/api/neuron.html) use this for their `Neuron*` explainers in exactly the same way - user defined callable that selects the neuron of interest within the layer.

Of course, this brings us to the old issue of how to specify callbacks from outside of Python, e.g. in a config file. It would be silly for someone to pickle a built-in `np.argmax` function just so that together with a config file the explainer can be fully spec'ed. This leads me to think more that we should embrace [catalogue](https://github.com/explosion/catalogue) internally, create internal `alibi` registries for functions that can be used as callbacks in `alibi` code and then enable users to refer to them via their registered strings either in config files (which resolve to the `alibi` function at runtime) or accepting the registered strings to the explainer constructor.

Would be keen to hear your thoughts @ascillitoe @sakoush @arnaudvl.

ascillitoe:
I agree, `target_fn` seems like a really nice option. The similarity with `preprocess_fn` in `alibi-detect` means we can of course reuse lots of the functionality between the two libraries, but more importantly I'd hope the similarity across the libraries would improve the user experience here.

[...]

PR #523, opened 2021-11-09
Integrated Gradients `target_fn`

jklaise:
This is a PR partially addressing #492. In particular, it introduces the ability to pass a callback via the `target_fn` argument to calculate the scalar target dimension from the model output within the method. This is to bypass the requirement of passing `target` directly to `explain` when the `target` of interest may depend on the prediction output as in the case of the highest probability class.

Now both of these are possible to define an explainer:
```python
# current
ig = IntegratedGradients(model)
preds = np.argmax(model(X), axis=1)  # need to define upfront
exp = ig.explain(X, target=preds)

# proposed
from functools import partial
target_fn = partial(np.argmax, axis=1)  # user defined
ig = IntegratedGradients(model, target_fn=target_fn)
exp = ig.explain(X) # no preds needed
```

**TODO:**
- [x] Validate the output of `target_fn`
- [x] Tests
- [x] Docs updates
- [x] Example updates
- [ ] (Out of scope?) Desired output shape of `target_fn`. If the output shape of the model is `(N,M)` as in probabilistic classification, then naturally the output of `target_fn` is expected to be an `(N,)` array of integers which specify, for each instance in the batch, which dimension of the output (e.g. maximum probabilitie) to take. However, if the output shape of the model is `>2` this would not apply, e.g. if the output shape of the model is `(N,M,K)`, the `target_fn` would have to specify the indices for both `M` and `K` dimensions which would be reflected in the shape of `target_fn`. That being said, I'm not certain our current implementation of `IntegratedGradients` supports `>2D` model outputs so the point may be moot (for now).

**To discuss:**

1. I put `target_fn` as an argument to `__init__` as it seemed more natural, however, a case could be made that his belongs in `explain` as `target` is part of explain ~(and so is `forward_kwargs`—although I would personally like to see this shift to `__init__`)~ EDIT: Actually `forward_kwargs` is data dependent so belongs in `explain`.
2. This PR does not resolve the issue of fully specifying the `target_fn` from outside Python—in particular, the user application must define it. We could approach this (outside of the scope of this PR) by introducing function registries within `alibi` itself for pre-defined functions using [catalogue](https://github.com/explosion/catalogue). It would then look something like this:

```python
# alibi.utils.registries
import catalogue
scalar_reducers = catalogue.create("alibi", "scalar_reducers)

@scalar_reducers.register("argmax")
def argmax(X):
    return np.argmax(X, axis=...)  # TODO: numpy doesn't support multi-axis reduction for argmax
```

```python
# user code option 1
# this does not fully solve the issue in #492 as the user still has to fetch the function
from alibi.utils.registries import scalar_reducers
target_fn = scalar_reducers['argmax']
explainer = IntegratedGradients(model, target_fn=target_fn)
```
```python
# user code option 2
# this fully solves the issue in #492 by extending the explainer constructor signature
# to accept strings which are looked up internally in the register
target_fn = "argmax"  # or provided from some config file
explainer = IntegratedGradients(model, target_fn=target_fn)

# inside IntegratedGradients:
    ...
    target_fn = scalar_reducers[target_fn]
```

We can separate the registry discussion in a separate issue in the future.

It would be nice imo to provide both callback functionality for maximum flexibility as well as strings mapping to `alibi` built-in functions for ease of use (in this example avoiding the use of `partial` and defining `target_fn` manually). Although, we may want to think about separating interfaces that act on pure callbacks (e.g. the public `IntegratedGradients` interface) and interfaces that act on string names to construct an explainer for greater clarity and separation of concerns.

### DIFF
--- a/alibi/explainers/integrated_gradients.py
+++ b/alibi/explainers/integrated_gradients.py
@@ -583,8 +583,8 @@ def _get_target_from_target_fn(target_fn: Callable,
     # raise a warning if the predictions are scalar valued already
     # TODO: in the future we want to support outputs that are >2D at which point this check should change
     if preds.shape[-1] == 1:
-        msg = f"Predictions from the model are scalar valued but `target_fn` was passed. `target_fn` is not necessary" \
-              f"when predictions are scalar valued already. Using `target_fn` here may result in unexpected behaviour."
+        msg = "Predictions from the model are scalar valued but `target_fn` was passed. `target_fn` is not necessary" \
+              "when predictions are scalar valued already. Using `target_fn` here may result in unexpected behaviour."
         warnings.warn(msg)
 
     target = target_fn(preds)
@@ -815,7 +815,7 @@ class IntegratedGradients(Explainer):
         """
         # target handling logic
         if self.target_fn and target is not None:
-            msg = f'Both `target_fn` and `target` were provided. Only one of these should be provided.'
+            msg = 'Both `target_fn` and `target` were provided. Only one of these should be provided.'
             raise ValueError(msg)
         if self.target_fn:
             target = _get_target_from_target_fn(self.target_fn, self.model, X, forward_kwargs)
--- a/alibi/explainers/integrated_gradients.py
+++ b/alibi/explainers/integrated_gradients.py
@@ -1,13 +1,15 @@
 import copy
 import logging
-import numpy as np
 import string
+import warnings
+from typing import Callable, List, Optional, Tuple, Union, cast
+
+import numpy as np
 import tensorflow as tf
 
 from alibi.api.defaults import DEFAULT_DATA_INTGRAD, DEFAULT_META_INTGRAD
-from alibi.utils.approximation_methods import approximation_parameters
 from alibi.api.interfaces import Explainer, Explanation
-from typing import Callable, Union, List, Tuple, Optional, cast
+from alibi.utils.approximation_methods import approximation_parameters
 
 logger = logging.getLogger(__name__)
 
@@ -577,6 +579,13 @@ def _get_target_from_target_fn(target_fn: Callable,
         preds = model(X)
     else:
         preds = model(X, **forward_kwargs)
+
+    # raise a warning if the predictions are scalar valued already
+    # TODO: in the future we want to support outputs that are >2D at which point this check should change
+    if preds.shape[-1] == 1:
+        msg = f"Predictions from the model are scalar valued but `target_fn` was passed. `target_fn` is not necessary" \
+              f"when predictions are scalar valued already. Using `target_fn` here may result in unexpected behaviour."
+        warnings.warn(msg)
 
     target = target_fn(preds)
     expected_shape = (target.shape[0],)
--- a/alibi/explainers/integrated_gradients.py
+++ b/alibi/explainers/integrated_gradients.py
@@ -553,14 +553,37 @@ def _format_target(target: Union[None, int, list, np.ndarray],
 def _get_target_from_target_fn(target_fn: Callable,
                                model: tf.keras.Model,
                                X: Union[np.ndarray, List[np.ndarray]],
-                               forward_kwargs: Optional[dict] = None) -> List[int]:
+                               forward_kwargs: Optional[dict] = None) -> np.ndarray:
+    """
+    Generate a target vector by using the `target_fn` to pick out a
+    scalar dimension from the predictions.
+
+    Parameters
+    ----------
+    target_fn
+        Target function.
+    model
+        Model
+    X
+        Data to be explained.
+    forward_kwargs
+        Any additional kwargs needed for the model forward pass.
+
+    Returns
+    -------
+    Integer array of dimension (N, ).
+    """
     if forward_kwargs is None:
         preds = model(X)
     else:
         preds = model(X, **forward_kwargs)
 
     target = target_fn(preds)
-    # TODO: validate target here before returning
+    expected_shape = (target.shape[0],)
+    if target.shape != expected_shape:
+        # TODO: in the future we want to support outputs that are >2D at which point this check should change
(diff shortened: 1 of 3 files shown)


jklaise:
@sakoush @ascillitoe @arnaudvl ready for review.

This PR **does not** address the issue of specifying the target via e.g. a string key outside of Python, this can be explored in a follow-up work.

I've kept the example notebooks unchanged but introduced the new usage in the method overview docs. The tests are covering now passing either `targets` or `target_fn`, although for the time being I am not considering test cases where either both or neither are passed due to the complexity of internal test logic.

alibi/explainers/integrated_gradients.py
```diff
@@ -663,6 +713,7 @@ class IntegratedGradients(Explainer):
     def __init__(self,
                  model: tf.keras.Model,
                  layer: Optional[tf.keras.layers.Layer] = None,
+                 target_fn: Optional[Callable] = None,
```
sakoush, line 716: could we have `partial(np.argmax, axis=1)` as default instead?

alibi/explainers/integrated_gradients.py
```diff
@@ -761,6 +814,13 @@ def explain(self,
             for each feature.
 
         """
+        # target handling logic
+        if self.target_fn and target is not None:
```
sakoush, line 818: another way to look at it is that  `target` takes precedence I am just think that the user might want to change this at explain time without having to create the expaliner object again?

alibi/explainers/integrated_gradients.py
```diff
@@ -722,6 +773,8 @@ def __init__(self,
         self._is_np: Optional[bool] = None
         self.orig_dummy_input: Optional[Union[list, np.ndarray]] = None
 
+        self.target_fn = target_fn
```
sakoush, line 776: I personally do not want to add more dummy calls,

but in this case we have the keras model and target_fn if set, we could check that this is ok at init time?

I guess you left it to explain time to have the check in one place but perhaps if we check `target_fn` this is towards checking the parameters passed and providing some hints back to the use if things are good.

sakoush (approved):
I left some comments regarding whether it is not going to be a big change to allow a default for `target_fn` that is not None. If so the explainer can be init on the fly which is I guess useful to have as discussed.

I understand that this will affect downstream logic and testing as you explained but I thought I will mention it.

In any case when we get the registry based / string config this can be addressed so no real urgency I think.

alibi/explainers/integrated_gradients.py
jklaise: Unfortunately not as `target_fn` depends on the output shape of the predictions which can be anything, so it's up to the user to specify, we cannot safely assume a default here.

alibi/explainers/integrated_gradients.py
jklaise: That's a good point, I didn't want to complicate the logic, but we could also have `target` override any pre-set `target_fn`, a warning would need to be emitted. Let me think about it.

alibi/explainers/integrated_gradients.py
jklaise: I think this is quite hard without not knowing what the shape of the inputs is which we only know at `explain` time. If we wanted to deeply validate user models and callables at construction time it would involve a sizable refactor as we would also need to ask the user for the type of data that can be put through the model.

Practically, this could be solved by asking a user to provide a sample batch of data at `__init__` that we can use to validate the model and any callables. In theory this could be an optional `kwarg` so that people would only use it when they want some guarantees about the functionality of the explainer post `__init__` (we wouldn't want to make it mandatory, I think).

I think this should be discussed in a different place (maybe a new issue?). Would be keen to hear thoughts from @ascillitoe and @arnaudvl.

ascillitoe: Agree with the sentiments of @sakoush and @jklaise here; as @sakoush hints at, any checks we can do to enable more helpful error messages are great. But as @jklaise alludes to, the complexity needed for this quickly ramps up if we don't assume the inputs pass a certain sanity threshold to begin with. This is made even more challenging by the number of ways tensorflow and pytorch models can fail!

Providing a sample batch of data could be a nice solution, but it might be a little unintuitive. If we were going to go down this route, perhaps it should be a well-documented optional functionality? i.e. have an option to provide sample data so that `target_fn` can be checked at init.

alibi/explainers/integrated_gradients.py
ascillitoe: Having detectors/explains which can be modified post-init generally seems like a nice idea (especially if init is expensive), but it does open up a little bit of a question of where to stop with this. For example, for many (or all?) of the offline detectors we could actually have `p_val` as a kwarg of `predict`. Perhaps this is best dealt with by opening new issue(s)?

ascillitoe (approved):
This looks good to me. There appear to be two wider design questions to consider (which I've left comments on), but I wonder if these are best left as future issues?

alibi/explainers/integrated_gradients.py
jklaise: Yea, it's a design discussion. We've had similar discussions in the past, the line between what you would want to change on every `explain` call vs what you want fixed is very fuzzy.

alibi/explainers/integrated_gradients.py
sakoush: If we have the keras model + layer, we can get the shape of the input data shape ((i.e. `model.get_layer(layer).input_shape`)) and therefore able to construct a valid dummy? 

This could also help constructing target_fn if we just refer to `argmax` as example computation.

but yeah out of scope of this PR perhaps.

jklaise: If I remember correctly, I think the `input_shape` being present is limited to a subset of `tensorflow` models as the model may be in a state where `input_shape` is unknown (there are some hints of this in the `integrated_gradients.py` module already). More generally, thinking ahead, for `torch` models input shapes are generally unknown unless provided by the user.

Issue #524, opened 2021-11-09
Bug report + fix: forward_kwargs not used in all self.model calls in integrated_gradients


sanjass:
In lines [808](https://github.com/SeldonIO/alibi/blob/master/alibi/explainers/integrated_gradients.py#L808), [856](https://github.com/SeldonIO/alibi/blob/master/alibi/explainers/integrated_gradients.py#L856) and [926](https://github.com/SeldonIO/alibi/blob/master/alibi/explainers/integrated_gradients.py#L926) in `integrated_gradients.py` the `forward_kwargs` argument is missing from the `self.model` function calls. 

The following change resolved the issue for me: `self.model(X)` -> `self.model(X, **forward_kwargs)`. Note in some cases `X` is `inputs`.

jklaise:
Good catch and thanks for the PR!

Although now I wonder how our (transformers example)[https://docs.seldon.io/projects/alibi/en/latest/examples/integrated_gradients_transformers.html] worked with this bug... @gipster

sanjass:
In the first example of [transformers example](https://docs.seldon.io/projects/alibi/en/latest/examples/integrated_gradients_transformers.html)
 the `self.model` call works even if the `attention_mask` is `None`, so it wasn't throwing any errors despite `forward_kwargs` not being used. You can confirm this by printing out the `attention_mask` in the model `call`; sometimes it's `None`, sometimes it's what it's supposed to be (bc `forward_kwargs` is used only in some calls). 

In my case, my model had `Input` layers so a `None` input for `attention_mask` (and `input_type_ids` for that matter) wouldn't work and I was getting errors such as `TypeError: Inputs to a layer should be tensors. Got: None`. With the above change it worked as expected in my case.

sanjass:
@jklaise thanks for assisting with this issue! I see the PR has been merged to master, but is there an estimate when would the fix be released, so it works after `pip install alibi`? I see the "licences" CI check fails, is there anything I can do about that?

jklaise:
@sanjass we can release a patch release today or tomorrow, in the meantime you can install from github directly via ```pip install git+https://github.com/SeldonIO/alibi.git 
```

The licences is an easy fix which we'll do prior releasing, no need to worry about that :)

jklaise:
Fixed by #525.

PR #525, opened 2021-11-09
Bug fix: Include forward_args in self.model function calls in intergrated_gradients.py

sanjass:
Bug report + fix: `forward_kwargs` not used in all `self.model` calls in `integrated_gradients.py`
https://github.com/SeldonIO/alibi/issues/524

### DIFF
--- a/alibi/explainers/integrated_gradients.py
+++ b/alibi/explainers/integrated_gradients.py
@@ -422,7 +422,6 @@ def _gradients_layer(model: Union[tf.keras.models.Model],
         Gradients for each element of layer.
 
     """
-    
     def watch_layer(layer, tape):
         """
         Make an intermediate hidden `layer` watchable by the `tape`.
--- a/alibi/explainers/integrated_gradients.py
+++ b/alibi/explainers/integrated_gradients.py
@@ -49,6 +49,8 @@ def _compute_convergence_delta(model: Union[tf.keras.models.Model],
     -------
         Convergence deltas for each data point.
     """
+    if forward_kwargs is None:
+        forward_kwargs = {}
     if _is_list:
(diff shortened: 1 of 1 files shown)


jklaise:
@sanjass the tests are failing because we also pass in `forward_kwargs` when it is `None`. We would need to add the same logic to these calls as here: https://github.com/SeldonIO/alibi/blob/610013595ad2da40286a22ebca03bd5b5ffe25e4/alibi/explainers/integrated_gradients.py#L143-L146

Given the repetition I think it would be good to split this out into a small utility function (or actually, would be good to check if we could just substitute `None` for the empty dictionary `{}` and it might run without having to do this check).

sanjass:
I see, I pushed a change as follows:
```
if forward_kwargs is None:
    forward_kwargs = {}
```
and replaced `If forward_kwargs is not None`-> `if forward_kwargs`
also removed `if forward_kwargs is None` and passed `**forward_kwargs` as before

Didn't get to run the checks though

jklaise:
Thanks! Changes look good, I will rerun the IG notebooks (especially the transformers one) before merging.

jklaise:
@sanjass just a recommendation, if you also do `pip install -r requirements/dev.txt` and run `flake8 alibi` from the repo source you can see all the linting errors and fix them locally :)

sanjass:
Yeah, did something like that but my previous fix introduced a different flake8 error and I didn't double check lol. Also I'm editing in the github editor directly. Anyways hopefully this finally works

RobertSamoilescu:
@sanjass we will be working on issue #527 before merging this PR.

jklaise (approved):

jklaise:
@sanjass thank you for your contribution! And thanks for the extra review @RobertSamoilescu!

Issue #509, opened 2021-10-21
Update `shap>=0.40.0`


jklaise:
It looks like with the [latest release](https://github.com/slundberg/shap/releases/tag/v0.40.0) a `pyproject.toml` files has appeared specifying `oldest-supported-numpy` as a build dependency thus resolving the long-standing issue of broken `shap` builds.

What happens is that `pip install shap` prioritizes the `pyproject.toml` file over `setup.py` to build the package which is good, although it is slightly concerning that the original `setup.py` has not been updated to `oldest-supported-numpy` for `setup_requires` (see https://github.com/slundberg/shap/pull/1802) which may at best cause confusion and at worst result in broken build for editable installs (`pyproject.toml` famously doesn't understand `pip install -e .` and falls back on `setup.py` - although having tested it, it still seems to result in a package built with an older `numpy` version, though I'm not sure I know why/how).

All in all I think it is safe to do the update to `shap>=0.40.0`. This would also enable us to simplify the install instructions for `shap` from `pip install alibi && pip install alibi[shap]` to the more straightforward `pip install alibi[shap]`.

jklaise:
Closed by #533.

PR #533, opened 2021-11-17
Bump shap minimum version

jklaise:
Addressing #509.

I've also removed the pin on `numba` as it shouldn't be an issue with `shap>=0.40.0`.


jklaise:
Ran through the notebooks as an extra sanity check with `shap 0.40.0` and both `numba 0.54.0` and `numba 0.54.1` and everything works.

PR #541, opened 2021-11-24
Dev/reinstate strict optional mypy

jklaise:
This PR reinstates strict `Optional` type-checking with `mypy` because it's both a prerequisite of #511 and also a good idea (there were multiple genuine bugs caught by re-enabling strict behaviour).

Specifically, `setup.cfg` now contains the following:
```
[mypy]
ignore_missing_imports = True
no_implicit_optional = True
```

Note that `strict_optional = True` is the default and disallows `None` as a valid type for functions that don't declare it explicitly as valid. Furthermore, `no_implicit_optional` now requires writing `Optional[some_type] = None` instead of `some_type = None` which is [recommended behaviour](https://www.python.org/dev/peps/pep-0484/#id29).

Most of the code is just type changes, but where genuine bugs were revealed some (minor) logic changes were needed also.

I will post some comments on the PR shortly to ease reviewing.

I have additionally taken the change to run `isort` on the affected modules for more pleasant import ordering.

One particularly pernicious pattern I have decided to refactor is to do with initializing instance attributes as `None` and later setting them to their proper values. So instead of this (which would fail type-checking now):
```python
class Foo:
    def __init__(self):
        # to be set later
        self.attr = None # type: int
```

we now declare the type of the non-initialized attribute in the class instead:
```python
class Foo:
    attr: int
```

Now you may ask, why not just stick with the first variant and annotate with `# type: Optional[int]`? The answer is that `None` is (almost always) not a valid value as in the `self.attr = None` pattern is used as a placeholder to be initialized later on. But this can lead to a variety of bugs if we don't explicitly check for `None` every time we want to use the (hopefully initialized) attribute. Type checkers like `mypy` are also unlikely to know that when accessing `self.attr` after initialization there is no risk of it being `None` because of potentially complex runtime logic that determines `self.attr` can't be `None` which the type-checker cannot follow. Finally, we may fail to consider all cases of setting `self.attr` to it's not-`None` value and have a genuine bug where we believe `self.attr` isn't `None` anymore when it still is... Thus, it is better to declare but not instantiate to a generally invalid `None` value. In this scenario we can also check that if the value needs instantiating by calling `hasattr(self, 'attr')` instead of comparing to `None` as in the previous scenario.

A final minor curiosity, because our `numpy` version is older and doesn't include `numpy.typing` advancements, anywhere `arr: np.ndarray = None` is declared doesn't actually result in a `mypy` error. However, we're good citizens and wrap it in `Optional` as well, this will future-proof when later versionf of `numpy` can be used.

### DIFF
--- a/alibi/confidence/model_linearity.py
+++ b/alibi/confidence/model_linearity.py
@@ -1,10 +1,11 @@
 import logging
-from time import time
-from typing import Tuple, Callable, Union, List
-from numpy.linalg import norm
-import numpy as np
-from sklearn.neighbors import NearestNeighbors
 import string
+from time import time
+from typing import Callable, List, Optional, Tuple, Union
+
+import numpy as np
+from numpy.linalg import norm
+from sklearn.neighbors import NearestNeighbors
 
 logger = logging.getLogger(__name__)
 
@@ -218,7 +219,7 @@ def _sample_knn(x: np.ndarray, X_train: np.ndarray, nb_samples: int = 10) -> np.
     return np.asarray(X_sampled)  # shape=(nb_instances, nb_samples, nb_features)
 
 
-def _sample_grid(x: np.ndarray, feature_range: np.ndarray = None, epsilon: float = 0.04,
+def _sample_grid(x: np.ndarray, feature_range: np.ndarray, epsilon: float = 0.04,
                  nb_samples: int = 10, res: int = 100) -> np.ndarray:
     """Samples data points uniformly from an interval centered at x and with size epsilon * Delta,
     with delta = f_max - f_min the features ranges.
@@ -259,10 +260,17 @@ def _sample_grid(x: np.ndarray, feature_range: np.ndarray = None, epsilon: float
     return X_sampled
 
 
-def _linearity_measure(predict_fn: Callable, x: np.ndarray, X_train: np.ndarray = None,
-                       feature_range: Union[List, np.ndarray] = None, method: str = None,
-                       epsilon: float = 0.04, nb_samples: int = 10, res: int = 100,
-                       alphas: np.ndarray = None, model_type: str = 'classifier', agg: str = 'global') -> np.ndarray:
+def _linearity_measure(predict_fn: Callable,
+                       x: np.ndarray,
+                       X_train: Optional[np.ndarray] = None,
+                       feature_range: Optional[Union[List, np.ndarray]] = None,
+                       method: Optional[str] = None,
+                       epsilon: float = 0.04,
+                       nb_samples: int = 10,
+                       res: int = 100,
+                       alphas: Optional[np.ndarray] = None,
+                       model_type: str = 'classifier',
+                       agg: str = 'global') -> np.ndarray:
     """Calculate the linearity measure of the model around an instance of interest x.
 
     Parameters
@@ -341,10 +349,16 @@ def _infer_feature_range(X_train: np.ndarray) -> np.ndarray:
     return np.vstack((X_train.min(axis=0), X_train.max(axis=0))).T
 
 
-class LinearityMeasure(object):
+class LinearityMeasure:
 
-    def __init__(self, method: str = 'grid', epsilon: float = 0.04, nb_samples: int = 10, res: int = 100,
-                 alphas: np.ndarray = None, model_type: str = 'classifier', agg: str = 'pairwise',
+    def __init__(self,
+                 method: str = 'grid',
+                 epsilon: float = 0.04,
+                 nb_samples: int = 10,
+                 res: int = 100,
+                 alphas: Optional[np.ndarray] = None,
+                 model_type: str = 'classifier',
+                 agg: str = 'pairwise',
                  verbose: bool = False) -> None:
         """
 
@@ -430,9 +444,16 @@ class LinearityMeasure(object):
         return lin
 
 
-def linearity_measure(predict_fn: Callable, x: np.ndarray, feature_range: Union[List, np.ndarray] = None,
-                      method: str = 'grid', X_train: np.ndarray = None, epsilon: float = 0.04,
-                      nb_samples: int = 10, res: int = 100, alphas: np.ndarray = None, agg: str = 'global',
+def linearity_measure(predict_fn: Callable,
+                      x: np.ndarray,
+                      feature_range: Optional[Union[List, np.ndarray]] = None,
+                      method: str = 'grid',
+                      X_train: Optional[np.ndarray] = None,
+                      epsilon: float = 0.04,
+                      nb_samples: int = 10,
+                      res: int = 100,
+                      alphas: Optional[np.ndarray] = None,
+                      agg: str = 'global',
                       model_type: str = 'classifier') -> np.ndarray:
     """Calculate the linearity measure of the model around an instance of interest x.
 
--- a/alibi/datasets.py
+++ b/alibi/datasets.py
@@ -51,15 +51,16 @@ def load_cats(target_size: tuple = (299, 299), return_X_y: bool = False) -> Unio
     (data, target)
         Tuple if ``return_X_y`` is true
     """
-    tar = tarfile.open(fileobj=BytesIO(pkgutil.get_data(__name__, "data/cats.tar.gz")), mode='r:gz')  # type: ignore
+    tar = tarfile.open(fileobj=BytesIO(pkgutil.get_data(__name__, "data/cats.tar.gz")),  # type: ignore[arg-type]
+                       mode='r:gz')
     images = []
     target = []
     target_names = []
     for member in tar.getmembers():
         # data
-        img = tar.extractfile(member).read()  # type: ignore
+        img = tar.extractfile(member).read()  # type: ignore[union-attr]
         img = PIL.Image.open(BytesIO(img))
-        img = np.expand_dims(img.resize(target_size), axis=0)  # type: ignore
+        img = np.expand_dims(img.resize(target_size), axis=0)
         images.append(img)
 
         # labels
@@ -71,7 +72,7 @@ def load_cats(target_size: tuple = (299, 299), return_X_y: bool = False) -> Unio
     images = np.concatenate(images, axis=0)
     targets = np.asarray(target)
     if return_X_y:
-        return images, targets  # type: ignore
+        return images, targets  # type: ignore[return-value] # TODO: allow redefiniton
     else:
         return Bunch(data=images, target=targets, target_names=target_names)
 
@@ -115,7 +116,7 @@ def fetch_movie_sentiment(return_X_y: bool = False, url_id: int = 0) -> Union[Bu
     labels = []
     for i, member in enumerate(tar.getnames()[1:]):
         f = tar.extractfile(member)
-        for line in f.readlines():  # type: ignore
+        for line in f.readlines():  # type: ignore[union-attr]
             try:
                 line.decode('utf8')
             except UnicodeDecodeError:
--- a/alibi/datasets.py
+++ b/alibi/datasets.py
@@ -1,17 +1,18 @@
-import PIL
+import logging
+import pkgutil
+import tarfile
 from io import BytesIO, StringIO
+from typing import Optional, Tuple, Union
+
 import numpy as np
 import pandas as pd
-import pkgutil
+import PIL
 import requests
+import tensorflow.keras as keras
 from requests import RequestException
 from sklearn.preprocessing import LabelEncoder
-import tarfile
-from typing import Tuple, Union
-import logging
-from alibi.utils.data import Bunch
 
-import tensorflow.keras as keras
+from alibi.utils.data import Bunch
 
 logger = logging.getLogger(__name__)
 
@@ -50,13 +51,13 @@ def load_cats(target_size: tuple = (299, 299), return_X_y: bool = False) -> Unio
     (data, target)
         Tuple if ``return_X_y`` is true
     """
-    tar = tarfile.open(fileobj=BytesIO(pkgutil.get_data(__name__, "data/cats.tar.gz")), mode='r:gz')
+    tar = tarfile.open(fileobj=BytesIO(pkgutil.get_data(__name__, "data/cats.tar.gz")), mode='r:gz')  # type: ignore
     images = []
     target = []
     target_names = []
     for member in tar.getmembers():
         # data
-        img = tar.extractfile(member).read()
+        img = tar.extractfile(member).read()  # type: ignore
         img = PIL.Image.open(BytesIO(img))
         img = np.expand_dims(img.resize(target_size), axis=0)  # type: ignore
         images.append(img)
@@ -114,7 +115,7 @@ def fetch_movie_sentiment(return_X_y: bool = False, url_id: int = 0) -> Union[Bu
     labels = []
     for i, member in enumerate(tar.getnames()[1:]):
         f = tar.extractfile(member)
-        for line in f.readlines():
+        for line in f.readlines():  # type: ignore
             try:
                 line.decode('utf8')
             except UnicodeDecodeError:
@@ -129,7 +130,7 @@ def fetch_movie_sentiment(return_X_y: bool = False, url_id: int = 0) -> Union[Bu
     return Bunch(data=data, target=labels, target_names=target_names)
 
 
-def fetch_adult(features_drop: list = None, return_X_y: bool = False, url_id: int = 0) -> \
+def fetch_adult(features_drop: Optional[list] = None, return_X_y: bool = False, url_id: int = 0) -> \
         Union[Bunch, Tuple[np.ndarray, np.ndarray]]:
     """
     Downloads and pre-processes 'adult' dataset.
--- a/alibi/explainers/ale.py
+++ b/alibi/explainers/ale.py
@@ -89,7 +89,7 @@ class ALE(Explainer):
                                    extrapolate_constant_perc=extrapolate_constant_perc,
                                    extrapolate_constant_min=extrapolate_constant_min)
 
-    def explain(self, X: np.ndarray, features: List[int] = None, min_bin_points: int = 4) -> Explanation:
+    def explain(self, X: np.ndarray, features: Optional[List[int]] = None, min_bin_points: int = 4) -> Explanation:
         """
         Calculate the ALE curves for each feature with respect to the dataset `X`.
 
@@ -130,7 +130,7 @@ class ALE(Explainer):
             feature_names = self.feature_names[features]  # type: ignore
         else:
             feature_names = self.feature_names
-            features = range(n_features)  # type: ignore
+            features = list(range(n_features))
 
         feature_values = []
         ale_values = []
--- a/alibi/explainers/anchor_base.py
+++ b/alibi/explainers/anchor_base.py
@@ -1,11 +1,12 @@
 import copy
 import logging
-import numpy as np
 from collections import defaultdict, namedtuple
 from functools import partial
-from typing import Callable, Tuple, Set, Dict, List
+from typing import Callable, Dict, List, Optional, Set, Tuple
 
-from alibi.utils.distributed import ActorPool, RAY_INSTALLED
+import numpy as np
+
+from alibi.utils.distributed import RAY_INSTALLED, ActorPool
 from alibi.utils.distributions import kl_bernoulli
 
 logger = logging.getLogger(__name__)
@@ -24,7 +25,7 @@ class AnchorBaseBeam:
         """
 
         self.sample_fcn = samplers[0]
-        self.samplers = None  # type: List[Callable]
+        self.samplers = None  # type: Optional[List[Callable]]
         # Initial size (in batches) of data/raw data samples cache.
         self.sample_cache_size = kwargs.get('sample_cache_size', 10000)
         # when only the max of self.margin or batch size remain emptpy, the cache is
@@ -175,7 +176,7 @@ class AnchorBaseBeam:
 
         return temp + np.log(temp)
 
-    def _get_coverage_samples(self, coverage_samples: int, samplers: List[Callable] = None) -> np.ndarray:
+    def _get_coverage_samples(self, coverage_samples: int, samplers: Optional[List[Callable]] = None) -> np.ndarray:
         """
         Draws samples uniformly at random from the training set.
 
@@ -610,7 +611,7 @@ class AnchorBaseBeam:
 
     def anchor_beam(self, delta: float = 0.05, epsilon: float = 0.1, desired_confidence: float = 1.,
                     beam_size: int = 1, epsilon_stop: float = 0.05, min_samples_start: int = 100,
-                    max_anchor_size: int = None, stop_on_first: bool = False, batch_size: int = 100,
+                    max_anchor_size: Optional[int] = None, stop_on_first: bool = False, batch_size: int = 100,
                     coverage_samples: int = 10000, verbose: bool = False, verbose_every: int = 1,
                     **kwargs) -> dict:
 
@@ -835,7 +836,7 @@ class DistributedAnchorBaseBeam(AnchorBaseBeam):
         self.pool = ActorPool(samplers)
         self.samplers = samplers
 
-    def _get_coverage_samples(self, coverage_samples: int, samplers: List[Callable] = None) -> np.ndarray:
+    def _get_coverage_samples(self, coverage_samples: int, samplers: List[Callable]) -> np.ndarray: # type: ignore
         """
         Sends a request for a coverage set to process running sampling tasks.
 
--- a/alibi/explainers/anchor_base.py
+++ b/alibi/explainers/anchor_base.py
@@ -199,7 +199,7 @@ class AnchorBaseBeam:
         return coverage_data
 
     def select_critical_arms(self, means: np.ndarray, ub: np.ndarray, lb: np.ndarray, n_samples: np.ndarray,
-                             delta: float, top_n: int, t: int):  # type: ignore
+                             delta: float, top_n: int, t: int):
         """
         Determines a set of two anchors by updating the upper bound for low emprical precision anchors and
         the lower bound for anchors with high empirical precision.
@@ -475,7 +475,7 @@ class AnchorBaseBeam:
                 (self.state['labels'], np.zeros(prealloc_size, labels.dtype))
             )
 
-        return labels.sum(), data.shape[0]  # type: ignore
+        return labels.sum(), data.shape[0]
 
     def get_init_stats(self, anchors: list, coverages=False) -> dict:
         """
@@ -836,7 +836,8 @@ class DistributedAnchorBaseBeam(AnchorBaseBeam):
         self.pool = ActorPool(samplers)
         self.samplers = samplers
 
-    def _get_coverage_samples(self, coverage_samples: int, samplers: List[Callable]) -> np.ndarray:  # type: ignore
+    def _get_coverage_samples(self, coverage_samples: int,  # type: ignore[override]
+                              samplers: List[Callable]) -> np.ndarray:
         """
         Sends a request for a coverage set to process running sampling tasks.
 
@@ -855,7 +856,7 @@ class DistributedAnchorBaseBeam(AnchorBaseBeam):
 
         return coverage_data
 
-    def draw_samples(self, anchors: list, batch_size: int) -> Tuple[np.ndarray, np.ndarray]:  # type: ignore
+    def draw_samples(self, anchors: list, batch_size: int) -> Tuple[np.ndarray, np.ndarray]:  # type: ignore[override]
         """
         Distributes sampling requests among processes running sampling tasks.
 
--- a/alibi/explainers/anchor_base.py
+++ b/alibi/explainers/anchor_base.py
@@ -836,7 +836,7 @@ class DistributedAnchorBaseBeam(AnchorBaseBeam):
         self.pool = ActorPool(samplers)
         self.samplers = samplers
 
-    def _get_coverage_samples(self, coverage_samples: int, samplers: List[Callable]) -> np.ndarray: # type: ignore
+    def _get_coverage_samples(self, coverage_samples: int, samplers: List[Callable]) -> np.ndarray:  # type: ignore
         """
         Sends a request for a coverage set to process running sampling tasks.
 
--- a/alibi/explainers/anchor_image.py
+++ b/alibi/explainers/anchor_image.py
@@ -1,18 +1,19 @@
 import copy
 import logging
+from functools import partial
+from typing import Any, Callable, Dict, List, Optional, Tuple, Type, Union
 
 import numpy as np
+from skimage.segmentation import felzenszwalb, quickshift, slic
 
-from functools import partial
-from typing import Any, Callable, List, Union, Tuple, Type
-
-from alibi.utils.wrappers import ArgmaxTransformer
+from alibi.api.defaults import DEFAULT_DATA_ANCHOR_IMG, DEFAULT_META_ANCHOR
 from alibi.api.interfaces import Explainer, Explanation
-from alibi.api.defaults import DEFAULT_META_ANCHOR, DEFAULT_DATA_ANCHOR_IMG
-from alibi.exceptions import AlibiPredictorCallException, AlibiPredictorReturnTypeError
+from alibi.exceptions import (AlibiPredictorCallException,
+                              AlibiPredictorReturnTypeError)
+from alibi.utils.wrappers import ArgmaxTransformer
+
 from .anchor_base import AnchorBaseBeam
 from .anchor_explanation import AnchorExplanation
-from skimage.segmentation import felzenszwalb, slic, quickshift
 
 logger = logging.getLogger(__name__)
 
@@ -20,7 +21,7 @@ DEFAULT_SEGMENTATION_KWARGS = {
     'felzenszwalb': {},
     'quickshift': {},
     'slic': {'n_segments': 10, 'compactness': 10, 'sigma': .5}
-}
+}  # type: Dict[str, Dict]
 
 
 def scale_image(image: np.ndarray, scale: tuple = (0, 255)) -> np.ndarray:
@@ -54,7 +55,7 @@ class AnchorImageSampler:
             segmentation_fn: Callable,
             custom_segmentation: bool,
             image: np.ndarray,
-            images_background: np.ndarray = None,
+            images_background: Optional[np.ndarray] = None,
             p_sample: float = 0.5,
             n_covered_ex: int = 10,
     ):
@@ -304,9 +305,9 @@ class AnchorImage(Explainer):
                  image_shape: tuple,
                  dtype: Type[np.generic] = np.float32,
                  segmentation_fn: Any = 'slic',
-                 segmentation_kwargs: dict = None,
-                 images_background: np.ndarray = None,
-                 seed: int = None) -> None:
+                 segmentation_kwargs: Optional[dict] = None,
+                 images_background: Optional[np.ndarray] = None,
+                 seed: Optional[int] = None) -> None:
         """
         Initialize anchor image explainer.
 
@@ -342,9 +343,11 @@ class AnchorImage(Explainer):
         super().__init__(meta=copy.deepcopy(DEFAULT_META_ANCHOR))
         np.random.seed(seed)
 
-        if isinstance(segmentation_fn, str) and not segmentation_kwargs:
+        # TODO: this logic needs improvement. We should check against a fixed set of strings
+        # for built-ins instead of any `str`.
+        if isinstance(segmentation_fn, str) and segmentation_kwargs is None:
             try:
-                segmentation_kwargs = DEFAULT_SEGMENTATION_KWARGS[segmentation_fn]  # type: ignore
+                segmentation_kwargs = DEFAULT_SEGMENTATION_KWARGS[segmentation_fn]
             except KeyError:
                 logger.warning(
                     'DEFAULT_SEGMENTATION_KWARGS did not contain any entry'
@@ -355,9 +358,11 @@ class AnchorImage(Explainer):
         elif callable(segmentation_fn) and segmentation_kwargs:
             logger.warning(
                 'Specified both a segmentation function to create superpixels and '
-                'keyword arguments for built segmentation functions. By default '
+                'keyword arguments for built-in segmentation functions. By default '
                 'the specified segmentation function will be used.'
             )
+        else:
+            segmentation_kwargs = {}
 
         # set the predictor
         self.image_shape = tuple(image_shape)  # coerce lists
@@ -441,7 +446,7 @@ class AnchorImage(Explainer):
                 coverage_samples: int = 10000,
                 beam_size: int = 1,
                 stop_on_first: bool = False,
-                max_anchor_size: int = None,
+                max_anchor_size: Optional[int] = None,
                 min_samples_start: int = 100,
                 n_covered_ex: int = 10,
                 binary_cache_size: int = 10000,
--- a/alibi/explainers/anchor_image.py
+++ b/alibi/explainers/anchor_image.py
@@ -250,7 +250,7 @@ class AnchorImageSampler:
                 mask[segments == superpixel] = True
             if background_idx is not None:
                 # replace values with those of background image
-                temp[mask] = self.images_background[background_idx][mask]
+                temp[mask] = self.images_background[background_idx][mask]  # type: ignore[index]
             else:
                 # ... or with the averaged superpixel value
                 temp[mask] = fudged_image[mask]
@@ -436,7 +436,7 @@ class AnchorImage(Explainer):
 
         return image_preproc
 
-    def explain(self,  # type: ignore
+    def explain(self,  # type: ignore[override]
                 image: np.ndarray,
                 p_sample: float = 0.5,
                 threshold: float = 0.95,
--- a/alibi/explainers/anchor_tabular.py
+++ b/alibi/explainers/anchor_tabular.py
@@ -292,7 +292,8 @@ class TabularSampler:
         if num_samples_pos == 0:
             samples_idxs = np.random.choice(partial_anchor_rows[-1], num_samples)
             samples[:, uniq_feat_ids] = self.train_data[np.ix_(samples_idxs, uniq_feat_ids)]  # type: ignore[arg-type]
-            d_samples[:, uniq_feat_ids] = self.d_train_data[np.ix_(samples_idxs, uniq_feat_ids)]  # type: ignore[arg-type]
+            d_samples[:, uniq_feat_ids] = self.d_train_data[
+                np.ix_(samples_idxs, uniq_feat_ids)]  # type: ignore[arg-type]
 
             return samples, d_samples, coverage
 
@@ -394,7 +395,8 @@ class TabularSampler:
                         replace=True,
                     )
                 n_samp = num_samples
-            samples[start:start + n_samp, uniq_feat_ids[idx:]] = self.train_data[np.ix_(samp_idxs, uniq_feat_ids[idx:])]  # type: ignore[arg-type]
+            samples[start:start + n_samp, uniq_feat_ids[idx:]] = self.train_data[
+                np.ix_(samp_idxs, uniq_feat_ids[idx:])]  # type: ignore[arg-type]
 
             # deal with partial anchors; idx = 0 means that we actually sample the entire anchor
             if idx > 0:
--- a/alibi/explainers/anchor_tabular.py
+++ b/alibi/explainers/anchor_tabular.py
@@ -1,28 +1,34 @@
 import copy
 import logging
-import numpy as np
 from collections import OrderedDict, defaultdict
 from itertools import accumulate
-from typing import Any, Callable, DefaultDict, Dict, List, Set, Tuple, Type, Union, Optional
+from typing import (Any, Callable, DefaultDict, Dict, List, Optional, Set,
+                    Tuple, Type, Union)
 
+import numpy as np
+
+from alibi.api.defaults import DEFAULT_DATA_ANCHOR, DEFAULT_META_ANCHOR
 from alibi.api.interfaces import Explainer, Explanation, FitMixin
-from alibi.api.defaults import DEFAULT_META_ANCHOR, DEFAULT_DATA_ANCHOR
-from alibi.exceptions import AlibiPredictorCallException, AlibiPredictorReturnTypeError
-from .anchor_base import AnchorBaseBeam, DistributedAnchorBaseBeam
-from .anchor_explanation import AnchorExplanation
-from alibi.utils.wrappers import ArgmaxTransformer
+from alibi.exceptions import (AlibiPredictorCallException,
+                              AlibiPredictorReturnTypeError)
 from alibi.utils.discretizer import Discretizer
 from alibi.utils.distributed import RAY_INSTALLED
 from alibi.utils.mapping import ohe_to_ord, ord_to_ohe
+from alibi.utils.wrappers import ArgmaxTransformer
+
+from .anchor_base import AnchorBaseBeam, DistributedAnchorBaseBeam
+from .anchor_explanation import AnchorExplanation
 
 
 class TabularSampler:
     """ A sampler that uses an underlying training set to draw records that have a subset of features with
-    values specified in an instance to be expalined, X."""
+    values specified in an instance to be explained, X."""
+
+    instance_label: int
 
     def __init__(self, predictor: Callable, disc_perc: Tuple[Union[int, float], ...], numerical_features: List[int],
                  categorical_features: List[int], feature_names: list, feature_values: dict, n_covered_ex: int = 10,
-                 seed: int = None) -> None:
+                 seed: Optional[int] = None) -> None:
         """
         Parameters
         ----------
@@ -47,7 +53,6 @@ class TabularSampler:
 
         np.random.seed(seed)
 
-        self.instance_label = None  # type: int
         self.predictor = predictor
         self.n_covered_ex = n_covered_ex
 
@@ -128,7 +133,7 @@ class TabularSampler:
              Instance to be explained.
         """
 
-        label = self.predictor(X.reshape(1, -1))[0]
+        label = self.predictor(X.reshape(1, -1))[0]  # type: int
         self.instance_label = label
 
     def set_n_covered(self, n_covered: int) -> None:
@@ -453,7 +458,8 @@ class TabularSampler:
         allowed_bins = {}  # type: Dict[int, Set[int]]
         # index of database rows (values) for each feature in result (key: feat id)
         allowed_rows = {}  # type: Dict[int, Any[int]]
-        unk_feat_values = []  # feats for which there are not training records in the desired bin/with that value
+        # feats for which there are not training records in the desired bin/with that value
+        unk_feat_values = []  # type: List[Tuple[int, str, Optional[int]]]
         cat_enc_ids = [enc_id for enc_id in anchor if enc_id in self.cat_lookup.keys()]
         ord_enc_ids = [enc_id for enc_id in anchor if enc_id in self.ord_lookup.keys()]
         if cat_enc_ids:
@@ -659,14 +665,15 @@ class RemoteSampler:
 
 
 class AnchorTabular(Explainer, FitMixin):
+    instance_label: int
 
     def __init__(self,
                  predictor: Callable[[np.ndarray], np.ndarray],
                  feature_names: List[str],
-                 categorical_names: Dict[int, List[str]] = None,
+                 categorical_names: Optional[Dict[int, List[str]]] = None,
                  dtype: Type[np.generic] = np.float32,
                  ohe: bool = False,
-                 seed: int = None) -> None:
+                 seed: Optional[int] = None) -> None:
         """
         Parameters
         ----------
@@ -699,7 +706,7 @@ class AnchorTabular(Explainer, FitMixin):
         self.ohe = ohe
         self.feature_names = feature_names
 
-        if ohe:
+        if ohe and categorical_names:
             self.cat_vars_ord = {col: len(values) for col, values in categorical_names.items()}
             self.cat_vars_ohe = ord_to_ohe(np.zeros((1, len(feature_names))), self.cat_vars_ord)[1]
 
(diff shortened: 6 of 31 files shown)


alibi/confidence/model_linearity.py
```diff
@@ -218,7 +219,7 @@ def _sample_knn(x: np.ndarray, X_train: np.ndarray, nb_samples: int = 10) -> np.
     return np.asarray(X_sampled)  # shape=(nb_instances, nb_samples, nb_features)
 
 
-def _sample_grid(x: np.ndarray, feature_range: np.ndarray = None, epsilon: float = 0.04,
+def _sample_grid(x: np.ndarray, feature_range: np.ndarray, epsilon: float = 0.04,
```
jklaise, line 222: There is no `None` handling in the function so `feature_range` must be required.

alibi/confidence/model_linearity.py
```diff
@@ -341,10 +349,16 @@ def _infer_feature_range(X_train: np.ndarray) -> np.ndarray:
     return np.vstack((X_train.min(axis=0), X_train.max(axis=0))).T
 
 
-class LinearityMeasure(object):
+class LinearityMeasure:
```
jklaise, line 352: Removed the old-style ineheritance from `object` as it's not necessary in Python 3.

alibi/datasets.py
```diff
@@ -50,13 +51,13 @@ def load_cats(target_size: tuple = (299, 299), return_X_y: bool = False) -> Unio
     (data, target)
         Tuple if ``return_X_y`` is true
     """
-    tar = tarfile.open(fileobj=BytesIO(pkgutil.get_data(__name__, "data/cats.tar.gz")), mode='r:gz')
+    tar = tarfile.open(fileobj=BytesIO(pkgutil.get_data(__name__, "data/cats.tar.gz")), mode='r:gz')  # type: ignore
```
jklaise, line 54: The three `type: ignore` in this module are because the various objects actually return `Optional`, presumably because the data is not guaranteed to exist or yield anything. I don't think we need to check for this explicitly as we are processing data that we ship with the library.

alibi/explainers/ale.py
```diff
@@ -130,7 +130,7 @@ def explain(self, X: np.ndarray, features: List[int] = None, min_bin_points: int
             feature_names = self.feature_names[features]  # type: ignore
         else:
             feature_names = self.feature_names
-            features = range(n_features)  # type: ignore
+            features = list(range(n_features))
```
jklaise, line 133: Expect a list, therefore a list you shall have.

alibi/explainers/anchor_base.py
```diff
@@ -24,7 +25,7 @@ def __init__(self, samplers: List[Callable], **kwargs) -> None:
         """
 
         self.sample_fcn = samplers[0]
-        self.samplers = None  # type: List[Callable]
+        self.samplers = None  # type: Optional[List[Callable]]
```
jklaise, line 28: This is an unfortuante exception in the tirade of the original comment because the class `AnchorBaseBeam` is inherited by `DistributedAnchorBaseBeam` which uses a `self.samplers` attribute even though it is never set for this base class. Sorting this out would require more refactoring than necessary for this PR.

alibi/explainers/anchor_base.py
```diff
@@ -835,7 +836,7 @@ def __init__(self, samplers: List[Callable], **kwargs) -> None:
         self.pool = ActorPool(samplers)
         self.samplers = samplers
 
-    def _get_coverage_samples(self, coverage_samples: int, samplers: List[Callable] = None) -> np.ndarray:
+    def _get_coverage_samples(self, coverage_samples: int, samplers: List[Callable]) -> np.ndarray: # type: ignore
```
jklaise, line 839: We ignore the type here because this method is an override of the one in the base class, except here `samplers` really is required.

alibi/explainers/anchor_image.py
```diff
@@ -344,9 +345,11 @@ def __init__(self,
         super().__init__(meta=copy.deepcopy(DEFAULT_META_ANCHOR))
         np.random.seed(seed)
 
-        if isinstance(segmentation_fn, str) and not segmentation_kwargs:
+        # TODO: this logic needs improvement. We should check against a fixed set of strings
+        # for built-ins instead of any `str`.
+        if isinstance(segmentation_fn, str) and segmentation_kwargs is None:
```
jklaise, line 350: `and not` was a bug as it would also match the empty dictionary. Now we explicitly deal with the `None` case which conveniently doesn't require the `type: ignore` later on.

alibi/explainers/anchor_tabular.py
```diff
@@ -128,7 +133,7 @@ def set_instance_label(self, X: np.ndarray) -> None:
              Instance to be explained.
         """
 
-        label = self.predictor(X.reshape(1, -1))[0]
+        label = self.predictor(X.reshape(1, -1))[0]  # type: int
```
jklaise, line 136: We need to declare the type here otherwise we will get warnings later on when passing `label` to functions.

alibi/explainers/anchor_tabular.py
```diff
@@ -698,7 +705,7 @@ def __init__(self,
         self.ohe = ohe
         self.feature_names = feature_names
 
-        if ohe:
+        if ohe and categorical_names:
```
jklaise, line 708: We never validate the combinations that are consistent, this check is the easiest one to add for now.

alibi/explainers/anchor_tabular.py
```diff
@@ -1003,19 +1009,19 @@ def predictor(self) -> Callable:
         return self._ohe_predictor if self.ohe else self._predictor
 
     @predictor.setter
-    def predictor(self, predictor: Optional[Callable]) -> None:
+    def predictor(self, predictor: Callable) -> None:
```
jklaise, line 1012: `predictor` should never be `None`.

alibi/explainers/backends/cfrl_tabular.py
```diff
@@ -285,7 +285,7 @@ def generate_condition(X_ohe: np.ndarray,
 
 def sample_numerical(X_hat_num_split: List[np.ndarray],
                      X_ohe_num_split: List[np.ndarray],
-                     C_num_split: List[np.ndarray],
+                     C_num_split: Optional[List[np.ndarray]],
```
jklaise, line 288: `None` case is dealt with in the body so it can clearly be `Optional` (as inferred by the type-checker elsewhere in the code).

alibi/explainers/backends/cfrl_tabular.py
```diff
@@ -335,7 +335,7 @@ def sample_numerical(X_hat_num_split: List[np.ndarray],
 
 
 def sample_categorical(X_hat_cat_split: List[np.ndarray],
-                       C_cat_split: List[np.ndarray]) -> List[np.ndarray]:
+                       C_cat_split: Optional[List[np.ndarray]]) -> List[np.ndarray]:
```
jklaise, line 338: `None` case is dealt with in the body so it can clearly be `Optional` (as inferred by the type-checker elsewhere in the code).

alibi/explainers/backends/cfrl_tabular.py
```diff
@@ -506,7 +506,7 @@ def get_inv_preprocessor(X_ohe: np.ndarray):
 
         np_X_inv = np_X_inv[:, inv_perm].astype(object)
         for i, fn in enumerate(feature_names):
-            type = feature_types[fn] if fn in feature_types else float
+            type = feature_types[fn] if fn in feature_types else float  # type: ignore # TODO: closure resets type info?
```
jklaise, line 509: See here https://mypy.readthedocs.io/en/stable/common_issues.html#narrowing-and-inner-functions

alibi/explainers/cfproto.py
```diff
@@ -182,16 +185,15 @@ def __init__(self,
         self.clip = clip
         self.write_dir = write_dir
 
-        # compute dimensionality after conversion from OHE to ordinal encoding
-        shape = ohe_to_ord_shape(shape, cat_vars=cat_vars, is_ohe=self.ohe)
-
         if self.is_cat:
+            # compute dimensionality after conversion from OHE to ordinal encoding
+            shape = ohe_to_ord_shape(shape, cat_vars=cat_vars, is_ohe=self.ohe)
```
jklaise, line 190: I moved this inside the `if self.is_cat` block because it really doesn't make sense to do if there is no `cat_vars` (again, a case of not validating allowed combinations of kwargs).

alibi/explainers/cfproto.py
```diff
@@ -182,16 +185,15 @@ def __init__(self,
...
 
             # define ragged tensor for mapping from categorical to numerical values
-            self.map_cat_to_num = tf.ragged.constant([np.zeros(v) for _, v in self.cat_vars.items()])
+            self.map_cat_to_num = tf.ragged.constant([np.zeros(v) for _, v in cat_vars.items()])
 
             # define placeholder for mapping which can be fed after the fit step
-            max_key = max(cat_vars, key=cat_vars.get)
+            max_key = max(cat_vars)
```
jklaise, line 196: `max` is by default over the keys. This also avoids a type error because `dict.get` actually returns an `Optional` type.

alibi/explainers/cfproto.py
```diff
@@ -701,7 +710,7 @@ def fit(self, train_data: np.ndarray, trustscore_kwargs: dict = None, d_type: st
         else:
             preds = np.argmax(self.predict(train_data), axis=1)
 
-        self.cat_vars_ord = None
+        self.cat_vars_ord = dict()  # type: dict
```
jklaise, line 713: Avoids checking for `None` later on when passing to functions that expect a `dict`.

alibi/explainers/cfproto.py
```diff
@@ -734,6 +743,10 @@ def fit(self, train_data: np.ndarray, trustscore_kwargs: dict = None, d_type: st
 
             # combined distance measure
             if d_type == 'abdm-mvdm':
+                if w is None:
+                    msg = f"Must specify a value for `w` if using d_type='abdm-mvdm'"
+                    raise ValueError(msg)
```
jklaise, line 748: This validation was missing which the type-checker caught as trying to do arithmetic with `w=None` and integers later in the function.

alibi/explainers/cfproto.py
```diff
@@ -826,7 +839,7 @@ def loss_fn(self, pred_proba: np.ndarray, Y: np.ndarray) -> np.ndarray:
         return loss_attack
 
     def get_gradients(self, X: np.ndarray, Y: np.ndarray, grads_shape: tuple,
-                      cat_vars_ord: dict = None) -> np.ndarray:
+                      cat_vars_ord: dict) -> np.ndarray:
```
jklaise, line 842: Now `cat_vars_ord` is always a `dict` even if it's the empty one.

alibi/explainers/cfproto.py
```diff
@@ -1045,15 +1058,16 @@ def compare(x: Union[float, int, np.ndarray], y: int) -> bool:
                 self.class_proto[c] = self.X_by_class[c][idx_c[0][-1]].reshape(1, -1)
 
         if self.enc_or_kdtree:
-            self.id_proto = min(dist_proto, key=dist_proto.get)
+            self.id_proto = min(dist_proto)
```
jklaise, line 1061: Similar comment to `max` above.

alibi/explainers/cfproto.py
```diff
@@ -1045,15 +1058,16 @@ def compare(x: Union[float, int, np.ndarray], y: int) -> bool:
...
             if verbose:
                 print('Prototype class: {}'.format(self.id_proto))
         else:  # no prototype loss term used
             proto_val = np.zeros(self.shape_enc)
 
-        # set shape for perturbed instance and gradients
-        pert_shape = ohe_to_ord_shape(self.shape, cat_vars=self.cat_vars, is_ohe=self.ohe)
+        if self.is_cat:
+            # set shape for perturbed instance and gradients
+            pert_shape = ohe_to_ord_shape(self.shape, cat_vars=self.cat_vars, is_ohe=self.ohe)
```
jklaise, line 1070: Only makes sense to do if `cat_vars` was passed.

alibi/explainers/cfrl_base.py
```diff
@@ -177,13 +180,13 @@ def sample(self) -> Dict[str, Optional[np.ndarray]]:
...
-        Z_cf_tilde = self.Z_cf_tilde[rand_idx]                      # noised counterfactual embedding
-        C = self.C[rand_idx] if (self.C is not None) else None      # conditional array if exists
-        R_tilde = self.R_tilde[rand_idx]                            # noised counterfactual reward
+        X = self.X[rand_idx]  # input array
+        Y_m = self.Y_m[rand_idx]  # model's prediction
+        Y_t = self.Y_t[rand_idx]  # counterfactual target
+        Z = self.Z[rand_idx]  # input embedding
+        Z_cf_tilde = self.Z_cf_tilde[rand_idx]  # noised counterfactual embedding
+        C = self.C[rand_idx] if (self.C is not None) else None  # conditional array if exists
+        R_tilde = self.R_tilde[rand_idx]  # noised counterfactual reward
```
jklaise, line 189: I can re-align the comments if desired, my IDE collapses everything when I format...

alibi/explainers/cfrl_base.py
```diff
@@ -491,7 +494,7 @@ def _validate_kwargs(self,
                          predictor: Callable,
                          encoder: 'Union[tensorflow.keras.Model, torch.nn.Module]',
                          decoder: 'Union[tensorflow.keras.Model, torch.nn.Module]',
-                         latent_dim: float,
+                         latent_dim: Optional[int],
```
jklaise, line 497: Bug in the type, it should be an `int` but may be left unspecified.

alibi/explainers/cfrl_base.py
```diff
@@ -780,7 +783,7 @@ def _is_classification(pred: np.ndarray) -> bool:
 
     def explain(self,
                 X: np.ndarray,
-                Y_t: np.ndarray = None,   # TODO: remove default value (mypy error. explanation in the validation step)
+                Y_t: np.ndarray = None,  # TODO: remove default value (mypy error. explanation in the validation step)
```
jklaise, line 786: Note to self: missing `Optional`.

alibi/explainers/cfrl_tabular.py
```diff
@@ -462,7 +465,7 @@ def _diversity(self,
             X_cf, Y_m_cf, Y_t = results["X_cf"], results["Y_m_cf"], results["Y_t"]
 
             # Select only counterfactuals where prediction matches the target.
-            X_cf = X_cf[Y_t == Y_m_cf]
+            X_cf = X_cf[Y_t == Y_m_cf]  # type: ignore # TODO: fix me
```
jklaise, line 468: `alibi/explainers/cfrl_tabular.py:468: error: Value of type "Optional[Any]" is not indexable` - felt like would need more refactoring

alibi/explainers/integrated_gradients.py
```diff
@@ -833,7 +838,7 @@ def explain(self,
                                                                    baselines,
                                                                    self.model,
                                                                    self.layer,
-                                                                   self.orig_call,
+                                                                   self.orig_call,  # type: ignore # TODO: fix me
```
jklaise, line 841: From the type-checker point of view `self.orig_call` can still bet `None` even though we narrowed it down via `else` clause (following `if self.layer is None`).

alibi/explainers/shap_wrappers.py
```diff
@@ -730,20 +732,20 @@ def fit(self,  # type: ignore
...
+        if self.create_group_names and groups:
             group_names = ['group_{}'.format(i) for i in range(len(groups))]
         # disable grouping or data weights if inputs are not correct
         if self.ignore_weights:
             weights = None
         if not self.use_groups:
             group_names, groups = None, None
         else:
-            self.feature_names = group_names
+            self.feature_names = group_names  # type:ignore # TODO: fix me
```
jklaise, line 743: `group_names` could be `None` as hinted by the method.

alibi/explainers/shap_wrappers.py
```diff
@@ -730,20 +732,20 @@ def fit(self,  # type: ignore
 
         # check user inputs to provide warnings if input is incorrect
         self._check_inputs(background_data, group_names, groups, weights)
-        if self.create_group_names:
+        if self.create_group_names and groups:
```
jklaise, line 735: We don't check if `groups` is `None` beforehand, so type-checker complains that `len(groups)` may fail.

alibi/explainers/shap_wrappers.py
```diff
@@ -730,20 +732,20 @@ def fit(self,  # type: ignore
...
             group_names, groups = None, None
         else:
-            self.feature_names = group_names
+            self.feature_names = group_names  # type:ignore # TODO: fix me
 
         # perform grouping if requested by the user
         self.background_data = self._get_data(background_data, group_names, groups, weights, **kwargs)
         explainer_args = (self.predictor, self.background_data)
-        explainer_kwargs = {'link': self.link}  # type: Dict[str, Union[str, int]]
+        explainer_kwargs = {'link': self.link}  # type: Dict[str, Union[str, int, None]]
```
jklaise, line 748: Allow `None` because `seed` can be `None`.

alibi/models/pytorch/model.py
```diff
@@ -49,7 +49,7 @@ def compile(self,
 
         if isinstance(loss, list):
             # check if the number of weights is the same as the number of partial losses
-            if len(loss_weights) != len(loss):
+            if len(self.loss_weights) != len(loss):
```
jklaise, line 52: `self.loss_weights` is already validated and can't be `None`.

alibi/saving.py
```diff
@@ -165,7 +165,7 @@ def _save_AnchorImage(explainer: 'AnchorImage', path: Union[str, os.PathLike]) -
         dill.dump(segmentation_fn, f, recurse=True)
 
     predictor = explainer.predictor
-    explainer.predictor = None
+    explainer.predictor = None  # type: ignore
```
jklaise, line 168: Ignores in this file are purely because we temporarily set some attributes to `None` so that they are not saved to disk. The attributes are reinstated to their original values afterwards.

ascillitoe:
@jklaise, thanks for the detailed comments, very helpful (I've learnt quite a few new things!).

alibi/utils/distance.py
```diff
@@ -161,10 +162,10 @@ def abdm(X: np.ndarray,
 
 
 def multidim_scaling(d_pair: dict,
+                     feature_range: tuple,
```
jklaise, line 165: `feature_range` is never checked if it can be `None` and all uses of it have non-trivial `feature_range`, therefore it should be a required argument. This also required refactoring to bring it up a few positions in the function signature.

alibi/utils/distributed.py
```diff
@@ -489,10 +490,9 @@ class DistributedOpts(TypedDict):
                 "No algorithm specified in distributed option, default target function will be selected."
             )
         self.target_fcn = default_target_fcn
-        self.concatenate = None
         # check global scope for any specific target function
         if concatenate_results:
-            self.concatenate = concatenate_minibatches
+            self.concatenate = concatenate_minibatches  # type: ignore
```
jklaise, line 495: See this https://github.com/python/mypy/issues/708

jklaise: Or this https://github.com/python/mypy/issues/2427

alibi/utils/lang_model.py
```diff
@@ -379,3 +390,7 @@ def mask(self) -> str:
 
     def is_subword_prefix(self, token: str) -> bool:
         return not token.startswith(RobertaBase.SUBWORD_PREFIX)
+
+
+if __name__ == '__main__':
+    model = DistilbertBaseUncased()
```
jklaise, line 396: Note to self - testing code, to be removed.

alibi/utils/mapping.py
```diff
@@ -1,8 +1,9 @@
+from typing import Dict, List, Tuple
+
 import numpy as np
-from typing import Tuple, List
 
 
-def ohe_to_ord_shape(shape: tuple, cat_vars: dict = None, is_ohe: bool = False) -> tuple:
+def ohe_to_ord_shape(shape: tuple, cat_vars: Dict[int, int], is_ohe: bool = False) -> tuple:
```
jklaise, line 6: `None` case is not dealt with in the body so must be required.

alibi/utils/tests/test_distributed.py
```diff
@@ -252,12 +250,13 @@ def test_invert_permutation(permutation_generator):
...
 # MockExplainer args and kwargs
-explainer_init_args = [[(0.01,), (0.05, )], ]   # type: ignore
+explainer_init_args = [[(0.01,), (0.05,)], ]  # type: ignore
 explainer_init_kwargs = [[{'multiplier': 2}, {'multiplier': 3}], [{'multiplier': 2}, ]]  # type: ignore
 
 batch_size = [2]
 ncpus = [1, len(explainer_init_kwargs[0])]  # set so that both the error raising inputs and correct inputs are tested
-actor_cpu_fraction = [1.0, 0.5, None]
+# TODO: check that redefining these (type: ignore) actually makes the tests run with the right parameters
+actor_cpu_fraction = [1.0, 0.5, None]  # type: ignore
```
jklaise, line 259: All `type: ignore` in this module are due to vairable re-definiton. This brings up a point, how is `pytest` running the correct test cases...?

jklaise:
The newer `spacy-lookups-data` version is throwing up the extra errors in `anchor_text.py`.

alibi/explainers/anchor_text.py
```diff
@@ -54,7 +56,7 @@ def _load_spacy_lexeme_prob(nlp: 'spacy.language.Language'):
         # https://github.com/explosion/spaCy/discussions/6388#discussioncomment-331096
         if 'lexeme_prob' not in nlp.vocab.lookups.tables:
             from spacy.lookups import load_lookups
-            lookups = load_lookups(nlp.lang, ['lexeme_prob'])
+            lookups = load_lookups(nlp.lang, ['lexeme_prob'])  # type: ignore
```
jklaise, line 59: Apparently type of `nlp.lang` is `Optional[str]` and `load_lookups` expects `str`.

alibi/explainers/anchor_text.py
```diff
@@ -283,7 +286,10 @@ def __init__(self, nlp: 'spacy.language.Language', perturb_opts: Dict):
 
         # dict containing an np.array of similar words with same part of speech and an np.array of similarities
         self.synonyms = {}  # type: Dict[str, Dict[str, np.ndarray]]
-        self.tokens, self.words, self.positions, self.punctuation = [], [], [], []  # type: List, List, List, List
+        self.tokens: 'spacy.tokens.Doc'
```
jklaise, line 289: We should be fine here as `self.set_text` is called first thing in `explain`

jklaise:
Hold on reviewing `cfproto.py` changes whilst I resolve the failing tests.

jklaise:
A few more failing tests for the second portion of `pytest` locally. I will place this PR back in draft mode until the issues have been resolved.

```
============================= test session starts ==============================
platform linux -- Python 3.8.12, pytest-6.2.5, py-1.10.0, pluggy-1.0.0
rootdir: /home/janis/PycharmProjects/public-fork-alibi, configfile: setup.cfg
plugins: forked-1.3.0, nbmake-0.8, xdist-2.3.0, cov-2.12.1, lazy-fixture-0.6.3, anyio-3.3.1, timeout-1.4.2, nbsmoke-0.5.0
collected 3803 items / 1648 deselected / 2155 selected

alibi/api/tests/test_interfaces.py ..........                            [  0%]
alibi/confidence/tests/test_model_linearity.py ......................... [  1%]
.................                                                        [  2%]
alibi/confidence/tests/test_trustscore.py ...                            [  2%]
alibi/explainers/tests/test_ale.py ..................................... [  4%]
.........................                                                [  5%]
alibi/explainers/tests/test_anchor_base.py ....                          [  5%]
alibi/explainers/tests/test_anchor_image.py ..........                   [  6%]
alibi/explainers/tests/test_anchor_tabular.py ..ss..FFFF..               [  6%]
alibi/explainers/tests/test_anchor_text.py ............ssss............. [  7%]
...................................................ssss................. [ 11%]
...............................................ssss..................... [ 14%]
..............................................                           [ 16%]
alibi/explainers/tests/test_cfrl.py ...........................F........ [ 18%]
..........                                                               [ 18%]
alibi/explainers/tests/test_counterfactual.py .....                      [ 19%]
alibi/explainers/tests/test_integrated_gradients.py .................... [ 20%]
........................................................................ [ 23%]
..........................................ssssssssssss                   [ 25%]
alibi/explainers/tests/test_shap_wrappers.py ........................... [ 27%]
........................................................................ [ 30%]
........................................................................ [ 33%]
........................................................................ [ 37%]
........................................................................ [ 40%]
........................................................................ [ 43%]
........................................................................ [ 47%]
........................................................................ [ 50%]
........................................................................ [ 53%]
........................................................................ [ 57%]
........................................................................ [ 60%]
........................................................................ [ 63%]
........................................................................ [ 67%]
........................................................................ [ 70%]
........................................................................ [ 73%]
........................................................................ [ 77%]
........................................................................ [ 80%]
........................................................................ [ 83%]
........................................................................ [ 87%]
........................................................................ [ 90%]
........................................................................ [ 94%]
...                                                                      [ 94%]
alibi/tests/test_datasets.py ......                                      [ 94%]
alibi/tests/test_saving.py ...F......                                    [ 94%]
alibi/tests/test_utils.py ..                                             [ 94%]
alibi/utils/tests/test_data.py ......                                    [ 95%]
alibi/utils/tests/test_distance.py ..................................... [ 96%]
............                                                             [ 97%]
alibi/utils/tests/test_distributed.py .................................. [ 99%]
......                                                                   [ 99%]
alibi/utils/tests/test_gradients.py ............                         [ 99%]
alibi/utils/tests/test_mapping.py .                                      [100%]

=================================== FAILURES ===================================
_ test_sampler[exp=<LazyFixture "at_adult_explainer">-clf=rf_<LazyFixture "adult_data">-dataset=<LazyFixture "adult_data">-anchors=((2,), (10,), (11,), (7, 10, 11), (3, 11))-nb_samples=100-test_instance_idx=0-predictor_type=proba] _
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/tests/test_anchor_tabular.py", line 251, in test_sampler
    cov_true, cov_false, labels, data, coverage, anchor_pos = sampler((0, anchor), nb_samples)
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/anchor_tabular.py", line 220, in __call__
    labels = self.compare_labels(raw_data)
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/anchor_tabular.py", line 242, in compare_labels
    return self.predictor(samples) == self.instance_label
AttributeError: 'TabularSampler' object has no attribute 'instance_label'
_ test_sampler[exp=<LazyFixture "at_adult_explainer">-clf=rf_<LazyFixture "adult_data">-dataset=<LazyFixture "adult_data">-anchors=((2,), (10,), (11,), (7, 10, 11), (3, 11))-nb_samples=100-test_instance_idx=0-predictor_type=class] _
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/tests/test_anchor_tabular.py", line 251, in test_sampler
    cov_true, cov_false, labels, data, coverage, anchor_pos = sampler((0, anchor), nb_samples)
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/anchor_tabular.py", line 220, in __call__
    labels = self.compare_labels(raw_data)
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/anchor_tabular.py", line 242, in compare_labels
    return self.predictor(samples) == self.instance_label
AttributeError: 'TabularSampler' object has no attribute 'instance_label'
_ test_sampler[exp=<LazyFixture "at_iris_explainer">-clf=rf_<LazyFixture "iris_data">-dataset=<LazyFixture "iris_data">-anchors=((2,), (10,), (11,), (7, 10, 11), (3, 11))-nb_samples=100-test_instance_idx=0-predictor_type=proba] _
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/tests/test_anchor_tabular.py", line 251, in test_sampler
    cov_true, cov_false, labels, data, coverage, anchor_pos = sampler((0, anchor), nb_samples)
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/anchor_tabular.py", line 220, in __call__
    labels = self.compare_labels(raw_data)
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/anchor_tabular.py", line 242, in compare_labels
    return self.predictor(samples) == self.instance_label
AttributeError: 'TabularSampler' object has no attribute 'instance_label'
_ test_sampler[exp=<LazyFixture "at_iris_explainer">-clf=rf_<LazyFixture "iris_data">-dataset=<LazyFixture "iris_data">-anchors=((2,), (10,), (11,), (7, 10, 11), (3, 11))-nb_samples=100-test_instance_idx=0-predictor_type=class] _
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/tests/test_anchor_tabular.py", line 251, in test_sampler
    cov_true, cov_false, labels, data, coverage, anchor_pos = sampler((0, anchor), nb_samples)
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/anchor_tabular.py", line 220, in __call__
    labels = self.compare_labels(raw_data)
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/anchor_tabular.py", line 242, in compare_labels
    return self.predictor(samples) == self.instance_label
AttributeError: 'TabularSampler' object has no attribute 'instance_label'
_ test_explainer[clf=rf_<LazyFixture "iris_data">-model=('iris-ae-tf2.2.0', 'iris-enc-tf2.2.0')] _
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/tests/test_cfrl.py", line 318, in test_explainer
    explainer.fit(X=iris_data["X_train"])
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/cfrl_tabular.py", line 277, in fit
    return super().fit(X)
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/cfrl_base.py", line 692, in fit
    sample = replay_buff.sample()
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/cfrl_base.py", line 188, in sample
    C = self.C[rand_idx] if (self.C is not None) else None  # conditional array if exists
AttributeError: 'ReplayBuffer' object has no attribute 'C'
------------------------------ Captured log setup ------------------------------
WARNING  tensorflow:load.py:133 SavedModel saved prior to TF 2.5 detected when loading Keras model. Please ensure that you are saving the model with model.save() or tf.keras.models.save_model(), *NOT* tf.saved_model.save(). To confirm, there should be a file named "keras_metadata.pb" in the SavedModel directory.
WARNING  tensorflow:load.py:133 SavedModel saved prior to TF 2.5 detected when loading Keras model. Please ensure that you are saving the model with model.save() or tf.keras.models.save_model(), *NOT* tf.saved_model.save(). To confirm, there should be a file named "keras_metadata.pb" in the SavedModel directory.
WARNING  tensorflow:load.py:171 No training configuration found in save file, so the model was *not* compiled. Compile it manually.
----------------------------- Captured stderr call -----------------------------

  0%|          | 0/100 [00:00<?, ?it/s]
  7%|▋         | 7/100 [00:00<00:01, 65.53it/s]
 11%|█         | 11/100 [00:00<00:01, 58.52it/s]
_____________________ test_save_AnchorTabular[adult_data] ______________________
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/anchor_tabular.py", line 1033, in _transform_predictor
    prediction = predictor(x)
TypeError: 'NoneType' object is not callable

The above exception was the direct cause of the following exception:

  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/tests/test_saving.py", line 359, in test_save_AnchorTabular
    atab_explainer.save(temp_dir)
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/api/interfaces.py", line 116, in save
    save_explainer(self, path)
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/saving.py", line 102, in save_explainer
    save_fn(explainer, path)
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/saving.py", line 107, in _simple_save
    explainer.predictor = None  # type: ignore
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/anchor_tabular.py", line 1024, in predictor
    self._predictor = self._transform_predictor(predictor)
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/anchor_tabular.py", line 1037, in _transform_predictor
    raise AlibiPredictorCallException(msg) from e
alibi.exceptions.AlibiPredictorCallException: Predictor failed to be called on <class 'numpy.ndarray'> of shape (1, 12) and dtype float32. Check that the parameter `feature_names` is correctly specified.

---------- coverage: platform linux, python 3.8.12-final-0 -----------
Name                                                   Stmts   Miss  Cover
--------------------------------------------------------------------------
alibi/__init__.py                                          3      0   100%
alibi/api/__init__.py                                      0      0   100%
alibi/api/defaults.py                                     22      0   100%
alibi/api/interfaces.py                                   84      4    95%
alibi/api/tests/__init__.py                                0      0   100%
alibi/api/tests/test_interfaces.py                        74      5    93%
alibi/confidence/__init__.py                               4      0   100%
alibi/confidence/model_linearity.py                      187      5    97%
alibi/confidence/tests/__init__.py                         0      0   100%
alibi/confidence/tests/test_model_linearity.py           157      0   100%
alibi/confidence/tests/test_trustscore.py                 41      0   100%
alibi/confidence/trustscore.py                            79      2    97%
alibi/data/__init__.py                                     0      0   100%
alibi/datasets.py                                        109      9    92%
alibi/exceptions.py                                        8      0   100%
alibi/explainers/__init__.py                              16      0   100%
alibi/explainers/_influence.py                            22     22     0%
alibi/explainers/ale.py                                  199      8    96%
alibi/explainers/anchor_base.py                          316     23    93%
alibi/explainers/anchor_explanation.py                    44     16    64%
alibi/explainers/anchor_image.py                         162      4    98%
alibi/explainers/anchor_tabular.py                       419     32    92%
alibi/explainers/anchor_text.py                          424     41    90%
alibi/explainers/backends/__init__.py                      0      0   100%
alibi/explainers/backends/cfrl_base.py                    46      8    83%
alibi/explainers/backends/cfrl_tabular.py                219     50    77%
alibi/explainers/backends/pytorch/__init__.py              0      0   100%
alibi/explainers/backends/pytorch/cfrl_base.py           156    112    28%
alibi/explainers/backends/pytorch/cfrl_tabular.py         47      1    98%
alibi/explainers/backends/tensorflow/__init__.py           0      0   100%
alibi/explainers/backends/tensorflow/cfrl_base.py        154     42    73%
alibi/explainers/backends/tensorflow/cfrl_tabular.py      54      2    96%
alibi/explainers/baseold.py                               97     97     0%
alibi/explainers/cem.py                                  347     71    80%
alibi/explainers/cf_debug.py                              17     17     0%
alibi/explainers/cfproto.py                              651    157    76%
alibi/explainers/cfrl_base.py                            293     26    91%
alibi/explainers/cfrl_tabular.py                         126     46    63%
alibi/explainers/counterfactual.py                       294     71    76%
alibi/explainers/integrated_gradients.py                 379     13    97%
alibi/explainers/session.py                               21     21     0%
alibi/explainers/shap_wrappers.py                        500     23    95%
alibi/explainers/tests/__init__.py                         0      0   100%
alibi/explainers/tests/conftest.py                       169      3    98%
alibi/explainers/tests/test_ale.py                        86      0   100%
alibi/explainers/tests/test_anchor_base.py                45      0   100%
alibi/explainers/tests/test_anchor_image.py               97      0   100%
alibi/explainers/tests/test_anchor_tabular.py            184      0   100%
alibi/explainers/tests/test_anchor_text.py               219     15    93%
alibi/explainers/tests/test_cem.py                        28      0   100%
alibi/explainers/tests/test_cfproto.py                    93      0   100%
alibi/explainers/tests/test_cfrl.py                      274      2    99%
alibi/explainers/tests/test_counterfactual.py            100      0   100%
alibi/explainers/tests/test_integrated_gradients.py      337      9    97%
alibi/explainers/tests/test_shap_wrappers.py             887     26    97%
alibi/explainers/tests/utils.py                           50      2    96%
alibi/models/__init__.py                                   0      0   100%
alibi/models/pytorch/__init__.py                           0      0   100%
alibi/models/pytorch/actor_critic.py                      29     20    31%
alibi/models/pytorch/autoencoder.py                       23     23     0%
alibi/models/pytorch/cfrl_models.py                       80     80     0%
alibi/models/pytorch/metrics.py                           57     57     0%
alibi/models/pytorch/model.py                            138    138     0%
alibi/models/pytorch/wrapper.py                            3      3     0%
alibi/models/tensorflow/__init__.py                        0      0   100%
alibi/models/tensorflow/actor_critic.py                   28      0   100%
alibi/models/tensorflow/autoencoder.py                    23     10    57%
alibi/models/tensorflow/cfrl_models.py                    74     74     0%
alibi/models/tensorflow/wrapper.py                         3      3     0%
alibi/models/wrapper.py                                    2      2     0%
alibi/saving.py                                          210     19    91%
alibi/tests/__init__.py                                    0      0   100%
alibi/tests/test_datasets.py                              49      4    92%
alibi/tests/test_saving.py                               252      1    99%
alibi/tests/test_utils.py                                 14      0   100%
alibi/tests/utils.py                                      63     13    79%
alibi/utils/__init__.py                                    0      0   100%
alibi/utils/_types.py                                      9      9     0%
alibi/utils/approximation_methods.py                      44      8    82%
alibi/utils/data.py                                       42     10    76%
alibi/utils/discretizer.py                                40      3    92%
alibi/utils/distance.py                                  106      2    98%
alibi/utils/distributed.py                               225     48    79%
alibi/utils/distributions.py                               5      0   100%
alibi/utils/download.py                                   10      4    60%
alibi/utils/frameworks.py                                 14      2    86%
alibi/utils/gradients.py                                  29      0   100%
alibi/utils/lang_model.py                                119      4    97%
alibi/utils/mapping.py                                    67      2    97%
alibi/utils/tests/__init__.py                              0      0   100%
alibi/utils/tests/conftest.py                            150     84    44%
alibi/utils/tests/test_data.py                            17      0   100%
alibi/utils/tests/test_distance.py                        64      0   100%
alibi/utils/tests/test_distributed.py                    207     11    95%
alibi/utils/tests/test_gradients.py                       35      0   100%
alibi/utils/tests/test_mapping.py                         21      0   100%
alibi/utils/tf.py                                         26      5    81%
alibi/utils/visualization.py                             106    106     0%
alibi/utils/wrappers.py                                   26      8    69%
alibi/version.py                                           1      0   100%
--------------------------------------------------------------------------
TOTAL                                                  10751   1738    84%

=========================== short test summary info ============================
FAILED alibi/explainers/tests/test_anchor_tabular.py::test_sampler[exp=<LazyFixture "at_adult_explainer">-clf=rf_<LazyFixture "adult_data">-dataset=<LazyFixture "adult_data">-anchors=((2,), (10,), (11,), (7, 10, 11), (3, 11))-nb_samples=100-test_instance_idx=0-predictor_type=proba]
FAILED alibi/explainers/tests/test_anchor_tabular.py::test_sampler[exp=<LazyFixture "at_adult_explainer">-clf=rf_<LazyFixture "adult_data">-dataset=<LazyFixture "adult_data">-anchors=((2,), (10,), (11,), (7, 10, 11), (3, 11))-nb_samples=100-test_instance_idx=0-predictor_type=class]
FAILED alibi/explainers/tests/test_anchor_tabular.py::test_sampler[exp=<LazyFixture "at_iris_explainer">-clf=rf_<LazyFixture "iris_data">-dataset=<LazyFixture "iris_data">-anchors=((2,), (10,), (11,), (7, 10, 11), (3, 11))-nb_samples=100-test_instance_idx=0-predictor_type=proba]
FAILED alibi/explainers/tests/test_anchor_tabular.py::test_sampler[exp=<LazyFixture "at_iris_explainer">-clf=rf_<LazyFixture "iris_data">-dataset=<LazyFixture "iris_data">-anchors=((2,), (10,), (11,), (7, 10, 11), (3, 11))-nb_samples=100-test_instance_idx=0-predictor_type=class]
FAILED alibi/explainers/tests/test_cfrl.py::test_explainer[clf=rf_<LazyFixture "iris_data">-model=('iris-ae-tf2.2.0', 'iris-enc-tf2.2.0')]
FAILED alibi/tests/test_saving.py::test_save_AnchorTabular[adult_data] - alib...
=== 6 failed, 2123 passed, 26 skipped, 1648 deselected in 1341.01s (0:22:21) ===
```

alibi/explainers/cfrl_base.py
```diff
@@ -97,6 +96,7 @@ def __init__(self, size: int = 1000) -> None:
         self.len = 0  # current length of the buffer
         self.size = size  # buffer's maximum capacity
         self.batch_size = 0  # batch size (inferred during `append`)
+        self.C: Optional[np.ndarray] = None  # buffer for the conditional tensor
```
jklaise, line 99: So `C` has a bit more logic than the other buffers, so seems like `None` is in fact a valid type.

I'm not super comfortable with the current refactor as the buffer declarations are now split into two places. Would be good if you could have a look here too @RobertSamoilescu .

jklaise:
The failing test in `test_save` is to do with `self.predictor=None` being routed to the `predictor` property setter which is assuming that the passed `predictor` is a *bona fide* predictor and not `None` as in the case of the saving function. This now explains why `None` was accepted as a valid input to the `predictor` setter. I will need to think how to resolve this cleanly.

jklaise:
PR is ready for review. I think the general guideline is that if you feel uncomfortable with any of the logic changes I'm happy to revert these and use `type: ignore` instead to silence the type checker @ascillitoe @arnaudvl @RobertSamoilescu .

alibi/explainers/cfproto.py
ascillitoe: Seems like it might be best to open a new issue about raising a warning for incorrect combinations of kwargs?

alibi/explainers/cfrl_base.py
ascillitoe: Deferring to @RobertSamoilescu on this one.

alibi/explainers/cfrl_base.py
ascillitoe: I'm also one for aligning nicely, but all IDE's seem to disagree with us...

alibi/explainers/cfrl_base.py
ascillitoe: Is this still to fix then?

alibi/explainers/integrated_gradients.py
ascillitoe: Mmn this one seems annoying...

alibi/utils/lang_model.py
```diff
@@ -33,6 +33,14 @@
 class LanguageModel(abc.ABC):
     SUBWORD_PREFIX = ''
 
+    # We don't type transformers objects here as it would likely require digging into
+    # some private base classes which may change in the future and cause breaking changes.
+    model: Any
+    tokenizer: Any
```
ascillitoe, line 39: Perhaps at some point we need to think more about what tokenizers we actually support, and type accordingly? I've been typing with `PreTrainedTokenizerBase` in the Alibi Detect config work as we instantiate the tokenizer with `AutoTokenizer.from_pretrainer`, but maybe this needs more thought...

alibi/utils/tests/test_distributed.py
ascillitoe: Well that is strange/worrying...

ascillitoe (commented):
I've commented on a few things, but this mostly looks good to me. Unless I've missed some critical issues none of the refactorings look like too much of an issue to me, except for the issue with correct kwarg combinations not being checked (although this is likely something for a new issue). 

As a more general comment, it might make sense for us to start being more specific with our `# type: ignore` statements (https://mypy.readthedocs.io/en/stable/error_codes.html). This would tighten up our mypy checks, but also make the code a little more readable.

alibi/explainers/anchor_tabular.py
```diff
@@ -1,28 +1,34 @@
...
 
+from alibi.api.defaults import DEFAULT_DATA_ANCHOR, DEFAULT_META_ANCHOR
 from alibi.api.interfaces import Explainer, Explanation, FitMixin
-from alibi.api.defaults import DEFAULT_META_ANCHOR, DEFAULT_DATA_ANCHOR
-from alibi.exceptions import AlibiPredictorCallException, AlibiPredictorReturnTypeError
-from .anchor_base import AnchorBaseBeam, DistributedAnchorBaseBeam
-from .anchor_explanation import AnchorExplanation
-from alibi.utils.wrappers import ArgmaxTransformer
+from alibi.exceptions import (AlibiPredictorCallException,
+                              AlibiPredictorReturnTypeError)
```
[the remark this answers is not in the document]
jklaise: Yup, I ran `isort` on the changed modules so the imports are more readable generally.

alibi/explainers/anchor_tabular.py
jklaise: I don't think any errors are raised currently, we should definitely validate and do it. Perhaps a new issue.

jklaise:
Thanks Ashley, here are some steps I will do:

- [x] Open issue re validation of combination of user arguments
- [ ] Go over newly introduced `type: ignore`s and see if I can narrow down the codes (#546)
- [ ] Consult @RobertSamoilescu on the `CFRL` refactor
- [ ] Fix any outstanding TODOs/notes to self

jklaise:
@ascillitoe I'm also starting to think we may want to enable variable redefinition (e.g. see here: https://mypy-lang.blogspot.com/2019/02/mypy-0670-released.html). This avoids having to deal with a very common pattern like e.g.

```python
def fun() -> np.ndarray:
    X = []  # how do we type this? Union[list, np.ndarray]? but then can't do append on an array 
    for i in range(10):
        X.append(array)
    X = np.concatenate(X)  # different type here

return X  # expect np.ndarray, but we need to either explicitly cast or redefine the type of X
```

As far as I can see there is no good way to type this without allowing variable redefinition. One might say just use a different variable when concatenating and creating an array, but this is busywork and doesn't seem very clean leaving around variables which we won't use.

ascillitoe:
@jklaise thanks for the above, I'd encountered the same issue with the config work, but wasn't aware there was a mypy setting for this. I ended up doing the latter, i.e. introducing new variables, which arguably did make the code more readable too. I can think of examples where the opposite might be true though, e.g. having to introduce new names into the alibi detect operation `x = preprocess_fn(x)` would IMO make it less readable. 

Maybe a compromise would be to allow mypy variable redefinition, but make an effort to avoid doing this in cases where it makes the code less readable?

alibi/tests/utils.py
```diff
@@ -55,7 +56,7 @@ def __call__(self, *args, **kwargs):
         elif self.out_type == 'raw' or self.out_type == 'log_loss' or self.out_type == 'continuous':
             return self._generate_logits(sz, *args, **kwargs)
 
-    def _generate_probas(self, sz: tuple = None, *args, **kwargs) -> np.ndarray:
+    def _generate_probas(self, sz: Optional[tuple] = None, *args, **kwargs) -> np.ndarray:
```
RobertSamoilescu, line 59: I think that there might be a problem with `sz` being `None` or `tuple()`:
https://github.com/SeldonIO/alibi/blob/785d2183b10a98dc6632de34023fe51f8f874571/alibi/tests/utils.py#L73
https://github.com/SeldonIO/alibi/blob/785d2183b10a98dc6632de34023fe51f8f874571/alibi/tests/utils.py#L90

I've noticed that some `numpy` functions can deal with the size being `None` or an empty tuple, but it doesn't seem to hold for all function such as `dirichlet`

alibi/tests/utils.py
```diff
@@ -88,15 +89,15 @@ def _generate_probas(self, sz: tuple = None, *args, **kwargs) -> np.ndarray:
 
         return np.random.dirichlet(alpha, size=sz)
 
-    def _generate_labels(self, sz: tuple = None, *args, **kwargs) -> np.ndarray:
+    def _generate_labels(self, sz: Optional[tuple] = None, *args, **kwargs) -> np.ndarray:
```
RobertSamoilescu, line 92: Same observation as above. Seems like `randint` can deal with `None` or empty tuples

alibi/tests/utils.py
```diff
@@ -88,15 +89,15 @@ def _generate_probas(self, sz: tuple = None, *args, **kwargs) -> np.ndarray:
...
         """
         Generates labels by sampling random integers in range(0, n_classes+1).
         """
         if sz:
             sz += (self.out_dim,)
         return np.random.randint(0, self.out_dim + 1, size=sz)
 
-    def _generate_logits(self, sz: tuple = None, *args, **kwargs) -> np.ndarray:
+    def _generate_logits(self, sz: Optional[tuple] = None, *args, **kwargs) -> np.ndarray:
```
RobertSamoilescu, line 100: Same observation as above. Seems like `randint` can deal with `None` or empty tuples

alibi/explainers/anchor_base.py
```diff
@@ -175,7 +176,7 @@ def compute_beta(n_features: int, t: int, delta: float) -> float:
 
         return temp + np.log(temp)
 
-    def _get_coverage_samples(self, coverage_samples: int, samplers: List[Callable] = None) -> np.ndarray:
+    def _get_coverage_samples(self, coverage_samples: int, samplers: Optional[List[Callable]] = None) -> np.ndarray:
```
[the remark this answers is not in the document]
jklaise: No, but it is used in the inherited class `DistributedAnchorBaseBeam` which means that the call to the MAB must be the same (i.e. include `samplers`) from within the explainer classes in both cases.

alibi/explainers/anchor_base.py
```diff
@@ -835,7 +836,7 @@ def __init__(self, samplers: List[Callable], **kwargs) -> None:
         self.pool = ActorPool(samplers)
         self.samplers = samplers
 
-    def _get_coverage_samples(self, coverage_samples: int, samplers: List[Callable] = None) -> np.ndarray:
+    def _get_coverage_samples(self, coverage_samples: int, samplers: List[Callable]) -> np.ndarray:  # type: ignore
```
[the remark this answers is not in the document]
jklaise: It's because this is an inherited class that overrides the parent method, but the parent methods has `samplers` as `Optional` whereas here it is required, this results in a typing error. I will see if I can narrow down the ignores in this PR to more specific error codes.

alibi/explainers/anchor_image.py
```diff
@@ -54,7 +55,7 @@ def __init__(
             segmentation_fn: Callable,
             custom_segmentation: bool,
             image: np.ndarray,
-            images_background: np.ndarray = None,
+            images_background: Optional[np.ndarray] = None,
```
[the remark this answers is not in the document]
jklaise: Yes I will rebase against master.

alibi/tests/utils.py
```diff
@@ -15,8 +16,8 @@ class MockPredictor:
     def __init__(self,
                  out_dim: int,
                  out_type: str = 'proba',
-                 model_type: str = None,
-                 seed: int = None,
+                 model_type: Optional[str] = None,
```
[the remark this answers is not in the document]
jklaise: Doesn't seem to be, will remove.

alibi/tests/utils.py
jklaise: I'll see if I can validate `sz` more, although it's a minor issue for a test utility function.

alibi/explainers/backends/cfrl_tabular.py
jklaise: The issue here is with the inferred type for `feature_types`, `mypy` can't see that it can't be `None` due to it being inside an inner function which is a limitation referred to in the link above. I will see if I can reduce the scope of the ignore with an error code though.

jklaise:
Enabling strict optional **and** testing against the latest `numpy` has flagged quite a few more issues, I will attempt to address them as part of this PR.

alibi/tests/utils.py
jklaise: I've just checked that all the considered functions, including `dirichlet`, can take `size=None` or `size=()` as an argument, so I will leave these unchanged.

alibi/explainers/backends/pytorch/cfrl_base.py
```diff
@@ -63,9 +63,9 @@ def __init__(self,
                                         batch_size=self.batch_size)
 
         # Define number of classes for classification & minimum and maximum labels for regression
-        self.num_classes: Optional[int] = None
-        self.min_m: Optional[float] = None
-        self.max_m: Optional[float] = None
+        self.num_classes: int
+        self.min_m: float
+        self.max_m: float
```
jklaise, line 68: I avoided initializing these to `None` as those throw `mypy` errors when passed into the `numpy.random.randint` function.

alibi/explainers/cfrl_base.py
```diff
@@ -839,7 +832,7 @@ def explain(self,
 
         # Perform prediction in mini-batches.
         n_minibatch = int(np.ceil(X.shape[0] / batch_size))
-        all_results: Dict[str, np.ndarray] = {}
+        all_results: Dict[str, Optional[np.ndarray]] = {}
```
jklaise, line 835: Changed this to `Optional` because `C` can be `None`. However, this complicates typing further down as all other entries cannot be `None`, but the type checker doesn't know that. A `TypedDict` might be a better fit here in the future.

jklaise:
Reverted removing `model_type` attribute from `MockPredictor` as it is actually being used to emulate the model wrapper class behaviour from the `shap` library. Added also a comment for future reference.

ascillitoe (approved):

alibi/explainers/cfproto.py
DBrughmans: #610 
De default is not over the keys. Both functions generate different outputs.

cat_vars
{0: 2, 2: 3, 5: 8, 13: 2, 15: 2, 17: 2, 19: 3}
max(cat_vars)
19
max(cat_vars,key=cat_vars.get)
5

PR #542, opened 2021-11-25
Fix bugs with passing images_background to AnchorImage

### LINKED ISSUE #540: 'UnboundLocalError: local variable 'fudged_image' referenced before assignment' in AnchorImage
Providing background images to the `explain` method of `AnchorImage` leads to an `UnboundLocalError`.

Specifically, in [line 257 of the `perturbation` method of `AnchorImageSampler`](https://github.com/SeldonIO/alibi/blob/00f7391132b761b9e28178bbde11ec2eed414ea5/alibi/explainers/anchor_image.py#L257) the variable `fudged_image` has not been defined, because it is only set in [line 235](https://github.com/SeldonIO/alibi/blob/00f7391132b761b9e28178bbde11ec2eed414ea5/alibi/explainers/anchor_image.py#L235) if no background images have been provided.

jklaise:
Resolves #540.

I have manually checked that the code now works passing an array of `images_background` and that the proposed anchor is actually superimposed on these background images. I will see if an automated test can also be written relatively easily.

### DIFF
--- a/alibi/explainers/anchor_image.py
+++ b/alibi/explainers/anchor_image.py
@@ -248,11 +248,9 @@ class AnchorImageSampler:
                 mask[segments == superpixel] = True
             if background_idx is not None:
                 # replace values with those of background image
-                # TODO: Could images_background be None herre?
                 temp[mask] = self.images_background[background_idx][mask]
             else:
                 # ... or with the averaged superpixel value
-                # TODO: Where is fudged_image defined?
                 temp[mask] = fudged_image[mask]
             pert_imgs.append(temp)
 
--- a/alibi/explainers/anchor_image.py
+++ b/alibi/explainers/anchor_image.py
@@ -221,13 +221,12 @@ class AnchorImageSampler:
         segments_mask[:, anchor] = 1
 
         # for each sample, need to sample one of the background images if provided
-        if self.images_background:
+        if self.images_background is not None:
             backgrounds = np.random.choice(
                 range(len(self.images_background)),
                 segments_mask.shape[0],
                 replace=True,
             )
-            segments_mask = np.hstack((segments_mask, backgrounds.reshape(-1, 1)))
         else:
             backgrounds = [None] * segments_mask.shape[0]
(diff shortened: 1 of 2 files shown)


alibi/explainers/anchor_image.py
```diff
@@ -221,13 +221,12 @@ def perturbation(
         segments_mask[:, anchor] = 1
 
         # for each sample, need to sample one of the background images if provided
-        if self.images_background:
+        if self.images_background is not None:
```
jklaise, line 224: `self.images_background` is either `None` or `np.ndarray` so we must check explicitly for `None` as `np.ndarray` doesn't have a working `__bool__` method.

alibi/explainers/anchor_image.py
```diff
@@ -247,7 +246,7 @@ def perturbation(
             mask = np.zeros(segments.shape).astype(bool)
             for superpixel in to_perturb:
                 mask[segments == superpixel] = True
-            if background_idx:
+            if background_idx is not None:
```
jklaise, line 249: The main bug. Here `background_idx` can be `0` but in the previous version it would cause it to go down this (wrong) branch and end up using `fudged_image` which is not defined. The fix is to explicitly check for `None`.

alibi/explainers/anchor_image.py
```diff
@@ -221,13 +221,12 @@ def perturbation(
...
 
         # for each sample, need to sample one of the background images if provided
-        if self.images_background:
+        if self.images_background is not None:
             backgrounds = np.random.choice(
                 range(len(self.images_background)),
                 segments_mask.shape[0],
                 replace=True,
             )
-            segments_mask = np.hstack((segments_mask, backgrounds.reshape(-1, 1)))
```
jklaise, line 230: This line appears to be completely random and just doesn't work.

jklaise:
Added a simple test by parametrizing `images_background`. PR should be ready for review.

RobertSamoilescu (approved):
@jklaise, looks good to me!

Issue #229, opened 2020-05-22
Update docstrings


jklaise:
Some docstrings are out of date, some need consolidation. of conventions.

 - CF methods still say they return `dict` instead of `Explanation`.
 - Conventions for referring to instance * features tensors (consistent naming and math notation)
 - Describe `Explanation` object returned consistently, potentially include the keys for each method or as a minimum a link to docs?

ascillitoe:
@RobertSamoilescu @jklaise I thought I would use this issue to track some of the other related issues we are having with docstrings. @RobertSamoilescu is working on some of these in #548.

#### 1) Duplicated class attributes

```python
class TabularSampler:
    """ 
    A sampler that uses an underlying training set to draw records that have a subset of features with
    values specified in an instance to be explained, `X`.

   Attributes
   -------------
   instance_label: Description here.
   """

   instance_label: int

   def __init__(...):
```

This leads to `instance_label` being present twice in the docs for `TabularSampler`. I suspect forcibly removing `undoc_members` from the api rst files would fix this, but it might have other unintended consequences. An easier fix is to do 

```python
class TabularSampler:
    """ 
    A sampler that uses an underlying training set to draw records that have a subset of features with
    values specified in an instance to be explained, `X`.
   """

   instance_label: int  #: Description here.

   def __init__(...):
```

The only issue with this might be when we have inline type hints, as @RobertSamoilescu has discovered we can't have two sets of inline comments e.g. for describing the attribute, and a type hint. For example:

```
attr = {} # type: dict #: A description.
```

does not work. However, we can instead do:

```
attr: dict = {} #: A description.
```

This requires the newer inline type hint, but we might be OK here as this is supported from python 3.6 onwards (https://www.python.org/dev/peps/pep-0526/).

Other options are:

```
#: A description.
attr = {} # type: dict
```

or:

```
attr = {} # type: dict
"""A description."""
```

I think both might work (haven't tested, but they require an extra line).

#### 2) Inconsistency in duplication of return args

In some cases, return descriptions are italicized, and in some cases they are not. For example:

Not italicized: https://docs.seldon.io/projects/alibi/en/latest/api/alibi.explainers.ale.html#alibi.explainers.ale.minimum_satisfied
Italicized: https://docs.seldon.io/projects/alibi/en/latest/api/alibi.explainers.ale.html#alibi.explainers.ale.plot_ale

[...]

Issue #465, opened 2021-08-23
More advanced linting to make PR checking more automated


jklaise:
The recent [PR checklist](https://github.com/SeldonIO/alibi/blob/master/CONTRIBUTING.md#pr-checklist) outlines what to check before a PR is considered complete, however, a portion of these are mundane checks that can be automated with existing tools, in particular using some `flake8` plugins. For example:

- https://github.com/sco1/flake8-annotations (detect absence of type hints)
- https://gitlab.com/pycqa/flake8-docstrings (check docstrings are present)
- https://github.com/PyCQA/flake8-bugbear (additional code correctness checks - running it on `alibi` revealed a few minor bugs!)
- https://github.com/terrencepreilly/darglint (Python docstring linter for checking that the docstring matches the definition)
- https://github.com/PyCQA/flake8-import-order (check import order follows a style guide)
- https://github.com/PyCQA/pep8-naming (PEP8 naming convention checker)

Not all the plugins are necessarily required and the adoption can be gradual (e.g. running `flake8-annotations` and figuring out the most appropriate configuration to go in `setup.cfg` in one PR before tackling others).

Same applies to Alibi Detect.

PR #548, opened 2021-12-02
Docstrings updates

RobertSamoilescu:
This PR addresses some of the issues listed in #229.

Docstring conventions:

- Variable, module, function, and class names should be written between single back-ticks:
     - A \`numpy\` scalar type that 
     - \`X\`, 
     - \`extrapolate_constant_perc\`

- simple math should be written between single back-ticks to be more readable in the console:
     - A callable that takes an \`N x F\` tensor, for
     - \`x >= v, fun(x) >= target\`

- complex math should be written in LaTeX:
    - function where :math:\`link(output - expected\_value) = sum(\phi)\`

- variable values or examples setting an argument to a specific values should be written in double back-ticks (this would make them more readable as they are rendered in a block in red/orange color):
   - is set to \`\`True\`\`
   - A list of features for which to plot the ALE curves or \`\`'all'\`\` for all features.
   - Possible values are: \`\`'all'\`\` | \`\`'row'\`\` | \`\`None\`\`.
   - The search is greedy if \`\`beam_size=1\`\`
   - if the result uses \`\`segment_labels=(1, 2, 3)\`\` and \`\`partial_index=1\`\`, this will return \`\`[1, 2]\`\`.
   
- returning the name of the variable and its description (this is standard convention and renders well)
```
Returns
-------
raw
    Array of perturbed text instances.
data
    Matrix with 1s and 0s indicating whether a word in the text has not been perturbed for each sample.
```

- returning just the description. Note that when we do not return the name of the variable, then sphinx wrongly interprets the description as the variable name and will render the text in italic. Moreover, we need to include a \\ after each line to avoid introducing bullet points for each row. Observation: when we include the name of variable/module/... between the single back-ticks, the text is rendered in normal font and only the name of the variable/module/... will be rendered in italic. 
```
Returns
-------
If the user has specified grouping, then the input object is subsampled and an object of the same \
type is returned. Otherwise, a `shap_utils.Data` object containing the result of a k-means algorithm \
is wrapped in a `shap_utils.DenseData` object and returned. The samples are weighted according to the \
frequency of the occurrence of the clusters in the original data.
```
 
- returning an object which contains multiple attributes and each attribute is explained. In this case the attribute name is written between single back-ticks and the type (if provided) would be written in double back-ticks.
```
Returns
-------
`Explanation` object containing the anchor explaining the instance with additional metadata as attributes. \
Contains the following data-related attributes

 - `anchor` : ``List[str]`` - a list of words in the proposed anchor.

 - `precision` : ``float`` - the fraction of times the sampled instances where the anchor holds yields \
 the same prediction as the original instance. The precision will always be  threshold for a valid anchor.

 - `coverage` : ``float`` - the fraction of sampled instances the anchor applies to.
```

- documenting a dictionary or a returned dictionary follows the same principle as above only that instead of the attribute name we write the key between double back-ticks:
```
Default perturbation options for ``'similarity'`` sampling

    - ``'sample_proba'`` : ``float`` - probability of a word to be masked.

    - ``'top_n'`` : ``int`` - number of similar words to sample for perturbations.

    - ``'temperature'`` : ``float`` - sample weight hyper-parameter if `use_proba=True`.

    - ``'use_proba'`` : ``bool`` - whether to sample according to the words similarity.
```

- attributes are commented inline to avoid duplication
```
class ReplayBuffer:
    """
    Circular experience replay buffer for `CounterfactualRL` (DDPG). When the buffer is filled, then the oldest
    experience is replaced by the new one (FIFO). The experience batch size is kept constant and inferred when
    the first batch of data is stored. Allowing flexible batch size can generate `tensorflow` warning due to
    the `tf.function` retracing, which can lead to a drop in performance.
    """
    X: np.ndarray  #: Inputs buffer.
    Y_m: np.ndarray  #: Model's prediction buffer.
    Y_t: np.ndarray  #: Counterfactual targets buffer.
    Z: np.ndarray  #: Input embedding buffer.
    Z_cf_tilde: np.ndarray  #: Noised counterfactual embedding buffer.
    R_tilde: np.ndarray  #: Noise counterfactual rewards buffer.
```


alibi/explainers/ale.py
```diff
@@ -272,7 +294,9 @@ def minimum_satisfied(values: np.ndarray, min_bin_points: int, n: int) -> int:
 
     Returns
     -------
+    int
```
RobertSamoilescu, line 297: Possible fix for the italic/normal font issue in the return.

alibi/explainers/ale.py
```diff
@@ -244,7 +265,8 @@ def bisect_fun(fun: Callable, target: float, lo: int, hi: int) -> int:
 
     Returns
     -------
-    Integer index.
+    lo : int
```
RobertSamoilescu, line 268: Possible fix for the italic/normal font issue in return.

jklaise: I would avoid repeating the type return information as that is already given by the type hint and is parsed into the static docs. Repeating this manually risks inconsistencies as time goes by. I would rather stick with slightly inconsistent formatting than risk type information being out of sync.

alibi/explainers/ale.py
```diff
@@ -171,13 +184,13 @@ def explain(self, X: np.ndarray, features: List[int] = None, min_bin_points: int
...
         )
 
-    def build_explanation(self,
-                          ale_values: List[np.ndarray],
-                          ale0: List[np.ndarray],
-                          constant_value: float,
-                          feature_values: List[np.ndarray],
-                          feature_deciles: List[np.ndarray],
-                          feature_names: np.ndarray) -> Explanation:
+    def __build_explanation(self,
```
jklaise, line 187: Is there a reason for the double-under instead of single-under?

RobertSamoilescu: No ... I meant single-under. I will update it.

alibi/explainers/ale.py
```diff
@@ -106,7 +106,20 @@ def explain(self, X: np.ndarray, features: List[int] = None, min_bin_points: int
...
+             can have multiple columns (if the number of targets is >1 as in classification).
+             - `constant_value` : float - the mean prediction over  (zeroth order effects).
+             - `ale0` : List[np.ndarray] - a list of “centering” values (one for each feature) used by the algorithm \
+             to center the `ale_values` around the expected effect for the feature (i.e. the sum of `ale_values` \
+             and `ale0` will be the uncentered ALE).
+             - `feature_values` : List[np.ndarray] - a list of arrays (one for each feature) of feature values at \
+             which the ALE values were computed.
+             - `feature_names` : np.ndarray - a list of feature names.
+             - `target_names` : np.ndarray - a list of target names.
+             - `feature_deciles` : List[np.ndarray] - a list of arrays (one for each feature) of the feature deciles.
```
jklaise, line 122: A quick idea to test - instead of repeating all this information (which is on the methods page) and risk information going out of sync, can we link to the relevant page/section insead? Could you check if a relative link works? You may need to look up hyperlink/relative link syntax in `.rst` files.

jklaise: On second thought this is not a good idea since the library isn't shipped with any external documentation and the docstrings should be fairly self-contained. Let's leave this as is and make a mental note for any changes to the returned `Explanation` to update docs in two places. In the future this would be hopefully superseeded by having `pydantic` explanation types we can link to which would be self-documenting.

jklaise (commented):
For the `api.defaults` module I think we need to clarify the `KERNEL_SHAP_PARAMS` and `TREE_SHAP_PARAMS` docstrings (e.g. `KernelShap parameters updated and return in metadata[‘params’].`) Also, I think it would be nice if we can use the sphinx directives to link to the classes, e.g. ```:py:class:`alibi.explainers.AnchorTabular``` etc.

alibi/explainers/ale.py
```diff
@@ -38,39 +38,39 @@ def __init__(self,
...
+            data points, `F` - number of features, `T` - number of outputs/targets (e.g. 1 for single output
+            regression, >=2 for classification)).
         feature_names
             A list of feature names used for displaying results.
         target_names
             A list of target/output names used for displaying results.
         check_feature_resolution
-            If true, the number of unique values is calculated for each feature and if it is less than
-            `low_resolution_threshold` then the feature values are used for gridpoints instead of quantiles.
+            If ``True``, the number of unique values is calculated for each feature and if it is less than
```
jklaise, line 49: This is interesting, I wasn't aware single backticks generate italic text and double backticks generate the red text in boxes.

For style, I'm thinking whether we should also use double backticks for references to parameters from other parameter definitions, e.g.
```
Only used if ``check_feature_resolution=True``. 
```
The it seems to me more clear which parameter has which effect on other parameters by just glancing at the API docs as references to parameter names would be highlighted.

@ascillitoe

alibi/explainers/ale.py
```diff
@@ -106,8 +106,12 @@ def explain(self, X: np.ndarray, features: Optional[List[int]] = None, min_bin_p
 
         Returns
         -------
+        explanation
             An `Explanation` object containing the data and the metadata of the calculated ALE curves.
+            See usage `examples`_ for details.
 
+            .. _examples:
+                https://docs.seldon.io/projects/alibi/en/latest/methods/ALE.html
```
jklaise, line 114: The formatting here is off on the built docs and the link doesn't work. I'm not exactly sure what the issue is, there are similar instances in `KernelShap` where some links have the same issue but others work as expected.

jklaise: This is probably some really annoying corner case of formatting, maybe @ascillitoe has come across this...

jklaise: Inline link works fine 
```
`examples <https://docs.seldon.io/projects/alibi/en/latest/methods/ALE.html>`_
```
which could be used as a workaround.

jklaise (commented):
The sphinx build on this PR gives a lot of new "duplicate target name" warnings for `"examples"`. It seems that this is a global thing as we only use the target name once per module. We should attempt to remove these warnings, potentially an anonymous target link? @ascillitoe https://github.com/sphinx-doc/sphinx/issues/3921

This is what's also causing the issues for the links not rendering. The solution is to use anonymous links (see previous comment).

alibi/explainers/ale.py
jklaise: OK so all RST targets live in the same namespace which both causes warnings in the sphinx build and makes links not work. The solution is to either use inline links (as above) or the following anonymous target syntax:
```
        explanation
            An `Explanation` object containing the data and the metadata of the calculated ALE curves.
            See usage examples__ for details.
            
            __ https://docs.seldon.io/projects/alibi/en/latest/methods/ALE.html
```

alibi/explainers/anchor_image.py
```diff
@@ -65,7 +65,7 @@ def __init__(
         Parameters
         ----------
         predictor
-            A callable that takes a tensor of N data points as inputs and returns N outputs.
+            A callable that takes a tensor of `N` data points as inputs and returns `N` outputs.
```
jklaise, line 68: Would be good to replace `tensor` with `array` everywhere we expect numpy arrays and reserve `tensor` for torch/tensorflow.

jklaise:
To ease the manual process, it's worth looking at some automation tools that could help. In [this list](https://github.com/SeldonIO/alibi/issues/465#issue-977030735) I mention `flake8-docstrings` plugin which could help flag issues with the formatting or missing docstrings and also `darglint` which seems to provide even more functionality.

We could run these with a few options and see if either help with this PR and if so could consider integrating into our CI suite. @RobertSamoilescu @ascillitoe

alibi/explainers/anchor_image.py
```diff
@@ -65,7 +65,7 @@ def __init__(
         Parameters
         ----------
         predictor
```
jklaise, line 67: For `image_shape`, can we add that the channel axis is expected to be last?

alibi/explainers/anchor_tabular.py
```diff
@@ -580,17 +592,13 @@ def __init__(self, *args):
     def __call__(self, anchors_batch: Union[Tuple[int, tuple], List[Tuple[int, tuple]]], num_samples: int,
                  compute_labels: bool = True) -> List:
         """
-        Wrapper around TabularSampler.__call__. It allows sampling a batch of anchors in the same process,
-        which can improve performance.
+        Wrapper around :py:meth:`alibi.explainers.anchor_tabularTabularSampler.__call__`. It allows sampling a batch
```
jklaise, line 595: Take care that these ref links are working, e.g. this is missing ` .`.

alibi/explainers/anchor_tabular.py
```diff
@@ -580,17 +592,13 @@ def __init__(self, *args):
...
         Parameters
         ----------
-        anchors_batch:
-            A list of result tuples. see TabularSampler.__call__ for details.
-        num_samples:
-            See TabularSampler.__call__.
-        compute_labels
-            See TabularSampler.__call__.
+        anchors_batch, num_samples, compute_labels
+            A list of result tuples. See :py:meth:`alibi.exapliners.anchor_tabular/TabularSampler.__call__` for details.
```
jklaise, line 601: Invalid URL:
 - typo `expaliners`
 - `/` instead of `.`

alibi/explainers/anchor_tabular.py
```diff
@@ -646,18 +654,18 @@ def _get_sampler(self) -> TabularSampler:
         """
         return self.sampler
 
-    def build_lookups(self, X):
+    def build_lookups(self, X: np.ndarray):
         """
-        Wrapper around TabularSampler.build_lookups.
+        Wrapper around :py:meth:`alibi.explaines.anchor_tabular.TabularSampler.build_lookups`.
```
jklaise, line 659: Typo in URL

alibi/explainers/cem.py
```diff
@@ -40,46 +40,46 @@ def __init__(self,
         Parameters
         ----------
         predict
-            Keras or TensorFlow model or any other model's prediction function returning class probabilities
+            `Keras` or `TensorFlow` model or any other model's prediction function returning class probabilities.
```
jklaise, line 43: I would remove `Keras` as this was before `keras` and `tensorflow` were different libraries, applies to CEM, CI and CFP.

alibi/api/defaults.py
```diff
@@ -104,7 +104,8 @@
     'kwargs',
 ]
 """
-KernelShap parameters updated and return in metadata['params'].
+KernelShap parameters updated and returned in ```metadata['params']``.
```
jklaise, line 107: Triple backtick typo.

alibi/api/interfaces.py
```diff
@@ -118,14 +118,14 @@ def save(self, path: Union[str, os.PathLike]) -> None:
     def _update_metadata(self, data_dict: dict, params: bool = False) -> None:
```
jklaise, line 118: Could you add a docstring for the `reset_predictor` method please?

alibi/explainers/anchor_tabular.py
jklaise: Just a reminder that this is still outstanding.

alibi/models/pytorch/autoencoder.py
```diff
@@ -42,6 +42,19 @@ def __init__(self,
         self.to(self.device)
 
     def forward(self, x: torch.Tensor):
```
jklaise, line 44: Return type hint missing.

alibi/models/tensorflow/autoencoder.py
```diff
@@ -35,6 +35,22 @@ def __init__(self,
         self.decoder = decoder
 
     def call(self, x: tf.Tensor, **kwargs):
```
jklaise, line 37: Missing return type hint.

jklaise (changes requested):
Looks good! Just a few minor details and we can merge.

The docstring guidelines are great and will be a big help for consistent style going forward. Let's open a new PR and put them into the repo (probably `doc/README.md` makes most sense).

jklaise (approved):
Thanks!

Issue #549, opened 2021-12-04
Counterfactual instance out of predefined range for the function CounterfactualRLTabular"


NikkiRoodenrijs:
Hi!! 

For the function CounterFactualRLTabular the ranges for age are defined in such a way that the age feature cannot decrease. However, I noticed, running the exact same notebook "Counterfactual with Reinforcement Learning (CFRL) on Adult Census" (https://github.com/SeldonIO/alibi/blob/master/examples/cfrl_adult.ipynb), with the only change being that I run the counterfactual instances for more than 5 examples (see line 18/19), that for some instances the age decreased. For example in line 9 the original instance the age is 47 and for the counterfactual instance it is 48. 
 
<img width="1108" alt="Schermafbeelding 2021-12-04 om 16 36 33" src="https://user-images.githubusercontent.com/95536356/144715377-8fc8fd01-d885-46b8-9543-567b51227f6d.png">

<img width="1108" alt="Schermafbeelding 2021-12-04 om 16 36 52" src="https://user-images.githubusercontent.com/95536356/144715395-68e7e7ee-6eec-4952-b105-438ff1bd2c43.png">

Since I am not getting any errors or warnings, I think that there is a bug in the code. Can you confirm this to me or do you think there is another problem that causes the counterfactual instance to be out of the predefined range. I hope you can give me some advice as to how to cope with the generated counterfactuals being out of the predefined ranges. 

Best,
Nikki

RobertSamoilescu:
Thanks @NikkiRoodenrijs for reporting this issue. 
The problem is due to some rounding error. This is because currently we apply the constraints in the standardized feature space and then project the features back to the original input space by the reverse transformation. For integer features, we cast them back to `int` and this can cause the projected feature to differ by 1 from the original value.  

We will fix this problem shortly.

RobertSamoilescu:
@NikkiRoodenrijs, PR #550 solves the issue you mentioned.
It will be probably merged soon into the master branch, but if it is urgent, then you can use the implementation in the PR.

Thank you for reporting the issue and please let me know if this fixes the problem.

NikkiRoodenrijs:
Unfortunately, this does not solve the issue. The following error appears when I am fit the explainer. 

<img width="1170" alt="Schermafbeelding 2021-12-05 om 12 11 24" src="https://user-images.githubusercontent.com/95536356/144744236-2db80556-135a-46cf-95c2-eb4adfc97169.png">

[...]

PR #550, opened 2021-12-04
Solved counterfactual instance out of predefined range for the function CounterfactualRLTabular

RobertSamoilescu:
This PR solves the issue #549.

The problem is due to some rounding error. This is because currently we apply the constraints in the standardized feature space and then project the features back to the original input space by the reverse transformation. For integer features, we cast them back to int and this can cause the projected feature to differ by 1 from the original value.

To give a concrete example:
```python
>>> x = np.random.randint(low=0, high=100, size=10)
>>> y = 1000 * np.random.rand(10).astype(np.float32)
>>> z = (x / y).astype(np.float32)
>>> 
>>> x
array([49, 77, 78, 63, 57, 53, 87, 96, 13, 33])
>>> 
>>> z * y
array([49.      , 77.      , 78.      , 63.      , 56.999996, 53.      ,
       86.99999 , 96.      , 12.999999, 33.      ], dtype=float32)
>>> 
>>> (z * y).astype(int)
array([49, 77, 78, 63, 56, 53, 86, 96, 12, 33])
>>> 
>>> np.rint(z * y).astype(int)
array([49, 77, 78, 63, 57, 53, 87, 96, 13, 33])
>>> 
```

The only part change is the inverse pre-processing step performed by the decoder:
```python
if ft_type == int:
    np_X_inv[:, i] = np.rint(np_X_inv[:, i].astype(np.float32))
```


jklaise (approved):
LGTM.

PR #554, opened 2021-12-06
Remove nbsphinx_link

### LINKED ISSUE #553: Incorrect links in nbsphinx prolog's
The link in nbsphinx prolog's (the "this page was generated from..." messages) in methods pages are broken (but examples seem OK). For example, see below:

image

This is occuring in `latest`, but `stable` is OK. The same issue is occurring in `alibi-detect`.

ascillitoe:
This PR is an `alibi` equivalent of https://github.com/SeldonIO/alibi-detect/pull/402. 

It also fixes #553.

~~Note: I've also removed the `nbqa-pyupgrade` pre-commit hook (for now) as it incorrectly replaces `%matplotlib inline` in notebooks with code like `int  # %matplotli 0x222146EC`. This can be quite irritating as it forcibly overwrites before the commit has been finalised.~~ Reverted since newer versions don't have this issue. Thanks @MarcoGorelli!


MarcoGorelli:
Hey, sorry to drop-by, just wanted to point out that

>  it incorrectly replaces %matplotlib inline in notebooks with code like int # %matplotli 0x222146EC.

was a bug in early versions (0.3.2 is probably > 1 year old), and that the latest version 1.2.2 (and any since 1.0.0) robustly handles IPython magics.

Having said that, no offence taken if you want to remove it anyway 😄

ascillitoe:
Hi @MarcoGorelli, no problem at all! Thank you for letting us know. I've upgraded us to a shiny new version 😀

jklaise:
Seems like there's an API change in newest matplotlib which causes the notebook `path_dependent_tree_shap_adult_xgb.ipynb` to fail (the trust score notebook sometimes failing is a known issue as it depends on how many instances are found).

jklaise (approved):
Thanks!

PR #555, opened 2021-12-07
Fix prolog links

ascillitoe:
Experimenting with fix for #553


ascillitoe:
Closing as #553 will be fixed in #554

PR #558, opened 2021-12-08
Dev/assets symlink

### LINKED ISSUE #556: Example symlinks not working with local artefacts
Following the change in #554 where we moved examples to `doc/source/examples` but for convenience left symlinks in `examples` it was discovered that opening the notebooks from symlinks doesn't allow loading local artifacts such as `adult_xgb.mdl` located in `doc/source/examples` because `ipython` is running the notebook as if it was located in `examples`.

A simple solution might be to group any assets under `doc/source/examples/assets` and add another symlink to this new directory under `examples`.

jklaise:
Fixes #556.

 - [x] Check how assets under `images` are treated by the notebooks and decide if we want to put them also under `assets`.


jklaise:
As part of further investigation we realized that `examples/images` were mostly used for the main `README` which, as a result of reshuffling, is now broken. Will fix in the next commit to this PR.

Issue #586, opened 2022-02-01
Support for custom grid point input for ALE


ryanpmccaffrey:
For ALE, instead of computing grid points "under the hood" based on inputs like min_bin_points, it would be great to give users the ability to specify their own custom grid points.  For example, this could be a dictionary with keys being the features and values being the corresponding list of grid points.  This would eliminate the need to calculate quantiles or run the adaptive grid method within the explainer and instead just use the grid points directly.

jimbudarz:
Just for reference, there is a similar feature being enabled in PDP for scikit-learn that might be valuable as inspiration: https://github.com/scikit-learn/scikit-learn/pull/21033

jimbudarz:
Hi @gipster is there any other information that I can provide to help out?

jklaise:
@jimbudarz thanks for the heads-up on the PDP one! We have the PR in the works (#599) would be great to hear if you have any thoughts so we don't miss anything.

RobertSamoilescu:
There are a few cases to consider and we would appreciate some feedback regarding those. 
We will first discuss some notations. Consider `f` to be a feature of interest. We will represent possible feature values of `f` by `X` (i.e. the values from the dataset column corresponding to feature `f`), and by `O` a user-specified grid-point value.

We will illustrate some of the edge cases we've identified:
1) Grid points outside the feature value range. We can illustrate such an example as follows:  `O O O X X O X O X O O`.  We observe that there are 3 grid-points that are smaller than the minimum value in `f`, but also have 2 grid-points that are larger than the maximum value in `f`. Currently, we are clipping the grid-points values in between the minimum and maximum values of `f`. This will look something like `(O|X) X O X O (X|O)`, where `(X|O)` means that the points overlap. An alternative solution would be to keep them and report a value of 0 for ALE in that interval. Feedback is appreciated for this alternative.

2) Grid points that do not cover the entire feature range. For example: `X X O X X O X O X X X X X`. An easy fix would be to add two new grid-points corresponding to the minim and maximum value of feature `f`. This means that it would look something like: `(O|X) X O X X O X O X X X X (X|O)`. The alternative would be to plot the ALE just for the interval specified by the minimum and maximum grid point. This would be: `O X X O X O`. Feedback is appreciated for this alternative.

[...]

PR #599, opened 2022-02-24
Extend `ALE` to allow passing a per-feature user-defined grid of points at which to evaluate the ALE 

gipster:
Proposal to address https://github.com/SeldonIO/alibi/issues/586. 

Added `grid_points` argument to `explain`, which allows to pass custom grids points for each feature in a dictionary format. 
If `grid_points = None ` (default value), the present behaviour remains unchanged. 
If a dict is passed, all features names must be included as keys, and the corresponding values in the dict must be numpy array with the custom grid points or `None`


alibi/explainers/ale.py
```diff
@@ -103,6 +107,9 @@ def explain(self, X: np.ndarray, features: Optional[List[int]] = None, min_bin_p
         min_bin_points
             Minimum number of points each discretized interval should contain to ensure more precise
             ALE estimation.
+        grid_points
+            Custom grid points. Must be a dict where the keys are the names of the features and the values are
+            numpy arrays defining the grid.
```
jklaise, line 112: Would be best to add an example in the docstring. Also, we will need to update the documentation notebook `ALE.ipynb`.

jklaise: Also, please add a more specific type, i.e. `Optional[Dict[str, np.ndaray]]`. Also, I think we should refer to features by their position (so `int` instead of `str`) as `feature_names` is optional.

alibi/explainers/ale.py
```diff
@@ -143,10 +150,15 @@ def explain(self, X: np.ndarray, features: Optional[List[int]] = None, min_bin_p
 
         # TODO: use joblib to paralelise?
         for feature in features:
+            if grid_points is not None:
+                feature_grid_points = grid_points[feature_names[feature]]
```
jklaise, line 154: We need to do some validation here or wrap in `try` `except` as the user might not provide a fully specified dictionary.

Also, according to the discussion on the issue, it should be desirable to allow partially specified dictionaries and infer the points for other features.

jklaise (changes requested):
Please see the comments.

For tests, we should test on fully-specified dictionaries as well as partially specified dictionaries and check that the returned grid-points are what we expect (if they are provided).

alibi/explainers/ale.py
RobertSamoilescu: Instead of `try - except`, maybe we can do something like:
* initialize at the very beginning of the function
```python
if grid_points is None:
   grid_points = {}
```
* obtain the value through `get`
```python
feature_grid_points = grid_points.get(feature_names[feature])   # returns None by default if key not in dict
```

alibi/explainers/ale.py
```diff
@@ -383,23 +398,27 @@ def ale_num(
...
-            # ALE is 0 at a constant feature value
-            return q, np.array([[0.]]), np.array([0.])
+        # if the feature is constant, calculate the ALE on a small interval surrounding the feature value
+        if len(q) == 1:
+            if extrapolate_constant:
+                delta = max(q * extrapolate_constant_perc / 100, extrapolate_constant_min)
+                q = np.hstack((q - delta, q + delta))
+            else:
+                # ALE is 0 at a constant feature value
+                return q, np.array([[0.]]), np.array([0.])
```
RobertSamoilescu, line 418: I believe we would like to move this `if` statement outside as it can cover the case when `feature_grid_points` grid contains a singe point.

RobertSamoilescu (changes requested):

alibi/explainers/ale.py
gipster: Ok, I get the point. yes, using get should work

klalit:
Alibi Team - please provide the feedback.
FYI - @cliveseldon @alexcoca

jklaise:
@klalit what feedback are you looking for? We've completed the PR in response to #586, however the nature of the algorithm revealed that we had to make some assumptions about edge cases of user-provided grids. We've asked the original issue author @ryanpmccaffrey if they had any thoughts or extra requirements on this but they haven't come back. For the time being we're keeping this PR unmerged to see if any further feedback arises.

klalit:
@jklaise thank you for the update. Let me get in touch with the team members to provide the required information.

klalit:
@jklaise we had discussion in the team and reviewed it. It provides us the ability to specify our own custom grid points. Thank you for taking it up and implementing. We are awaiting merge closure.

doc/source/methods/ALE.ipynb
```diff
@@ -60,6 +60,16 @@
     "```python\n",
```
jklaise, line 60: <p>Move the paragraph below before the newly introduced paragraph for greater coherence?</p><p> </p><p>"This is useful if the number of total features is large and only small number is of interest. Also, it can be particularly useful to filter out categorical variable columns as there is no consistent ALE formulation and hence any results for categorical variables would be misleading."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/599/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

alibi/explainers/ale.py
```diff
@@ -103,6 +111,13 @@ def explain(self, X: np.ndarray, features: Optional[List[int]] = None, min_bin_p
         min_bin_points
             Minimum number of points each discretized interval should contain to ensure more precise
             ALE estimation.
+        grid_points
+            Custom grid points. Must be a `dict` where the keys features indices  and the values are
+            `numpy` arrays defining the grid points for each feature.
```
jklaise, line 116: Please expand the docstring:
 - "keys features indices" - grammar
 - Add constraints on what's expected for the array (i.e. monotonically increasing)
 - Expand what happens for features not defined in this dictionarry (i.e. fallback to quantiles)
 - Briefly summarize how we deal with edge cases using insights here: https://github.com/SeldonIO/alibi/issues/586#issuecomment-1055657486

alibi/explainers/ale.py
```diff
@@ -103,6 +111,13 @@ def explain(self, X: np.ndarray, features: Optional[List[int]] = None, min_bin_p
...
+
+        tol
+            Tolerance to delete consecutive quantiles obtained after clipping their values between the minimum and
+            maximum feature values. Only used when `feature_grid_points` is not ``None``.
```
jklaise, line 120: `feature_grid_points` -> `grid_points`

jklaise: I'm not 100% sure what this does from the description (quantiles + grid_points + clipping?), would be great to clarify.

alibi/explainers/ale.py
```diff
@@ -401,14 +417,11 @@ def ale_num(
...
         Refer to :class:`ALE` documentation.
-    tol
-        Tolerance to delete consecutive quantiles obtained after clipping their values between the minimum and maximum
-        feature values. Only used when `feature_grid_points` is not ``None``.
 
     Returns
     -------
-    q
-        Array of quantiles of the input values.
+    fvals
```
RobertSamoilescu, line 423: renamed it since `q` was standing for quantiles which is no longer the case

alibi/explainers/ale.py
```diff
@@ -112,12 +111,11 @@ def explain(self,
             Minimum number of points each discretized interval should contain to ensure more precise
             ALE estimation.
         grid_points
```
jklaise, line 113: I just realized that perhaps we should add to the various parameters in the constructor docstring controlling the quantile selection behaviour that they are only relevant if `grid_points` is `None` (well it's actually relevant feature-wise...)? Or perhaps a better way to add to the `grid_points` docstring that for the features for which `grid_points` are passed the relevant parameters passed in the constructor are not considered?

alibi/explainers/tests/test_ale.py
```diff
@@ -117,6 +117,11 @@ def test_explain(mock_ale_explainer, features, input_dim, batch_size, custom_gri
     if not custom_grid:
         for alev, featv in zip(exp.ale_values, exp.feature_values):
             assert alev.shape == (featv.shape[0], out_dim)
+    else:
+        for i, f in enumerate(grid_points.keys()):
+            # need to remove the first and last element just in case the feature values are extended
+            # with the min & max feature value
+            assert np.all(np.isin(exp.feature_values[i][1:-1], grid_points[f]))
```
jklaise, line 124: I'm not sure I follow the logic this test branch, the original branch was intended to check that `exp.ale_values` and `exp.feature_values` match in dimension so can be plotted, whereas this branch now considers the input `grid_points` rather than the output contained in `exp`?

jklaise: Perhaps we need 2 tests?
 - Check that `exp.ale_values` and `exp.feature_values` always match (so can be plotted)
 - In the case that `grid_points` is passed, check that the ultimate `exp.feature_values` match with the original `grid_points` (up to potentially adding the endpoints - however, would this not fail if there are gridpoints passed that don't contain any feature values?)

jklaise (approved):

jklaise:
Thanks @RobertSamoilescu and @gipster!

Issue #601, opened 2022-02-25
Python 3.10 support


jklaise:
The newly released `tensorflow` 2.8 has Python 3.10 support. We can proceed with adding CI tests for 3.10 after merging #588.

ascillitoe:
This is currently blocked by https://github.com/pytorch/pytorch/issues/66424.

jklaise:
Closed via #605.

PR #605, opened 2022-03-09
Add Python 3.10

ascillitoe:
Addresses https://github.com/SeldonIO/alibi/issues/601


ascillitoe:
This is currently blocked by https://github.com/pytorch/pytorch/issues/66424.

jklaise:
I guess technically `torch` is optional so we could say we support Python 3.10, but practically it wouldn't work and it would be a mess to get our CI working without the `torch` components. So probably best to wait.

ascillitoe:
According to the PyTorch changelog (https://github.com/pytorch/pytorch/releases) the latest pytorch version (1.11) now supports Python 3.10 (although https://github.com/pytorch/pytorch/issues/66424 suggests they won't have CI for it until 1.12). Given that `torch` is optional we're probably good to merge this in once the `alibi-testing` related CI error is resolved.

ascillitoe:
Looks like the latest build failure down to the `typed-ast` package, which doesn't play nicely with Python 3.10 on Windows. Without doing a full deps analysis, I suspect `typed-ast` might be being pulled in by `mypy`. See https://github.com/python/typed_ast/issues/156#issuecomment-792313554 and https://github.com/python/mypy/issues/10124. This might be a good time to move to `mypy>=0.900`, which drops `typed-ast` entirely (see [here](http://mypy-lang.blogspot.com/2021/06/mypy-0900-released.html)). Although we'll have to deal with the 3rd party library stubs change...

We might also just be able to enforce a newer `typed-ast` version (https://github.com/python/typed_ast/pull/155 suggests the issue was recently fixed), although it still looks flaky to me...

ascillitoe:
The typo `explainer._check_interactions.asert_not_called()` in `test_shap_wrappers.py` caused the following error in Python 3.10:

`AttributeError: 'asert_not_called' is not a valid assertion. Use a spec for the mock if 'asert_not_called' is meant to be an attribute.`

`asert` has now been fixed to `assert`.

README.md
```diff
@@ -5,7 +5,7 @@
 <!--- BADGES: START --->
 
 <!---
-![Python version](https://img.shields.io/badge/python-3.7%20%7C%203.8%20%7C%203.9-blue.svg)
+![Python version](https://img.shields.io/badge/python-3.7%20%7C%203.8%20%7C%203.9%20%7C%203.10-blue.svg)
```
jklaise, line 8: Will need a rebase which should throw up another merge conflict as this has been automated on `master`.

jklaise (approved):

README.md
ascillitoe: This is because I screwed up. The above two lines are fenced inside comments, so the above change did nothing. 

I actually updated the correct line (line 14) to use the automated versioning yesterday (https://github.com/SeldonIO/alibi/commit/5efc04b4963a39082e73f3b3a091133162494d0e), so there is no need to update this manually at all now...

Will delete the confusing commented lines in `master` and rebase.

PR #611, opened 2022-03-18
Fix setup.py licence classifier

ascillitoe:
Fix the licence classifier in setup.py, to match Alibi Detect. `"License :: OSI Approved :: Apache 2.0 License"` is not a valid classifier, and should be `"License :: OSI Approved :: Apache Software License"`.

@jklaise @mauicv  just to flag, we're having to specify the less specific "Apache Software Licence" since there are no version specific classifiers. This appears to be a long-running issue (https://github.com/pypa/pypi-legacy/issues/91, https://github.com/pypa/trove-classifiers/issues/17). I doubt this is a problem as take the view what's in `LICENCE` is our formal licence, and the classifier is a lossy mechanism to indicate the broad family of licence. But flagging here just in case...


ascillitoe:
In fact, it appears on PyPi as: 
image

So the full 2.0 is indicated. The classifier just adds the tag at the bottom which can be used as a search filter on PyPI...

PR #624, opened 2022-04-06
Replace all occurences of "en/latest" with "en/stable"

jklaise:
This is to ensure that our docs always point to the stable release
instead of latest to avoid potential discrepancy between someone reading
the latest docs but installing a stable version.


jklaise:
Same comments apply to this PR as here: https://github.com/SeldonIO/alibi-detect/pull/479#issuecomment-1090274958

README.md
```diff
@@ -121,7 +121,7 @@ Alibi can be installed from:
 
 ### With conda 
 
-To install from [conda-forge](https://conda-forge.org/) it is recommended to use [mamba](https://mamba.readthedocs.io/en/latest/), 
+To install from [conda-forge](https://conda-forge.org/) it is recommended to use [mamba](https://mamba.readthedocs.io/en/stable/),
```
mauicv, line 124: is this intentional. It's a `mamba` link not `alibi`?

mauicv (approved):
There are a lot of places where we've changed other library links to the stable docs but I assume that's intentional.

LGTM

jklaise:
Yup that was intentional both for consistency and because users will almost always run `stable` versions of everything.

PR #626, opened 2022-04-06
Implement `ProtoSelect` dataset summarization method.

RobertSamoilescu:
This PR contains an implementation of the ProtoSelect algorithm for dataset distilation/prototype selection. The implementation is backend agnostic an uses only `numpy`. The PR implements 2 types of kernel distances (`GaussainRBF` & `EuclideanDistance`) which computes the (di)similarity in batches.  In addition, a cross-validation hyperparameter selection is implemented for  `EuclideanDistance` kernel disimilarity based on the accuracy of a 1-KNN classifier. Notebook examples are provided for 2 data modalities: tabular and images. For images, the prototypes and their importance can be visualized through a 2D image scatter plot.

Potential follow-ups:
1. Cross-validation for GaussianRBF - potentially find a way to reuse the CV implementation for Euclidean distance (e.g. unify them)
2. Improve visualizations
3. Saving and loading functionality - currently not supported. I believe that would mean to store the kernel matrix, but I can see pros & cons to that


doc/source/index.md
```diff
@@ -33,6 +33,7 @@ methods/KernelSHAP.ipynb
 methods/LinearityMeasure.ipynb
 methods/TrustScores.ipynb
 methods/TreeSHAP.ipynb
+methods/ProtoSelect.ipynb
```
jklaise, line 36: Let's put the new entry so the alphabetical order is preserved.

doc/source/examples/protoselect_adult_cifar10.ipynb
```diff
@@ -0,0 +1,1140 @@
+{
```
jklaise, line 1: <ol><li>Fix the output warning by installing <code>IProgress</code> ?</li><li>Do we need to place <code>apply_category_mapping</code> somewhere else as the import path is confusing for this notebook (nothing to do with <code>cfrl</code>)?</li><li>If <code>cv_protoselect_euclidean</code> and <code>visualize_prototypes</code> are to be puvlic they should be exposed in <code>alibi.prototypes</code> (although let's discuss after I've seen the code as they might be specific to <code>ProtoSelect</code>?   </li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/626/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

jklaise, line 1: <pre class="ql-syntax" spellcheck="false">Line #15.    # fit data perprocessor
</pre><p>"perprocessor" -&gt; "preprocessor"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/626/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

jklaise, line 1: <pre class="ql-syntax" spellcheck="false">Line #5.    # search for the best epslion-radius value
</pre><p>"epslion" -&gt; "epsilon"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/626/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

jklaise, line 1: <p>"and whom have a partner" -&gt; "and who have a partner"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/626/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

doc/source/methods/ProtoSelect.ipynb
```diff
@@ -0,0 +1,269 @@
+{
```
jklaise, line 1: <ol><li>Protoselect -&gt; ProtoSelect (check spelling consistent across materials).</li><li>Perhaps we should stress at the beginning that the whole discussion and method applies to classification problems.</li><li>Nitpick - can we have an "argmin" macro with the index l centered below? Not sure if possible in Jupyter Latex... (https://tex.stackexchange.com/questions/5223/command-for-argmin-or-argmax)</li><li>Please link to the Arxiv landing page for the paper, not directly to the pdf (see below too)</li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/626/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

jklaise, line 1: <ol><li>Should we be consistent with the set inclusion symbol (e.g. paragraph above uses \subseteq whereas here we have \subset</li><li>centered in -&gt; centered on</li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/626/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

jklaise, line 1: <ol><li>The bullet list at the start isn't rendered properly (might need a space before the preceding sentence: https://github.com/SeldonIO/alibi/tree/master/doc#gotchas-when-writing-notebooks-as-examples)</li><li>the 3rd condition (is sparse) is confusing - a desirable prototype is 1 point but we talk here about it containing things (so the set P_l ?)</li><li>"integer program" - add link to wiki page? (https://en.wikipedia.org/wiki/Integer_programming)</li><li>Nitpick - are the labels (a) and (b) for the constraints necessary since we only refer to them as "first" and "second" constraint respectively? We could change labels to be (1) and (2) or change the text to refer to (a) and (b) for increased clarity. Also perhaps worth checking in we can add a space or right-justify the labels in the equation so they stand apart from the actual inequalities.</li><li>"can not" -&gt; "cannot" is better style :)</li><li>"proposed two alternative solution" -&gt; "propose two alternative solutions" (generally use present tense)</li><li>"&nbsp;The first one consist in a" -&gt; "The first one consists of" and "and transforming" -&gt; "and transformation of"</li><li>Please link to the Arxiv landing page for the paper, not directly to the pdf</li><li>Nitpick - I find it easier to read if not introducing terms like "state", e.g. "Given the current choice of prototype subsets (P_1, ..., P_l), in the next iteration we update it to (P_1, ...)"</li><li>Should we use display math for the Obj(x_j, l) ?</li><li>"given that" -&gt; "where" ? (since they are not constraints but definitions)</li><li>Same comments around labels (a) and (b) as above for the definitions of ksi, nu</li><li>Nitpick - perhaps using \lvert and \revert instead of \mid could give better typography</li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/626/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

jklaise, line 1: <ol><li>Since the method is technically not an explainer, I would rename the object variable to something else (same in example notebook).</li><li>At the top link to the API docs for a full description of all parameters?</li><li>[PLACHEHOLDER - to discuss last paragraph]</li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/626/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

jklaise, line 1: <ol><li>Import missing (useful for completeness to show where the function comes from)</li><li>Link to API docs for a full description</li><li>"meta-data" -&gt; "metadata"</li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/626/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jklaise, line 1: <ol><li>Is it worth spelling out the full type signature expected of <code>preprocess_fn</code>?</li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/626/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

jklaise, line 1: <ol><li>"log" -&gt; "logarithm" (I find it easier to read if it's spelled out in written form)</li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/626/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

jklaise, line 1: <ol><li>Link to API docs for a full description.</li><li>"reducer operated" -&gt; "reducer operates"</li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/626/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

jklaise, line 1: <ol><li>Link not working?</li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/626/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

alibi/prototypes/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""
+The 'alibi.prototypes' modules includes prototypes and criticism selection methods.
+"""
+
+from .protoselect import ProtoSelect
```
jklaise, line 5: TBD - perhaps export the public utility functions here as well (unless they're quite specific to `ProtoSelect`)?

alibi/api/defaults.py
```diff
@@ -249,3 +249,20 @@
 """
 Default CounterfactualRL data.
 """
+
+
+DEFAULT_META_PROTOSELECT = {"name": None,
+                            "type": ["prototypes"],
+                            "explanation": ["global"],
```
jklaise, line 256: These aren't really consistent/required because the method is not an explainer, let's discuss.

testing/test_notebooks.py
```diff
@@ -32,6 +32,7 @@
     'kernel_shap_adult_lr.ipynb',  # slow to explain 128 instances
     'xgboost_model_fitting_adult.ipynb',  # very expensive hyperparameter tuning
     'integrated_gradients_transformers.ipynb',  # forward pass through BERT to get embeddings is very slow
+    'protoselect_adult_cifar10.ipynb',  # wget cifar10 model from google cloud
```
jklaise, line 35: Why is this a reason not to test? Fetching the model should be quick enough.

alibi/saving.py
```diff
@@ -373,6 +376,14 @@ def _load_CounterfactualRLTabular(path: Union[str, os.PathLike],
     return _helper_load_CounterfactualRL(path, predictor, explainer)
 
 
+def _save_ProtoSelect(path: Union[str, os.PathLike]) -> None:
+    raise NotImplementedError('ProtoSelect saving functionality not implemented.')
```
jklaise, line 380: Would be good to expand on the pros/cons for saving the kernel matrix etc.

alibi/utils/kernel.py
```diff
@@ -0,0 +1,131 @@
+import numpy as np
+from typing import Union, Optional
+from alibi.utils.distance import squared_pairwise_distance
+
+
+class GaussianRBF:
+    def __init__(self, sigma: Optional[Union[float, np.ndarray]] = None, **kwargs) -> None:
```
jklaise, line 7: `kwargs` are unused in all three classes in this module, can we remove them?

alibi/utils/distance.py
```diff
@@ -243,3 +242,78 @@ def multidim_scaling(d_pair: dict,
         feature_range = new_feature_range
 
     return d_abs_scaled, feature_range
+
+
+def squared_pairwise_distance(x: np.ndarray, y: np.ndarray, a_min: float = 1e-30, a_max: float = 1e30) -> np.ndarray:
```
jklaise, line 247: What's the motivation and choice behind `a_min` and `a_max`? I assume `a_min` is to avoid division by zero, but does the specific value of `1e-30` have some significance (e.g. machine precision for `float64` is given by `np.finfo(float).eps` which is `~2.2e-16`).

Similar for `a_max`, are we worried about overflow in downstream applications?

alibi/prototypes/tests/test_utils.py
```diff
@@ -0,0 +1,103 @@
...
+from alibi.utils.kernel import GaussianRBF, GaussianRBFDistance, EuclideanDistance
+
+
+@pytest.mark.parametrize('size', [100])
+@pytest.mark.parametrize('ft_size', [2, 5, 10, 100])
+def test_squared_pairwise_distance(size, ft_size):
+    """
+    Test vectorize implementation of the squared pairwise distance
+    """
+    np.random.seed(0)
```
jklaise, line 13: Not sure we need to set the seed here and in other functions as it should work for any random array?

alibi/utils/kernel.py
```diff
@@ -0,0 +1,131 @@
...
+        ----------
+        sigma
+            Kernel bandwidth. Need not to be specified if being inferred or trained.
+            Can pass multiple values to eval kernel with and then average.
+        **kwargs
+            Other arguments. Not used.
+        """
+        super().__init__()
+        self.config = {'sigma': sigma}
+        self.min_val, self.max_val = 0., 1.   # minimum and maximum values that the kernel can take.
```
jklaise, line 23: Are these used anywhere (apart from the `GaussianRBFDistance`, see below)? If not, not sure I see a reason for these?

alibi/utils/distance.py
```diff
@@ -243,3 +242,78 @@ def multidim_scaling(d_pair: dict,
...
+        Batch size to be used for each prediction.
+    preprocess_fn
+        Optional preprocessing function for each batch.
+
+    Returns
+    -------
+    Kernel matrix in the form of a `numpy` array.
+    """
+    if type(x) != type(y):
+        raise ValueError("x and y should be of the same type")
```
jklaise, line 299: Do we expect this to realistically arise in practice given that this is a function primarily for internal use and that the type hints would only accept `x` and `y` to be `np.ndarray`?

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+
+from alibi.utils.distance import batch_compute_kernel_matrix
+from alibi.api.interfaces import Explainer, Explanation, FitMixin
+from alibi.api.defaults import DEFAULT_META_PROTOSELECT, DEFAULT_DATA_PROTOSELECT
+from alibi.utils.kernel import EuclideanDistance
+
+logger = logging.getLogger(__name__)
+
+
+class ProtoSelect(Explainer, FitMixin):
```
jklaise, line 22: Let's discuss if this should even follow the `Explainer` interface...

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+    def __init__(self,
+                 eps: float,
+                 kernel_distance: Callable[[np.ndarray, np.ndarray], np.ndarray],
+                 lbd: Optional[float] = None,
+                 batch_size: int = int(1e10),
+                 preprocess_fn: Optional[Callable[[np.ndarray], np.ndarray]] = None,
+                 verbose: bool = False,
+                 **kwargs):
```
jklaise, line 30: `kwargs` unused

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+        """
+        Constructor
```
jklaise, line 32: Please add a few sentences about the method here as this is the primary dosctring the users would see.

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+        kernel_distance
+            Kernel distance to be used. Expected to support computation in batches.
+        eps
+            Epsilon ball size.
+        lbd
+            Penalty for each prototype. Encourages a lower number of prototypes to be selected.
+        batch_size
+            Batch size to be used for kernel matrix computation.
+        preprocess_fn
+            Preprocessing function for kernel matrix computation.
```
jklaise, line 45: Would like to see more discussion about the use cases here, perhaps also link to the Method description?

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+                 preprocess_fn: Optional[Callable[[np.ndarray], np.ndarray]] = None,
+                 verbose: bool = False,
+                 **kwargs):
+        """
+        Constructor
+
+        Parameters
+        ----------
+        kernel_distance
+            Kernel distance to be used. Expected to support computation in batches.
```
jklaise, line 37: Probably need to expand what the expectation means exactly in terms of the `kernel_distance` interface beyond the provided type hint.

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+            'eps': eps,
+            'lbd': lbd,
+            'batch_size': batch_size,
+            'verbose': verbose
+        })
+
+    def fit(self,
+            X: np.ndarray,
+            X_labels: Optional[np.ndarray] = None,
+            Y: Optional[np.ndarray] = None) -> 'ProtoSelect':
```
jklaise, line 77: Let's discuss parameter naming here as usually `y` or `Y` plays the role of labels for `X`.

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+        -------
+        self
+            Reference to itself.
+        """
+        self.X = X
+        # if the `X_labels` are not provided, then consider that all elements belong to the same class. This means
+        # that loss term which tries to avoid including in an epsilon ball elements belonging to other classes
+        # will always be 0. Still the first term of the loss tries to cover as many examples as possible with
+        # minimal overlap between the epsilon balls corresponding to the other prototypes.
+        self.X_labels = X_labels.astype(np.int32) if (X_labels is not None) else np.zeros((len(X), ), dtype=np.int32)
```
jklaise, line 101: Perhaps we should validate first that the shape/length of `X_labels` is what we expect, given the shape/length of `X`.

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+        """
+        Constructor
+
+        Parameters
+        ----------
+        kernel_distance
+            Kernel distance to be used. Expected to support computation in batches.
+        eps
+            Epsilon ball size.
+        lbd
```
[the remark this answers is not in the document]
jklaise: Also describe the default behaviour if `None`.

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+        self.max_label = np.max(self.X_labels)
+        self.kmatrix_yx = batch_compute_kernel_matrix(x=self.Y,
+                                                      y=self.X,
+                                                      kernel=self.kernel_distance,
+                                                      batch_size=self.batch_size,
+                                                      preprocess_fn=self.preprocess_fn)
+
+        return self
+
+    def explain(self, num_prototypes: int = 1) -> Explanation:
```
jklaise, line 118: It's a bit of a weird interface for this method now that I think of it. Ideally since `X` is your reference and `Y` is your dataset of interest, you'd want `X` to go into `fit` and `Y` into `explain`, but fitting actually requires both `X` and `Y`. Still, the current pattern probably makes most sense.

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+            'batch_size': batch_size,
+            'verbose': verbose
+        })
+
+    def fit(self,
+            X: np.ndarray,
+            X_labels: Optional[np.ndarray] = None,
+            Y: Optional[np.ndarray] = None) -> 'ProtoSelect':
+        """
+        Fit the explainer by setting the reference dataset.
```
jklaise, line 79: Perhaps we should be explicit that this step forms the kernel matrix in memory and what the implied space complexity of that is (don't think there is intermediate memory allocated beyond the `len(x) x len(y)`?

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+
+        Parameters
+        ----------
+        num_prototypes
+            Number of maximum prototypes to be selected.
+
+        Returns
+        -------
+        An `Explanation` object containing the prototypes, prototypes indices and protoypes labels with additional \
+        metadata as attributes
```
jklaise, line 132: Minor grammar improvements - > "containing the prototypes, prototype indices and prototype labels..."

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+        """
+        if num_prototypes > len(self.Y):
+            num_prototypes = len(self.Y)
+            logger.warning('The number of prototypes requested is larger than the number of elements from '
+                           f'the prototypes selection set. Automatically setting `num_prototypes={num_prototypes}`.')
+
+        # dictionary of prototypes indices for each class
+        protos = {l: [] for l in range(self.max_label + 1)}  # type: Dict[int, List[int]]  # noqa: E741
```
jklaise, line 140: It seems that the implicit assumption we're making here is that labels are always represented as integers in `[0, ..., n-1]` and all labels are represented? Should we make this explicit in the description of `X_labels`?

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+        # other classes different then `l` will be covered if we add the prototype `y`.
+        delta_nu_summed = np.sum(delta_nu_all, axis=-1)
+        # compute the tradeoff score - each prototype tries to cover as many new elements as possible
+        # belonging to the same class, while trying to avoid covering elements belonging to another class
+        scores_all = delta_xi_summed - delta_nu_summed - self.lbd
+
+        # add progressing bar if `verbose=True`
+        generator = range(num_prototypes)
+        if self.verbose:
+            generator = tqdm(generator)
```
jklaise, line 175: Could it be cleaner to use the `disable` parameter to `tqdm` like Alex did in #628? I.e. `generator = tqdm(generator, disable=not verbose)` ?

alibi/prototypes/tests/test_protoselect.py
```diff
@@ -0,0 +1,100 @@
...
+from alibi.prototypes.protoselect import cv_protoselect_euclidean
+
+
+@pytest.mark.parametrize('n_classes', [2, 3, 5, 10])
+@pytest.mark.parametrize('ft_factor', [1, 2, 3, 10])
+@pytest.mark.parametrize('kernel_distance', [EuclideanDistance()])
+@pytest.mark.parametrize('num_prototypes', [30, 40, 50, 100])
+@pytest.mark.parametrize('eps', [0.2, 0.5])
+def test_multimodal(n_classes, ft_factor, kernel_distance, num_prototypes, eps):
+    """ Integration test. """
```
jklaise, line 17: Can you clarify why this is called `test_multimodal` ?

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+
+    @classmethod
+    def load(cls, path: Union[str, os.PathLike]) -> "Explainer":  # type: ignore[override]
+        return super().load(path, predictor=None)
+
+
+def _helper_protoselect_euclidean_1knn(explainer: ProtoSelect,
+                                       num_prototypes: int,
+                                       eps: float,
+                                       preprocess_fn: Optional[Callable[[np.ndarray], np.ndarray]] = None
```
jklaise, line 225: `preprocess_fn` is currently unused

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+        validation labels. 1-KNN classifier is evaluated on the validation dataset to obtain the best epsilon radius.
+        In case ``valset=None``, then `n-splits` cross-validation is performed on the `refset`.
+    num_prototypes
+        The number of prototypes to be selected.
+    eps_range
+        Optional ranges of values to select the epsilon radius from. If not specified, the search range is
+        automatically proposed based on the inter-distances between `X_ref` and `X_proto`. The distances are filtered
+        by considering only values in between the `quantiles` values. The minimum and maximum distance values are
+        used to define the range of values to search the epsilon radius. The interval is discretized in `grid_size`
+        equal-distant bins.
```
jklaise, line 293: Not sure exactly what `eps_range` should look like. I assume it's not a range of 2 values otherwise it would be of type `Tuple[float, float]`. I see later in code it's actually a `linspace` so it is a set of discrete epsilon values, perhaps need to rephrase the description here?

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+    -------
+    Dictionary containing
+     - ``'best_eps'``: ``float`` - best epsilon radius according to the accuracy of a 1-KNN classifier.
+     - ``'meta'``: ``dict`` - dictionary containing argument and data gather throughout cross-validation.
+    """
+    X_ref, X_ref_labels = refset
+    X_proto = protoset[0]
+    if preprocess_fn is not None:
+        X_ref = _batch_preprocessing(X=X_ref, preprocess_fn=preprocess_fn, batch_size=batch_size)
+        X_proto = _batch_preprocessing(X=X_proto, preprocess_fn=preprocess_fn, batch_size=batch_size)
```
jklaise, line 317: Curious why we have to do it in batches?

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+                             preprocess_fn: Optional[Callable[[np.ndarray], np.ndarray]] = None,
+                             **kwargs) -> dict:
+    """
+    Cross-validation parameter selection for ProtoSelect with Euclidean distance. The method computes
+    the best epsilon radius.
+
+    Parameters
+    ----------
+    refset
+        Tuple, `(X_ref, X_ref_labels)`, consisting of the reference data instances with the corresponding reference
```
jklaise, line 277: To discuss - consistent naming between this and `ProtoSelect.fit` where it's `X` and `X_labels`.

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+    # train 1-knn classifier
+    proto, proto_labels = explanation.data['prototypes'], explanation.data['prototypes_labels']
+    if len(proto) == 0:
+        return None
+
+    knn = KNeighborsClassifier(n_neighbors=1)
+    return knn.fit(X=proto, y=proto_labels)
+
+
+def cv_protoselect_euclidean(refset: Tuple[np.ndarray, np.ndarray],
```
jklaise, line 259: TBD - if possible to make this function more generic and take in a kernel so we could reuse it with `GaussianRBF` etc.

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+                             protoset: Tuple[np.ndarray, ],
```
jklaise, line 260: Could this be `Optional` and if `None` then `X_ref` is used, same as in the `ProtoSelect` implementation?

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+        X_proto = _batch_preprocessing(X=X_proto, preprocess_fn=preprocess_fn, batch_size=batch_size)
+
+    # propose eps_range if not specified
+    if eps_range is None:
+        dist = batch_compute_kernel_matrix(x=X_ref, y=X_proto, kernel=EuclideanDistance()).reshape(-1)
+        if quantiles is not None:
+            if quantiles[0] > quantiles[1]:
+                raise ValueError('The quantile lower-bound is greater then the quantile upper-bound.')
+
+            quantiles = np.clip(quantiles, a_min=0, a_max=1)
```
jklaise, line 326: From this and the `quantiles` description it's not quite clear what is an allowed range? Probably just need to add in the description that the range of the quantiles is in `[0, 1]` and clipped to these values if outside.

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+    for i in range(len(imgs)):
+        x0, y0, im = x[i], y[i], imgs[i]
+        ab = AnnotationBbox(im, (x0, y0), xycoords='data', frameon=False)
+        artists.append(ax.add_artist(ab))
+
+    ax.update_datalim(np.column_stack([x, y]))
+    ax.autoscale()
+
+
+def visualize_prototypes(explanation: 'Explanation',
```
jklaise, line 475: Perhaps this needs a slightly different name as it's specific to image datasets?

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+                         zoom_ub: float = 3.0):
+    """
+    Plot the images of the prototypes at the location given by the `reducer` representation.
+    The size of each prototype is proportional to the log of the number of correct-class training images covered
+    by that prototype (Bien and Tibshiran (2012): https://arxiv.org/abs/1202.5933).
+
+    Parameters
+    ----------
+    explanation
+        Explanation object.
```
jklaise, line 491: Not just any `Explanation` object, but one that is of type `ProtoSelect`. Also ties in with discussion above whether this method should be an explainer or not.

alibi/prototypes/protoselect.py
jklaise: Perhaps we could go with (as used in other functions) with:
`X` -> `X_ref`
`X_labels` -> `y_ref`
`Y` -> `X_test` or `X_proto` or `X`

Alternatively, if we think by far the most common use case would be to just have one dataset, we cold do:
`X` -> `X`
`X_labels` -> `y`
`Y` -> `X_test` or `X_proto`

Also compare with `alibi-detect` where we have `x_ref` and `x`.

alibi/prototypes/protoselect.py
jklaise, line 475: If relevant, consider if we can take some implementation details from `plot_ale`: https://github.com/SeldonIO/alibi/blob/61f0c6afd7e5e23585fc3661b4a4aa137db9abb7/alibi/explainers/ale.py#L455

E.g. we might want to return the axes for further customization by the user if desired?

alibi/prototypes/protoselect.py
```diff
@@ -0,0 +1,543 @@
...
+    refset
+        Tuple, `(X_ref, X_ref_labels)`, consisting of the reference data instances with the corresponding reference
+        labels.
+    reducer
+        2D reducer. Reduces the input feature representation to 2D. Note that the reducer operated directly on the
```
jklaise, line 496: operated -> operates

alibi/prototypes/protoselect.py
jklaise: Or be more consistent with arguments affecting the figure/axes, e.g. for `plot_ale` I exposed `fig_kw` dictionary which allows users greater control. Passing a figsize then could be done as inhttps://docs.seldon.io/projects/alibi/en/latest/examples/ale_regression_boston.html#ALE-for-the-linear-regression-model

jklaise (changes requested):
Thanks Robert, I think most of these are small comments around the discussion in the method docs and examples with a few more mostly around conventions in the code. Let's discuss when you're back.

doc/source/methods/ProtoSelect.ipynb
RobertSamoilescu: <ol><li><br></li><li> You are correct. The properties refer to a prototype set. I updated.</li></ol>

RobertSamoilescu: <ol><li>Need to discuss what would be a proper naming</li><li>All the parameters are already presented</li><li>...</li></ol>

alibi/prototypes/__init__.py
RobertSamoilescu: They are specific to `ProtoSelect`. One is for cross-validation an the other is for visualization. Better to discuss whether it makes sens to export them.

testing/test_notebooks.py
RobertSamoilescu: It's not. Forgot to remove it after I tested something. Some tests were failing and complaining about `wget`.

alibi/saving.py
RobertSamoilescu: The only pros will be that we can store some hyperparameters from cross-validation. The bad thing is that is quadratic in the number of elements and storing the matrix can be a bit demanding ... an alternative will be to store only the params but not the kernel matrices. This would mean that it requires refitting and I guess this is connected with another comment related to the API. Better discuss it offline ...

alibi/utils/distance.py
RobertSamoilescu: Good point ... I actually took them from [detect](https://github.com/SeldonIO/alibi-detect/blob/a0e7082ebb12206fa7f5417cab0353c55169cfa3/alibi_detect/utils/tensorflow/distance.py#L9) and I didn't realize

alibi/utils/kernel.py
```diff
@@ -0,0 +1,131 @@
...
+            See :py:meth:`alibi.utils.kernel.GaussianRBF.__init__`.
+        **kwargs
+            Other arguments. Not used.
+        """
+        super().__init__()
+        self.kernel = GaussianRBF(sigma=sigma)
+
+    def __call__(self, x: np.ndarray, y: np.ndarray, infer_sigma: bool = False) -> np.ndarray:
+        kmatrix = self.kernel(x, y, infer_sigma)
+        return self.kernel.max_val - kmatrix
```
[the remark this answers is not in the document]
RobertSamoilescu: Not really. I removed it.

alibi/prototypes/protoselect.py
RobertSamoilescu: I believe it can be reused. The only problem is that different kernels might have different hyperparameters to be tuned. For example, for GaussianRBF, one should also search for the standard deviation. I was thinking to leave it for another iteration.

alibi/prototypes/protoselect.py
RobertSamoilescu: One can use large models (e.g. transformers) to extract a feature representation, for which you need to do the preprocessing in batches.

alibi/utils/kernel.py
```diff
@@ -0,0 +1,131 @@
...
+from typing import Union, Optional
+from alibi.utils.distance import squared_pairwise_distance
+
+
+class GaussianRBF:
+    def __init__(self, sigma: Optional[Union[float, np.ndarray]] = None, **kwargs) -> None:
+        """
+        Gaussian RBF kernel: :math:`k(x,y) = \\exp(-\\frac{||x-y||^2}{2\\sigma^2})`.
+        A forward pass takes a batch of instances `x` of size `Nx x features` and `y` of size `Ny x features`
+        and returns the kernel matrix of size `Nx x Ny`.
```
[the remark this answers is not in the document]
RobertSamoilescu: That's correct. I updated the docs

alibi/prototypes/tests/test_utils.py
RobertSamoilescu: It's not needed.

alibi/utils/distance.py
RobertSamoilescu: also inherited from [detect](https://github.com/SeldonIO/alibi-detect/blob/a0e7082ebb12206fa7f5417cab0353c55169cfa3/alibi_detect/utils/tensorflow/distance.py#L62-L63)

alibi/prototypes/protoselect.py
RobertSamoilescu: The assumption of integer representation is correct, but I think this is quite standard. 
On the other hand, we do not need all labels to be represented. It is true that there is some memory allocation that considers all the labels (I believe might be improved by adding an intermediate mapping and allocate the memory only for the one available), but we definitely do not assume that all labels from `0` to `n-1` are available. Might be the case that only `0, 1 & n-1` are there, which works just fine.

alibi/prototypes/tests/test_protoselect.py
RobertSamoilescu: Supposed to be something else initially and I forgot to rename it.

alibi/prototypes/protoselect.py
RobertSamoilescu: Forgot to remove it after removing some unnecessary re-computation.

alibi/prototypes/protoselect.py
RobertSamoilescu: `visualize_image_prototypes` ?

alibi/prototypes/protoselect.py
RobertSamoilescu: Added clarification, but still need to discuss

alibi/prototypes/protoselect.py
RobertSamoilescu: I've renamed them to be consistent with the suggestions for the fit function.

alibi/prototypes/protoselect.py
RobertSamoilescu: Yes. It can be optional.

alibi/prototypes/protoselect.py
RobertSamoilescu: Made some changes.

doc/source/examples/protoselect_adult_cifar10.ipynb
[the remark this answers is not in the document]
RobertSamoilescu: added a few sentences of ease of read

doc/source/methods/CFRL.ipynb
```diff
@@ -557,7 +557,7 @@
    "name": "python",
```
jklaise, line 557: Minor, but please keep an eye and try not to commit notebooks that have changed locally but should not be part of the PR (this one and many others below).

testing/test_notebooks.py
jklaise: To be excluded on Windows due to use of `wget`.

alibi/prototypes/protoselect.py
```diff
@@ -21,35 +21,41 @@
...
+                 lambda_penalty: Optional[float] = None,
                  batch_size: int = int(1e10),
-                 preprocess_fn: Optional[Callable[[np.ndarray], np.ndarray]] = None,
-                 verbose: bool = False,
-                 **kwargs):
+                 preprocess_fn: Optional[Callable[[Union[list, np.ndarray]], np.ndarray]] = None,
+                 verbose: bool = False):
         """
-        Constructor
+        Prototype selection for dataset distillation and interpretable classification.
```
jklaise, line 31: Add link to the paper too?

alibi/prototypes/__init__.py
RobertSamoilescu: just expose `visualize_prototypes`

alibi/api/defaults.py
RobertSamoilescu: add some data terminology?

alibi/prototypes/protoselect.py
RobertSamoilescu: check what happens if the labels do not follow `[0, L-1]` -> memory penalty and runtime

alibi/prototypes/__init__.py
RobertSamoilescu: Addressed in [here](https://github.com/SeldonIO/alibi/pull/626#discussion_r859867164)

alibi/saving.py
RobertSamoilescu: Decided to not implement saving/loading functionality because the data is "fixed" in the sense that you can't reuse the same object for finding prototypes in new data (unlike explainers, ML models etc.)

alibi/prototypes/protoselect.py
RobertSamoilescu: Decided to follow `Explainer` interface for now. To be changed in other iterations if required.

alibi/prototypes/protoselect.py
RobertSamoilescu: Decided to follow with the first suggestion as it is consistent with `alibi-detect`. For consistency reasons (see notation for counterfactual) I decided to change `y_ref` to `Y_ref`.

alibi/prototypes/protoselect.py
RobertSamoilescu: Fixed and included a test.

alibi/prototypes/protoselect.py
RobertSamoilescu: Decided to keep it for now.

README.md
```diff
@@ -196,79 +197,91 @@ Key:
...
+    [Iris dataset](https://docs.seldon.io/projects/alibi/en/stable/examples/anchor_tabular_iris.html),
+    [movie sentiment classification](https://docs.seldon.io/projects/alibi/en/stable/examples/anchor_text_movie.html),
+    [ImageNet](https://docs.seldon.io/projects/alibi/en/stable/examples/anchor_image_imagenet.html),
+    [fashion MNIST](https://docs.seldon.io/projects/alibi/en/stable/examples/anchor_image_fashion_mnist.html)
 
 - Contrastive Explanation Method (CEM, [Dhurandhar et al., 2018](https://papers.nips.cc/paper/7340-explanations-based-on-the-missing-towards-contrastive-explanations-with-pertinent-negatives))
-  - [Documentation](https://docs.seldon.io/projects/alibi/en/stable/methods/CEM.html)
-  - Examples: [MNIST](https://docs.seldon.io/projects/alibi/en/stable/examples/cem_mnist.html),
-    [Iris dataset](https://docs.seldon.io/projects/alibi/en/stable/examples/cem_iris.html)
+  - [Documentation](https://docs.seldon.io/projects/alibi/en/latest/methods/CEM.html)
```
jklaise, line 224: Please revert all occurrences of `latest` to `stable`.

alibi/api/defaults.py
```diff
@@ -252,7 +252,7 @@
 
 
 DEFAULT_META_PROTOSELECT = {"name": None,
-                            "type": ["prototypes"],
+                            "type": ["data"],
                             "explanation": ["global"],
                             "params": {},
                             "version": None}  # type: dict
```
jklaise, line 258: @arnaudvl we went with this metadata format.

alibi/api/interfaces.py
```diff
@@ -150,6 +150,23 @@ def save(self, path: Union[str, os.PathLike]) -> None:
         save_explainer(self, path)
 
 
+class Summariser(abc.ABC, Base):
+    """
+    Base class for explainer algorithms from :py:mod:`alibi.prototypes`.
```
jklaise, line 155: explainer -> prototype

alibi/prototypes/protoselect.py
```diff
@@ -12,44 +11,51 @@
...
+                 lambda_penalty: Optional[float] = None,
                  batch_size: int = int(1e10),
-                 preprocess_fn: Optional[Callable[[np.ndarray], np.ndarray]] = None,
-                 verbose: bool = False,
-                 **kwargs):
+                 preprocess_fn: Optional[Callable[[Union[list, np.ndarray]], np.ndarray]] = None,
+                 verbose: bool = False):
         """
-        Constructor
+        Prototype selection for dataset distillation and interpretable classification propose by
```
jklaise, line 30: propose -> proposed

alibi/prototypes/protoselect.py
```diff
@@ -66,60 +72,76 @@ def __init__(self,
...
+            Labels of the reference dataset. The labels are expected to be represented as integers `[0, 1, ..., L-1]`,
+            where `L` is the number of classes in the reference dataset.
         X
-            Reference dataset to be summarized.
-        X_labels
-            Labels of the reference dataset.
-        Y
             Dataset to choose the prototypes from. If ``None``, the prototypes will be selected from the reference
-            dataset `X`.
+            dataset `X_ref`.
```
jklaise, line 98: I still find it slightly confusing because of the different behaviour whether `X` is present or not.

For instance, if `X` is present then the docstring for `X_ref` seems inaccurate (it would be `X` that is summarized?).

Also wording could be confusing, I assume `dataset to be summarised` and `dataset to choose the prototypes from` mean the same thing (i.e. "summarize" == "choose prototypes"?).

Finally, it would probably be helpful to include some of this from the method page in the docstrings too: 
image

alibi/prototypes/protoselect.py
```diff
@@ -66,60 +72,76 @@ def __init__(self,
...
         # that loss term which tries to avoid including in an epsilon ball elements belonging to other classes
         # will always be 0. Still the first term of the loss tries to cover as many examples as possible with
         # minimal overlap between the epsilon balls corresponding to the other prototypes.
-        self.X_labels = X_labels.astype(np.int32) if (X_labels is not None) else np.zeros((len(X), ), dtype=np.int32)
+        self.Y_ref = Y_ref.astype(np.int32) if (Y_ref is not None) else np.zeros((len(X_ref),), dtype=np.int32)
+
+        # redefine the labels, so they are in the interval [0, len(np.unique(Y_ref)) - 1].
+        # For example, if the labels provided were [40, 51], internally, we relabel them as [0, 1].
+        # This approach can reduce computation and memory allocation, as without the intermediate mapping we had to
+        # allocate memory corresponding 52 labels for some internal matrices.
```
jklaise, line 119: 52?

RobertSamoilescu: 0-51 -> 52 numbers

doc/source/methods/ProtoSelect.ipynb
```diff
@@ -18,12 +18,12 @@
...
+    "[Bien and Tibshirani (2012)](https://arxiv.org/abs/1202.5933) proposed ProtoSelect, which is a prototype selection method with the goal of constructing not only a condensed view of dataset but also on an interpretable model (applicable to classification only). Prototypes can be defined as instances that are representative of the entire reference data distribution. Formally, consider a dataset of training points $\\mathcal{X} = \\{x_1, ..., x_n \\} \\subset \\mathbf{R}^p$ and their corresponding labels $\\mathcal{Y} = \\{y_1, ..., y_n\\}$, where $y_i \\in \\{1, 2, ..., L\\}$. ProtoSelect finds sets $\\mathcal{P}_{l} \\subseteq \\mathcal{X}$ for each class $l$ such that the set union of $\\mathcal{P}_1, \\mathcal{P}_2, ..., \\mathcal{P}_L$ would provided a distilled view of the reference dataset $(\\mathcal{X}, \\mathcal{Y})$.\n",
```
jklaise, line 21: of dataset -> of a dataset
on an interpretable model -> an interpretable model

jklaise (changes requested):
To summarize an offline discussion:
 - Since `X_ref` (and `y_ref`) is the main dataset of interest (the one being summarized), it makes sense to rename. In fact, to limit confusion let's stick to the paper's notation (`X_ref` -> `X`, `Y_ref` -> `y`, `X`->`Z`)
 - Expand a little bit in the docstrings and the method description on what happens if a `Z` is provided - i.e. the dataset to be summarized is still `X` but it's summarized by prototypes of a different dataset. There are some obvious limitations of this (e.g. have to assume `Z` is distributed the same as `X` to provide any utility) and the use cases are unclear, and we don't have an example of this either, so given that should keep the discussion succinct but accurate.

alibi/prototypes/protoselect.py
```diff
@@ -78,59 +78,59 @@ def __init__(self,
...
-            X: Optional[Union[list, np.ndarray]] = None) -> 'ProtoSelect':
+            X: Union[list, np.ndarray],
+            y: Optional[np.ndarray] = None,
+            Z: Optional[Union[list, np.ndarray]] = None) -> 'ProtoSelect':
         """
-        Fit the summariser by setting the reference dataset. This step forms the kernel matrix in memory
-        which has a shape of `Nx x Ny`, where `Nx` is the number of instances in `X` and `Ny` is the
-        number of instances in `Y`.
+        Fit the summariser. This step forms the kernel matrix in memory which has a shape of `NX x NZ`,
+        where `NX` is the number of instances in `X` and `NZ` is the number of instances in `Z`.
```
jklaise, line 86: Add that the kernel matrix will be of size `NX x NX` if `Z` is not provided (can't expect users to read all of the docstring). Even better, I would first mention the usual use case `NX x NX` and then mention the other case where `Z` is not None.

doc/source/methods/ProtoSelect.ipynb
```diff
@@ -123,36 +123,42 @@
...
+    "In a more general case, we can specify in optional dataset $Z$ to choose the prototypes from (see the documentation of the [fit](https://docs.seldon.io/projects/alibi/en/stable/api/alibi.prototypes.protoselect.html#alibi.prototypes.protoselect.ProtoSelect.fit) method). In this scenario, the dataset to be summarised is still $X$, but it is summarised by prototypes belonging to the dataset $Z$. Furthermore, note that we only need to specify the labels for the $X$ set through $y$, but not for $Z$. In case the labels $y$ are missing, the method implicitly assumes that all the instances belong to the same class. This means that the second term in the objective, $\\Delta\\nu(x_j, l)$, will be 0. Thus, the algorithm will try to find prototypes that cover as many data instances as possible, with minimum overlap between their corresponding $\\epsilon$-balls.\n",
```
jklaise, line 142: in optional -> an optional

doc/source/methods/ProtoSelect.ipynb
```diff
@@ -123,36 +123,42 @@
...
+    "Another important observation is that the summary returns the prototypes with their corresponding labels although no labels were provided for $Z$. This is possible since each prototype $z$ will belong to a prototype set $\\mathcal{P}_l$, and thus we can assign a label $l$ to $z$. Following the summarisation step, one can train an interpretable 1-KNN classifier on the returned prototypes even for an unlabeled dataset $Z$.\n",
+    "\n",
+    "<div class=\"alert alert-block alert-warning\">\n",
+    "<b>Warning</b> \n",
+    "    \n",
+    "If the optional argument $Z$ is not provided, it is automatically set to $X$. Although the labels of the data instances belonging to $Z$ are available in this case, the dataset $Z$ is still viewed as an unlabeled dataset. This means that a prototype $z_i \\in Z$ belonging to the class $l$ according to the labels $y$, can be a prototype for a class $k \\ne l$. \n",
+    "\n",
+    "</div>"
```
jklaise, line 161: This is not rendering properly, have a look at the source code of an existing notebook doc that uses these directives, e.g. https://docs.seldon.io/projects/alibi/en/latest/methods/TreeSHAP.html

doc/source/examples/protoselect_adult_cifar10.ipynb
```diff
@@ -41,21 +45,27 @@
...
     "\n",
+    "from alibi.api.interfaces import Explanation\n",
     "from alibi.datasets import fetch_adult\n",
     "from alibi.prototypes import ProtoSelect, visualize_image_prototypes\n",
     "from alibi.utils.kernel import EuclideanDistance\n",
-    "from alibi.prototypes.protoselect import cv_protoselect_euclidean"
+    "from alibi.prototypes.protoselect import cv_protoselect_euclidean\n",
+    "\n",
+    "%load_ext autoreload\n",
+    "%autoreload 2"
```
jklaise, line 55: Remove autoreload commands.

doc/source/examples/protoselect_adult_cifar10.ipynb
```diff
@@ -81,41 +92,93 @@
...
     "        \n",
     "    return pd.concat(dfs)"
    ]
   },
   {
    "cell_type": "markdown",
+   "id": "99ef2422",
+   "metadata": {},
+   "source": [
+    "Utility function to display image prototypes."
```
jklaise, line 133: This is currently under "Adult Census" heading, shuffle headings around?

jklaise (commented):
Thanks Robert, looks good! Just a couple more comments on typos and docs.

jklaise (approved):
Thanks Robert! Marking as "ready to merge". Will merge when we're closer to the release.

Issue #627, opened 2022-04-13
Integrated Gradients target


Huxwell:
I am struggle to understand what target parameter means and how to use it
According to https://docs.seldon.io/projects/alibi/en/latest/methods/IntegratedGradients.html
`target: Defines which element of the model output is considered to compute the gradients.`

My model is a multicategory images regression with last layer
`layers.Dense(3, activation=tf.nn.relu)` , 3 because I have 3 differents regressed scores

When I do
```
import tensorflow as tf
from alibi.explainers import IntegratedGradients


ig  = IntegratedGradients(model_ours,
                          layer=None,
                          method="gausslegendre",
                          n_steps=50,
                          internal_batch_size=100)
ig.explain(X, baselines=None, target=1)
``` 
I will actually get some nicely looking explanations, but they don't seem meaningful. I can change target to 99999999, or -999, or np.array([[1,1,1]]) or in fact np.array([[1,1,1,1,1,1,1,1,1,1]]) to get different explanations, but I don't know what they mean in such case. 
The explanation from your manual seems to be wrong, target doesn't seem to say "which element of the model output is considered to compute the gradient", it rather seems to supply this element's value. However I don't understand dimensionality in such case.

Let's say I make sure I am using the last layer with:

```
print(X.shape,y)
ig  = IntegratedGradients(model_ours,
                          layer = model_ours.layers[-1],
                          method="gausslegendre",
                          n_steps=50,
                          internal_batch_size=100)
ig.explain(X, baselines=None, target=np.array([[5.5,1.0,4.5]])) #this is correct label for this picture
```
Ok, got some explanations, I supplied numbers for 3 outputs, should be fine.
However I can also do, `ig.explain(X, baselines=None, target=np.array([[99]*99]))` and I also get some explanations. I don't understand why didn't the code crash - I just provided a 99 values to 3 neurons, how were they distributed?
Did only first [99,99,99] got feed to the network? Or maybe just the first value [99]?

I have a second question, after I run the code above, I will get:
```
 'predictions': array([[3.5566006, 0.7617401, 3.6028929]], dtype=float32),
  'deltas': array([0.]),
  'target': [array([99, 99, 99, 99, 99, 99, 99, 99, 99, 99, 99, 99, 99, 99, 99, 99, 99,
       99, 99, 99, 99, 99, 99, 99, 99, 99, 99, 99, 99, 99, 99, 99, 99, 99,
[...]

PR #635, opened 2022-04-21
Extend `IntegratedGradients` `target` parameter selection to support rank > 2 model outputs

gipster:
This pull request addresses the [issue](https://github.com/SeldonIO/alibi/issues/627) related to target dimensionality in the [IntegratedGradients](https://github.com/SeldonIO/alibi/blob/61f0c6afd7e5e23585fc3661b4a4aa137db9abb7/alibi/explainers/integrated_gradients.py#L711) class.  

The present code behaves correctly for the most common use cases (classification and regression models with output dimensionality <= 2) when the targets passed are correctly formatted. However, the `_select_target` function returns incorrect outputs for output dimensions > 2 and in some case doesn't raise errors when unusual targets are passed. 

This pull request should address this issue with the following changes

* Substitute the line 
  ```
  preds = tf.linalg.diag_part(tf.gather(preds, targets, axis=1))
  ```
  with the line 
  ```
  preds = tf.gather_nd(preds, tf.expand_dims(targets, axis=1), batch_dims=1)
  ```
  in the function [_select_target](https://github.com/SeldonIO/alibi/blob/61f0c6afd7e5e23585fc3661b4a4aa137db9abb7/alibi/explainers/integrated_gradients.py#L98) . In this new version, passing out or range target will raise an error for any output dimension.
* Added the function `_check_target` ensuring that the target dimensionality is compatible with the model's output dimensionality


alibi/explainers/integrated_gradients.py
```diff
@@ -112,13 +112,17 @@ def _select_target(preds: tf.Tensor,
     Selected predictions.
 
     """
+    if not isinstance(targets, tf.Tensor):
+        targets = tf.convert_to_tensor(targets)
```
jklaise, line 116: Is this necessary? If it is, then the `preds` type hint should change to accomodate other types. Otherwise conversion to `tf.Tensor` should be done outside this function.

gipster:
* Added exceptions in the `_check_target` function to ensure that any badly formatted targets will result in a error.
* Added tests for the functions `_check_target` and [_select_target](https://github.com/SeldonIO/alibi/blob/61f0c6afd7e5e23585fc3661b4a4aa137db9abb7/alibi/explainers/integrated_gradients.py#L98).
* Converted the output of [_get_target_from_target_fn](https://github.com/SeldonIO/alibi/blob/61f0c6afd7e5e23585fc3661b4a4aa137db9abb7/alibi/explainers/integrated_gradients.py#L560) into integers types. Targets must always be integers.

alibi/explainers/integrated_gradients.py
```diff
@@ -543,20 +547,85 @@ def _format_target(target: Union[None, int, list, np.ndarray],
...
+    """
+
+    if target is not None:
+
+        if not (target.dtype == int):
+            raise ValueError("Targets must be integers")
+
+        if target.shape[0] != nb_samples:
+            raise ValueError(f"First dimension in target must be egual to nb of samples. "
+                             f"Found target 1st dimension {target.shape[0]}; nb samples: {nb_samples}")
```
jklaise, line 592: egual -> equal
dimension -> dimension:

alibi/explainers/integrated_gradients.py
```diff
@@ -543,20 +547,85 @@ def _format_target(target: Union[None, int, list, np.ndarray],
...
+
+        if len(target.shape) > 2:
+            raise ValueError("Targets must be 1-d or 2-d arrays. In 2-d arrays, each column must contain "
+                             "the target index of the corresponding dimension in the model's output tensor.")
+
+        if len(output_shape) == 1:
+            onbdim, tnbdim = 1, len(target.shape)
```
jklaise, line 599: Variables names confusing, took me a minute to decipher, consider renaming to `output_nbdim` and `target_nbdim`.

alibi/explainers/integrated_gradients.py
```diff
@@ -543,20 +547,85 @@ def _format_target(target: Union[None, int, list, np.ndarray],
...
+            tmax, tmin = target.max(axis=0), target.min(axis=0)
+
+            if tmax > 1:
+                raise ValueError(f"Targets values {tmax} out of range for output shape {output_shape[-1]} ")
```
jklaise, line 603: Please check grammar here and elsewhere.

alibi/explainers/integrated_gradients.py
```diff
@@ -543,20 +547,85 @@ def _format_target(target: Union[None, int, list, np.ndarray],
...
+
+        if target.shape[0] != nb_samples:
+            raise ValueError(f"First dimension in target must be egual to nb of samples. "
+                             f"Found target 1st dimension {target.shape[0]}; nb samples: {nb_samples}")
+
+        if len(target.shape) > 2:
+            raise ValueError("Targets must be 1-d or 2-d arrays. In 2-d arrays, each column must contain "
+                             "the target index of the corresponding dimension in the model's output tensor.")
+
+        if len(output_shape) == 1:
```
jklaise, line 598: This whole branch is only valid if the (implicit) task is classification and outputs are labels rather than probabilities (hence output shape is 1-d), right? Can you add a comment here to clarify?

alibi/explainers/integrated_gradients.py
jklaise: Also perhaps need to make the message less confusing in general (here and elsewhere), e.g. consider:
`ValueError: Targets value 4 out of range for output shape 2`

But `output_shape` is the whole tuple not the length of the last dimension.

alibi/explainers/tests/test_integrated_gradients.py
```diff
@@ -650,6 +652,68 @@ def test_run_forward_from_layer(layer_nb,
...
+@pytest.mark.parametrize('args', SELECT_ARGS)
+def test_select_target(args):
+    preds = tf.convert_to_tensor(args['preds'])
+    target = tf.convert_to_tensor(args['target'])
+    expected = args['expected']
+
+    selected = _select_target(preds, target)
+    print(expected)
+    print('=======')
+    print(selected)
```
jklaise, line 713: `print` statements unnecessary

alibi/explainers/integrated_gradients.py
```diff
@@ -543,20 +547,85 @@ def _format_target(target: Union[None, int, list, np.ndarray],
...
+        if not (target.dtype == int):
+            raise ValueError("Targets must be integers")
+
+        if target.shape[0] != nb_samples:
+            raise ValueError(f"First dimension in target must be egual to nb of samples. "
+                             f"Found target 1st dimension {target.shape[0]}; nb samples: {nb_samples}")
+
+        if len(target.shape) > 2:
+            raise ValueError("Targets must be 1-d or 2-d arrays. In 2-d arrays, each column must contain "
+                             "the target index of the corresponding dimension in the model's output tensor.")
```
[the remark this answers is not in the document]
gipster: ok, I think I used the word dimension when I should have used the word rank.

With `gather_nd`, you can actually get the correct target for output tensors with rank higher than 2. 

For examples, if you have a classification problem with 10 classes, your output tensor will be a  rank 2 tensor (a matrix) of dimension nb_samples X 10 . Now in this case you target must be rank 1 (1-d), of length nb_samples, and with range 0-9, each value representing the column position in the output tensor. 

If you have a tensor of higher rank  as output, for example an auto encoder with (nb_samples, 28, 28, 3) rank 3 tensor in the  output (ignoring the nb_samples dimension), you need a rank 2 target (a matrix) nb_samples X 3 , where the first 2 columns have range 0-27 and the last column have range 0-2. 

In other words, for output ranks > 1 , if you have a rank-n tensor as output, you need a rank 2 tensor with dimensions nb_samples x n. Each row represents the location of an element in the output tensor.

Not sure if it's clear enough

alibi/explainers/tests/test_integrated_gradients.py
```diff
@@ -650,6 +652,68 @@ def test_run_forward_from_layer(layer_nb,
...
+                'target': np.array([0,
+                                    0]),
+                'expected': np.array([0.0,
+                                      1.0])},
+               {'preds': np.array([[[0.0, 0.1], [1.0, 1.1]],
+                                   [[2.0, 2.1], [3.0, 4.1]]]),
+                'target': np.array([[0, 0],
+                                   [0, 0]]),
+                'expected': np.array([[0.0],
+                                      [2.0]])}]
```
jklaise, line 687: I see that this test case tests a 3-dimensional output and a 2-dimensional target which partially answers my question above why 2D targets are allowed. However, I though we only support 2-dimensional outputs anyway?

Does this mean that this PR implicitly allows up to 3D output?

gipster: yes, that's right. Any rank for the output tensor are allowed now, not only 3D. Refer to comment above

alibi/explainers/integrated_gradients.py
gipster: It's valid when you have the binary classification with  single squash output (probability of class 1 but returns tensor of shape (nb_samples) instead of (nb_samples, 1) ).

jklaise: Hmm remind me, is it valid to pass a model that outputs labels instead of probabilities? Or does the model always have to be probabilistic (in the classification case that is)?

alibi/explainers/integrated_gradients.py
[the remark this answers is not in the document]
jklaise: I think that makes sense, perhaps worth writing a bit about it in the docstrings (not necessary to be comprehensive as we don't currently have examples on outputs of >2 rank. `captum` describes it quite well: https://captum.ai/api/integrated_gradients.html

alibi/explainers/integrated_gradients.py
```diff
@@ -798,11 +867,16 @@ def explain(self,
...
-            for the same element of the output for all data points.
-            It must be provided if the model output dimension is higher than 1.
+            Target can be a numpy array, a list or a numeric value.
+            Numeric values are only valid if the model's output is a rank-n tensor
+            with n <= 2 (regression and classification models).
+            If a numeric value is passed, the gradients are calculated for
+            the same element of the output for all data points.
             For regression models whose output is a scalar, target should not be provided.
             For classification models `target` can be either the true classes or the classes predicted by the model.
+            It must be provided if the model output dimension is higher than 1.
```
jklaise, line 877: Make the language consistent to always talk about the rank of the output (instead of dimesnion)?

alibi/explainers/integrated_gradients.py
```diff
@@ -798,11 +867,16 @@ def explain(self,
...
+            If the model's output is a rank-n tensor with n > 2,
+            the target must a rank-2 numpy array or a list of lists (a matrix) with dimensions nb_samples X (n-1) .
```
jklaise, line 879: must ->  must be

alibi/explainers/integrated_gradients.py
```diff
@@ -543,20 +547,85 @@ def _format_target(target: Union[None, int, list, np.ndarray],
...
+            raise ValueError("Target must be a rank-1 or a rank-2 tensor. If target is a rank-2 tensor, "
+                             "each column contains the index of the corresponding dimension "
+                             "in the model's output tensor.")
+
+        if len(output_shape) == 1:
+            out_rank, target_rank = 1, len(target.shape)
+            tmax, tmin = target.max(axis=0), target.min(axis=0)
+
+            if tmax > 1:
+                raise ValueError(f"Target value {tmax} out of range for output shape = 1 ")
```
jklaise, line 604: Should this be either `out of range for output shape {output_shape}` or `out of range for rank-1 output` instead?

alibi/explainers/integrated_gradients.py
gipster: no, but in this case dimension doesn't refer to the rank of the tensor, it refers to the number of classes

alibi/explainers/integrated_gradients.py
gipster: ok, I think it's actually better to put `for output shape {output_shape}` everywhere

alibi/explainers/integrated_gradients.py
gipster: Not sure about that.

alibi/explainers/integrated_gradients.py
```diff
@@ -543,20 +547,85 @@ def _format_target(target: Union[None, int, list, np.ndarray],
...
+            if (tmax >= output_shape[1:]).any():
+                raise ValueError(f"Target value {tmax} out of range for output shape {output_shape} ")
+
+        if (tmin < 0).any():
+            raise ValueError(f"Negative value {tmin} for target. Targets represent positional "
+                             f"arguments and cannot be negative")
+
+        if out_rank != target_rank:
+            raise ValueError(f"The last dimension of target  must match the rank of the model's output tensor. "
+                             f"Found target last dimension: {target_rank}; model's output rank: {out_rank}")
```
jklaise, line 626: Should this be done as the very first check as it could mess up everything immediately if not passed right?

Although I'm confused since it seems `target_rank` is not always `len(target.shape)` as it's defined as `target.shape[-1]` for rank>2 outputs...

alibi/explainers/integrated_gradients.py
```diff
@@ -543,20 +547,85 @@ def _format_target(target: Union[None, int, list, np.ndarray],
...
+
+        if len(output_shape) == 1:
+            out_rank, target_rank = 1, len(target.shape)
+            tmax, tmin = target.max(axis=0), target.min(axis=0)
+
+            if tmax > 1:
+                raise ValueError(f"Target value {tmax} out of range for output shape = 1 ")
+
+        elif len(output_shape) == 2:
+            out_rank, target_rank = 1, len(target.shape)
```
jklaise, line 607: I fee like `out_rank` should be `2` here but it seems you are using different conventions which makes it more confusing...

alibi/explainers/integrated_gradients.py
jklaise: Ok so in that case the sentence should be changed so it's clear it applies to classification and talks about classes (as opposed to the niche case of 2-class classification with 1-class output probability?).

alibi/explainers/integrated_gradients.py
jklaise: I.e. are we ignoring batch dimension in some cases but not others?

gipster: no, because out_rank doesn't include the dimension referring to the number of samples

gipster: we are not ignoring it only in the case of the squash output (output_shape = 1)

alibi/explainers/integrated_gradients.py
gipster: The case with output_shape > 2 is different because the last dimension of the target must match the rank of the output tensor (excluding batch dimension )

alibi/explainers/integrated_gradients.py
jklaise: But `if len(output_shape) == 1` then `out_rank=1` so we are not ignoring it in this case, but are ignoring it in all other cases?

It's just confusing to read that `out_rank=1` in both `len(output_shape) == 1` and `len(output_shape) == 2`... Is there a better way to code this up or at least comment in the code?

jklaise: (I think supporting squash-output for 2-class classification is giving a lot of headache here and elsewhere...)

gipster: well it gave me a headache yesterday when I was writing these warnings and getting errors for squash output

jklaise (approved):

PR #636, opened 2022-04-22
Update CI to trigger on PRs for feature branches but not draft PRs

### LINKED ISSUE #633: Update CI to run on branches other than `master`
Since we're starting to use feature branches to split out new feature work between code and docs+examples (e.g. #577 ) it would be useful to configure CI to run for PRs pointing to non-`master` branches:

https://github.com/SeldonIO/alibi/blob/61f0c6afd7e5e23585fc3661b4a4aa137db9abb7/.github/workflows/ci.yml#L7-L15

jklaise:
Resolves #633.

This largely brings the CI workflow closer to the one in `alibi-detect`: https://github.com/SeldonIO/alibi-detect/blob/master/.github/workflows/ci.yml

1. Trigger CI on PRs to any branch since we're starting to use feature branches more. I allow now all branches instead of just `features/*` as we already have one that is named `feature/*`...
2. Don't trigger CI on draft PRs
3. Updated comment on daily crontab (previously said weekly - same comment should be changed on `alibi-detect` workflow)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,14 +6,14 @@ defaults:
 
 on:
   # Trigger the workflow on push or pull request,
-  # but only for the master branch
+  # push only for the master branch.
   push:
     branches:
       - master
   pull_request:
-    branches:
-      - master
-  # Trigger workflow once per week
+    # don't trigger for draft PRs
+    types: [opened, synchronize, reopened, ready_for_review]
+  # Trigger workflow once per day at midnight
   schedule:
     - cron: '0 0 * * *'
   # Trigger the workflow on manual dispatch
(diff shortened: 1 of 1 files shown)


ascillitoe:
> 1. Trigger CI on PRs to any branch since we're starting to use feature branches more. I allow now all branches instead of just `features/*` as we already have one that is named `feature/*`...

We could also go with `feature`? There are no feature branches using either of the new naming conventions in alibi-detect so we could update that to match? Unless you really do think it best to test PR's into all branches...


> 3\. Updated comment on daily crontab (previously said weekly - same comment should be changed on `alibi-detect` workflow)

I'll update along with any changes required wrt to first decision (the feature branch naming one).

jklaise:
@ascillitoe changed to run PRs on `master` and `'feature/*'` as suggested.

ascillitoe:
> @ascillitoe changed to run PRs on `master` and `'feature/*'` as suggested.

Nice, I'll update `alibi-detect` to `feature` too.

ascillitoe (approved):
LGTM!

PR #638, opened 2022-04-25
Update to mypy 0.9

ascillitoe:
Update to mypy 0.900. This is a precursor to #605.

See https://github.com/SeldonIO/alibi-detect/pull/490 for more details.


ascillitoe:
@jklaise @mauicv any thoughts on the new type ignores detailed above?

jklaise (approved):
LGTM. I would wager that the `type-var` to `arg-type` error was a bug in the older version of `mypy`.

ascillitoe:
> I would wager that the `type-var` to `arg-type` error was a bug in the older version of `mypy`.

Yeh agreed, `arg-type` does make more sense for those lines.

PR #641, opened 2022-04-26
Add missing batched similarity method explanations functionality

mauicv:
# Adds batched similarity explanations

### What this changes:

- `explain` method returns a matrix of shape `(n_test, n_train)`. The elements of which are the similarity between each test instance to each training instance. `n_test` is the number of test instances passed to the method. If a single instance is passed without a leading batch dimension then the matrix returned will have shape `(1, n_train)`
- The similarity functions now expect two matrices with shapes: `(n_test, n_test_grads)` and `(n_train, n_train_grads)`. So the vectors of the gradients for each test instance and train instance. Note ` n_test_grads==n_train_grads`.
- I've also added tests for each of `grad_cos` and `grad_dot` for multiple samples with and without the gradients being precomputed. 

___

### Method

The explain method is now implemented as follows:

Note: If `precompute_grads` is true then the gradients for each of the training instances will have been computed already and stored in the `self.grad_X_train`.

1. First we preprocess the `X` and `Y` parameters so that they both have leading batch dimension. So if they're a single test instance they now are a member of a batch of size 1. 
2. We then compute the gradients for the test instances and accumulate them into a vector of shape `(batch_size, n_gradients)`. This is places in a variable `grads_X_test`
3. If `self.precompute_grads` is true then we pass `self.grads_X_train` to the similarity function to get the scores. The scores will be a matrix of shape `(n_test, n_train)`.
4. If `self.precompute_grads` is false then we compete the training gradients on the fly. 
    - We iterate through the training data and compute the gradients for each training instance.
    - for each instance we pass it to the similarity function along with all the test instances. This returns a vector of similarities for that training instance to each of the test instances. We accumulate these in a score vector of the correct size.
5. Finally the score values are passed to the `build_explanation` function.

___

### Examples:

1. [Tensorflow notebook](https://gist.github.com/mauicv/aab7233515cb7349f9620304be2911c3)
2. [Torch notebook](https://gist.github.com/mauicv/20263a3dd5746a17f89271bdad456830)

### DIFF
--- a/alibi/explainers/tests/test_simiarlity/test_grad_methods_integration.py
+++ b/alibi/explainers/tests/test_simiarlity/test_grad_methods_integration.py
@@ -61,7 +61,7 @@ def test_correct_grad_dot_sim_result_torch(seed, normed_ds):
     explainer = explainer.fit(normed_ds, normed_ds)
     explanation = explainer.explain(normed_ds[0], Y=normed_ds[0])
     last = np.dot(normed_ds[0], normed_ds[0])
-    for ind in explanation['ordered_indices'][1:]:
+    for ind in explanation['ordered_indices'][0][1:]:
         current = np.dot(normed_ds[0], normed_ds[ind])
         assert current <= last
         last = current
@@ -83,7 +83,7 @@ def test_correct_grad_cos_sim_result_torch(seed, ds):
     explainer = explainer.fit(ds, ds)
     explanation = explainer.explain(ds[0], Y=ds[0])
     last = 0
-    for ind in explanation['ordered_indices'][1:]:
+    for ind in explanation['ordered_indices'][0][1:]:
         current = compute_angle(ds[0], ds[ind])
         assert current >= last
         last = current
@@ -105,8 +105,8 @@ def test_grad_cos_result_order_torch(seed):
     )
     explainer = explainer.fit(ds, ds)
     explanation = explainer.explain(ds[0], Y=ds[0])
-    assert (ds[explanation['ordered_indices'][1]] == ds[1]).all()
-    assert (ds[explanation['ordered_indices'][-1]] == ds[-1]).all()
+    assert (ds[explanation['ordered_indices'][0][1]] == ds[1]).all()
+    assert (ds[explanation['ordered_indices'][0][-1]] == ds[-1]).all()
 
 
 def test_grad_dot_result_order_torch(seed):
@@ -124,8 +124,8 @@ def test_grad_dot_result_order_torch(seed):
     )
     explainer = explainer.fit(ds, ds)
     explanation = explainer.explain(ds[0], Y=ds[0])
-    assert (ds[explanation['ordered_indices'][0]] == ds[-1]).all()
-    assert (ds[explanation['ordered_indices'][-1]] == ds[1]).all()
+    assert (ds[explanation['ordered_indices'][0][0]] == ds[-1]).all()
+    assert (ds[explanation['ordered_indices'][0][-1]] == ds[1]).all()
 
 
 def loss_tf(y, x):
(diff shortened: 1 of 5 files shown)


alibi/explainers/tests/test_simiarlity/test_grad_methods_integration.py
```diff
@@ -211,5 +211,123 @@ def test_grad_cos_result_order_tf(seed):
     )
     explainer = explainer.fit(ds, ds)
     explanation = explainer.explain(ds[0], Y=ds[0])
-    assert (ds[explanation['ordered_indices'][1]] == ds[1]).all()
-    assert (ds[explanation['ordered_indices'][-1]] == ds[-1]).all()
+    assert (ds[explanation['ordered_indices'][0][1]] == ds[1]).all()
+    assert (ds[explanation['ordered_indices'][0][-1]] == ds[-1]).all()
+
+
+def test_multiple_test_instances_grad_cos():
```
RobertSamoilescu, line 218: Is there a reason why we do not combine `test_multiple_test_instances_grad_cos` and `test_multiple_test_instances_stored_grads_grad_cos` in a single test so we avoid code duplication?
At a first glance it seem like the only difference is `precompute_grads=True`, which you can set it on or of through `pytest.mark.parametrize`. 
Or am I missing something? 
Same applies for the other tests, so I won't repeat myself.

alibi/explainers/tests/test_simiarlity/test_grad_methods_integration.py
```diff
@@ -282,7 +282,17 @@ def test_multiple_test_instances_stored_grads_asym_dot(precompute_grads):
     explainer = explainer.fit(ds, ds)
     explanation = explainer.explain(ds[0:2], Y=ds[0:2])
 
-    # Check asymmetric dot product scores are correct
+    # In the case of `grad_asym_dot` we manually check the similarities as the asymmetry of the metric means ordering
+    # is important.
+    sim_ds_0 = np.array([(1 * 1 + 0 * 0) / 1., (1 * 0.9 + 0 * 0.1) / 1., (1 * 50 + 0 * 50) / 1.])
```
RobertSamoilescu, line 287: I would remove the division by 1. I understand that you want to be explicit about the norm, but I don't think is that necessary. If you really want to make it obvious, just add a comment.

alibi/explainers/tests/test_simiarlity/test_grad_methods_integration.py
```diff
@@ -282,7 +282,17 @@ def test_multiple_test_instances_stored_grads_asym_dot(precompute_grads):
...
+    sim_ds_0.sort()
+    np.testing.assert_almost_equal(explanation.scores[0], sim_ds_0[::-1], decimal=3)
+
+    d = (0.9**2 + 0.1**2)
+    sim_ds_1 = np.array([(0.9 * 1 + 0.1 * 0) / d, (0.9 * 0.9 + 0.1 * 0.1) / d, (0.9 * 50 + 0.1 * 50) / d])
```
RobertSamoilescu, line 292: I would add the division by `d` in a vectorized form: `np.array([...]) / d`.
Not sure about this, but note that the implementation of the similarity function uses and `eps` to avoid division by 0. Although it is not the case, might improve the precision of the assertion bellow: `np.testing.assert_almost_equal(explanation.scores[1], sim_ds_1[::-1], decimal=3)`

alibi/explainers/tests/test_simiarlity/test_grad_methods_integration.py
```diff
@@ -211,5 +211,89 @@ def test_grad_cos_result_order_tf(seed):
...
+    sim_ds_1.sort()
+    np.testing.assert_almost_equal(explanation.scores[1], sim_ds_1[::-1], decimal=3)
+
+    assert (ds[explanation['ordered_indices'][0][0]] == ds[-1]).all()
+    assert (ds[explanation['ordered_indices'][1][0]] == ds[-1]).all()
+    explanation = explainer.explain(ds[-1], Y=ds[-1])
+    assert (explanation.scores == np.array([[1., 0.01, 0.01]], dtype=np.float32)).all()
```
RobertSamoilescu, line 299: I would use something like `np.testing.assert_allclose` for floats

alibi/explainers/tests/test_simiarlity/test_grad_methods_integration.py
mauicv: I wasn't too bothered about the accuracy. It's more just for general sense checking, but I've added `eps` to both.

PR #644, opened 2022-04-27
Add nbstripout

mauicv:
Add [nbstripout](https://github.com/kynan/nbstripout) as a  pre-commit-hook. `nbstripout` is configured to remove everything except the cell outputs. I've also run it against all the files.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,3 +15,7 @@ repos:
         additional_dependencies: [
           types-requests>=2.25.0,
         ]
+  - repo: https://github.com/kynan/nbstripout
+    rev: 0.5.0
+    hooks:
+      - id: nbstripout
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,7 +7,7 @@ repos:
     rev: 1.2.2
     hooks:
       - id: nbqa-pyupgrade
-        args: [--py37-plus]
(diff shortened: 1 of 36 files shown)


jklaise:
This slightly changes our example notebooks as there are no more execution counts besides cells (i.e. `[1]` becomes `[ ]`), this is probably not an issue per-se, but as this is as pre-commit hook then it will not be used by everyone so we will end up with a mix of real and `null` execution counts in the docs. Is there an option to keep the execution counts?

.pre-commit-config.yaml
```diff
@@ -15,3 +15,8 @@ repos:
         additional_dependencies: [
           types-requests>=2.25.0,
         ]
+  - repo: https://github.com/kynan/nbstripout
+    rev: 0.5.0
+    hooks:
+      - id: nbstripout
+        args: [--keep-output]
```
jklaise, line 22: Consider adding `--keep-count`.

mauicv:
- `nbstripout` doesn't actually [remove metadata introduced by pycharm](https://github.com/kynan/nbstripout/issues/163). 
- It can be run so as not to submit changes to the notebooks using `nbstriptout run --dry-run` but seemingly this option doesn't work with the `--keep-outputs` option. 

In order to remove the `pycharm` metadata in #583 I just wrote a quick script that does so using `nbformat`. I've committed it here as potentially it's a simpler solution.

Issue #642, opened 2022-04-26
'multidim_scaling' with inconsistent parameter 'feature_range' for tuple of floats


JEwerszumrode:
As in the docstring of the [multidim_scaling](https://github.com/SeldonIO/alibi/blob/6ce8a51c30041cfaaf656abab0c1202a05284cf3/alibi/utils/distance.py#L164) utils function it is said, that the parameter `feature_range` can be tuple of floats:
```
def  multidim_scaling(...):
   ...
   feature_range
        Tuple with `min` and `max` ranges to allow for perturbed instances. `Min` and `max` ranges can be `float` or
        `numpy` arrays with dimension (`1 x nb of features`) for feature-wise ranges.
```

But some lines later you used to copy each element of the tuble by `new_feature_range = tuple([f.copy() for f in feature_range])`
which yields to comprehensible error of `'float' object has no attribute 'copy'`

This is also used in [CounterfactualsProto](https://github.com/SeldonIO/alibi/blob/1fd0e797a78007d7a0ec8082e12185b9fa731201/alibi/explainers/cfproto.py#L41) as default:
```
class CounterfactualProto(...):
    def __init__(..., feature_range: tuple = (-1e10, 1e10), ...)
```

jklaise:
Fixed in #646.

PR #646, opened 2022-04-28
fix types multidim scaling

arnaudvl:
Addresses #642 .

### DIFF
--- a/alibi/explainers/cfproto.py
+++ b/alibi/explainers/cfproto.py
@@ -38,7 +38,7 @@ class CounterfactualProto(Explainer, FitMixin):
                  shape: tuple,
                  kappa: float = 0.,
                  beta: float = .1,
-                 feature_range: tuple = (-1e10, 1e10),
+                 feature_range: Tuple[Union[float, np.ndarray], Union[float, np.ndarray]] = (-1e10, 1e10),
                  gamma: float = 0.,
                  ae_model: Optional[tf.keras.Model] = None,
                  enc_model: Optional[tf.keras.Model] = None,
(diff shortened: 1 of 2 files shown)


alibi/explainers/cfproto.py
```diff
@@ -178,7 +178,11 @@ def __init__(self,
         self.max_iterations = max_iterations
         self.c_init = c_init
         self.c_steps = c_steps
-        self.feature_range = feature_range
+        self.feature_range = list(feature_range)
+        for _ in range(2):
```
jklaise, line 182: At first I though iterating over only the first 2 items of the tuple is a clever way to get rid of any extra items the user may have passed, but that is actually not the case as we have already assigned to `self.feature_range`, so it will always be of the length the user passed (which could be longer than 2).

Not saying we have to validate explicitly here as we don't do much validation elsewhere (also assuming implicitly below that `feature_range` has at least 2 items), but more of a note for the future.

alibi/explainers/cfproto.py
```diff
@@ -178,7 +178,11 @@ def __init__(self,
         self.max_iterations = max_iterations
         self.c_init = c_init
         self.c_steps = c_steps
-        self.feature_range = feature_range
+        self.feature_range = list(feature_range)
```
jklaise, line 181: Nitpick: might be slightly cleaner to just `feature_range = list(feature_range)` and assign as an attribute once we're done with the pre-processing.

jklaise (approved):

alibi/explainers/cfproto.py
jklaise: Yes, but we never check/validate that it is the case.

alibi/explainers/cfproto.py
```diff
@@ -758,13 +756,13 @@ def fit(self,
 
                 # multidim scaled distances
                 d_abs_abdm, _ = multidim_scaling(d_abdm, n_components=2, use_metric=True,
-                                                 feature_range=self.feature_range,
+                                                 feature_range=self.feature_range,  # type: ignore
```
jklaise, line 759: What was the error here? I thought `multidim_scaling` takes tuples?

If can't be resolved then would be great to narrow down the error code (i.e. `type: ignore[error-code] (same for below).

arnaudvl: `Argument "feature_range" to "multidim_scaling" has incompatible type "Tuple[Any, ...]"; expected "Tuple[Any, Any]"`
Due to for loop being used.

jklaise: OK, let's add the correct `error-code` to the new type ignore's then.

PR #649, opened 2022-05-03
Split Explainer into Base & Explainer

RobertSamoilescu:
This PR splits the `Explainer` class into a `Base` and `Explainer`.
The `Base` class currently implements a structured approach to handle metadata (to be extended). This split allows us to introduce other base classes for new packages that do not conform to the `Explainer` interface. For example, algorithms for prototype selection are not explainers. Calling the `explain` method when, in fact, they are summarizing/distilling the dataset is not appropriate. Furthermore, functionalities such as saving and loading might expect other arguments, and resetting the prediction might not apply. 

Possible to be used for the `confidence` package too.

**Observation.** The functionality of the `Explanation` class is already quite general and works for all packages. Changing the name to something more general would involve a lot of refactoring, which we don’t want to do, at least for now.

### DIFF
--- a/alibi/api/interfaces.py
+++ b/alibi/api/interfaces.py
@@ -69,7 +69,7 @@ class Base:
     Base class for all `alibi` algorithms. Implements a structured approach to handle metadata.
     """
 
-    meta: dict = attr.ib(default=attr.Factory(default_meta), repr=alibi_pformat)  #: Explainer meta-data.
+    meta: dict = attr.ib(default=attr.Factory(default_meta), repr=alibi_pformat)  #: Object's metadata.
 
     def __attrs_post_init__(self):
         # add a name and version to the metadata dictionary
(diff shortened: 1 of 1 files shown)


alibi/api/interfaces.py
```diff
@@ -64,10 +64,11 @@ def _pprint_dict(self, object, stream, indent, allowance, context, level):
...
 
 @attr.s
-class Explainer(abc.ABC):
+class Base:
     """
-    Base class for explainer algorithms
+    Base class for all `alibi` algorithms. Implements a structured approach to handle metadata.
     """
+
     meta: dict = attr.ib(default=attr.Factory(default_meta), repr=alibi_pformat)  #: Explainer meta-data.
```
jklaise, line 72: Explainer -> Object / Algorithm?

PR #652, opened 2022-05-11
Fix incorrect dependency message

mauicv:
# Fix error message issue

This PR fixes the potential for incorrect optional dependency pip install messages to be issued for certain missing dependencies. Previously we used `ERROR_TYPES = {'ray', 'tensorflow', 'torch', 'shap', 'numba'}` for validation of the error type. We used the `err.name` as the optional dependency bucket name within the `pip install` command we give in the import error message. This means that if `numba` was not installed but `shap` was and a user tried to import something like `KernelShap` that's dependent on both then they'd get the message `pip install alibi[numba]` instead of the correct, `pip install alibi[shap]`. 

I've updated the `ERROR_TYPES` set to be a map from the missing dependency to the optional install bucket name and updated the tests to check for this behaviour.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -177,7 +177,7 @@ you will need to:
 
 1. Add it to `extras_require` in `setup.py`.
 2. Create a new `tox` environment in `setup.cfg` with the new dependency.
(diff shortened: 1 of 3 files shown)


jklaise (approved):
LGTM!

PR #663, opened 2022-05-13
Fix `asym_sim_fn` incorrect denominator 

mauicv:
Previously denominator was the test instance where it should be the training instance. Fix changes this and corrects tests to match.


RobertSamoilescu:
Looks good to me!

PR #678, opened 2022-05-17
Doc/reorganize sidebar

jklaise:
Reorganized docs so the sidebar is not as cluttered. Crucially, didn't move or rename any example or methods pages so as not to break the internet. This has a slight quirk in that when you're reading an example notebook, the sidebar displays *all* examples, this behaviour is probably built into the template. In any case I don't think it's a showstopper.


doc/source/overview/high_level.md
```diff
@@ -1082,6 +1082,7 @@ that the ALE plots potentially miss details local to individual instances as the
 :align: center
 :alt: Ale plots for those features that the above counterfactuals have changed the most. 
 ```
+(similarity-explanations)=
 
 ### 5. Similarity explanations
```
ascillitoe, line 1087: Technically not a change, but is the "TBA" in this section intended? Assuming the overview is yet to be written, but maybe we could just add a one-sentence description?

ascillitoe (approved):
LGTM! One minor comment about the "TBA" in the overview.

doc/source/overview/high_level.md
jklaise: I agree, will add a sentence.

gipster: > Technically not a change, but is the "TBA" in this section intended? Assuming the overview is yet to be written, but maybe we could just add a one-sentence description?

Hey Ashley. I left it there as the section is only a placeholder for now. I guess we can add something like "Similarity explanations present instances in the training set that are similar to the instance of interest according to the model", but it's already in the table above.

jklaise: @ascillitoe added a paragraph and an image as we're lacking images in the docs.

ascillitoe: Looks great! One minor typo:
"which forms an explanations"

PR #679, opened 2022-05-17
Export prototypes subpackage into alibi namespace

jklaise:
Noticed that `alibi.prototypes` wasn't exported into `alibi` so things like `alibi.prototypes.__all__` would fail. Adjusted the docs also.

For future, I think we can remove the `*__all__` lists from the getting started docs as it doesn't add much and introduces another place that needs to be manually maintained.

### DIFF
--- a/alibi/__init__.py
+++ b/alibi/__init__.py
@@ -1,4 +1,4 @@
-from . import confidence, datasets, explainers, utils
+from . import confidence, datasets, explainers, prototypes, utils
 from .version import __version__  # noqa F401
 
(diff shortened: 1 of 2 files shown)


ascillitoe:
Agreed that displaying the result of `alibi.prototypes.__all__` in the docs doesn't add much. We could just mention it can be used, but not show the result. Alternatively, for code executions like these we could explore [sphinx-exec-code](https://sphinx-exec-code.readthedocs.io/en/latest/) but maybe its overkill here...

ascillitoe (approved):
LGTM!

PR #684, opened 2022-05-26
Fix numpy1.22 typing issues

mauicv:
With the release of numba 0.55.2 (https://github.com/numba/numba/blob/0.55.2/CHANGE_LOG) numpy 1.22 is now installable. This PR fixes typing issues that arose as a result of this version of numpy. See also [Related pr in alibi-detect](https://github.com/SeldonIO/alibi-detect/pull/520)

### DIFF
--- a/alibi/prototypes/protoselect.py
+++ b/alibi/prototypes/protoselect.py
@@ -403,7 +403,7 @@ def cv_protoselect_euclidean(trainset: Tuple[np.ndarray, np.ndarray],
             if quantiles[0] > quantiles[1]:
                 raise ValueError('The quantile lower-bound is greater then the quantile upper-bound.')
             quantiles = np.clip(quantiles, a_min=0, a_max=1)
-            min_dist, max_dist = np.quantile(a=dist, q=quantiles)
+            min_dist, max_dist = np.quantile(a=dist, q=np.array(quantiles))
         else:
             min_dist, max_dist = np.min(dist), np.max(dist)
         # define list of values for eps
--- a/alibi/confidence/trustscore.py
+++ b/alibi/confidence/trustscore.py
@@ -179,7 +179,7 @@ class TrustScore:
             X = X.reshape(X.shape[0], -1)
 
         # init distance matrix: [nb instances, nb classes]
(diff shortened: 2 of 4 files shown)


alibi/prototypes/protoselect.py
```diff
@@ -403,7 +403,7 @@ def cv_protoselect_euclidean(trainset: Tuple[np.ndarray, np.ndarray],
             if quantiles[0] > quantiles[1]:
                 raise ValueError('The quantile lower-bound is greater then the quantile upper-bound.')
             quantiles = np.clip(quantiles, a_min=0, a_max=1)
-            min_dist, max_dist = np.quantile(a=dist, q=quantiles)
+            min_dist, max_dist = np.quantile(a=dist, q=np.array(quantiles))
```
jklaise, line 406: I guess this is to do with not allowing type reassignment, with that enabled the typechecker should infer that `np.clip` returns an array and update the type accordingly (but we haven't considered enabling it yet).

I'm a bit torn whether it might be more useful to do the `type: ignore[arg-type]` here as technically we have done nothing wrong and `q` is a genuine array.

mauicv: It was the following error:

```
Argument "q" has incompatible type "Optional[Tuple[float, float]]"; expected "Union[_SupportsArray[dtype[Union[bool_, integer[Any], floating[Any]]]], _NestedSequence[_SupportsArray[dtype[Union[bool_, integer[Any], floating[Any]]]]], bool, int, float, _NestedSequence[Union[bool, int, float]]]"
```

My understanding was that quantile expects a numpy array so wrapping it like so fixed the error. That being said I was a bit confused as `clip` also expects a numpy array but that error only arose when I tested a simplified version:

```py
from typing import Optional, Tuple
import numpy as np

quantiles: Optional[Tuple[float, float]] = (0.1, 0.9)
quantiles = np.clip(quantiles, a_min=0, a_max=1)
min_dist, max_dist = np.quantile(a=np.array([]), q=quantiles)
```

The above gets a typing error for both the `clip` and `quantile` calls.

PR #695, opened 2022-06-09
GradientSimilarity explanations images.

RobertSamoilescu:
This PR adds two images in the method description for the GradientSimilarity.


jklaise (changes requested):
Looks good! A couple of minor suggestions:
1. The figure captions could be made more obvious that they're captions, e.g. by using _italics_
2. could you remove the `[source]` link at the top as it's not leading anywhere?

PR #698, opened 2022-06-13
Fix broken source links in docs, fix #675

### LINKED ISSUE #675: Remove or improve `[source]` links in method descriptions
Notices that the `[source]` links at the top of method descriptions are sometimes broken, don't exist, or if functional, lead to the implementation module which should be considered private. Proposing to either completely remove these links or linking to the relevant API documentation in the public API.

mauicv:
The Anchor doc source links are broken due to refactoring. This fixes them. The Top link points to the module folder and I've added ones for each anchor method next to their section on usage.


jklaise (commented):
Thanks. Can you check all the other pages as well (some might be missing a link or still broken) so we can close #675 ?

One thing to note is that ideally we would point all these links to the public API, e.g. to `https://docs.seldon.io/projects/alibi/en/latest/api/alibi.explainers.html#alibi.explainers.ALE` (sections of the `alibi.explainers` public API docs). Of course this is not always possible as in the case of Anchors where you want to point to a set of explainers not just one class.

mauicv:
> Thanks. Can you check all the other pages as well (some might be missing a link or still broken) so we can close #675 ?

The Anchor page was the only one I found with broken source links. I think this was introduced when I refactored `anchors.py` into a module. 


> One thing to note is that ideally we would point all these links to the public API, e.g. to `https://docs.seldon.io/projects/alibi/en/latest/api/alibi.explainers.html#alibi.explainers.ALE` (sections of the `alibi.explainers` public API docs). Of course this is not always possible as in the case of Anchors where you want to point to a set of explainers not just one class.

ok, noted.

I've fixed some other broken links I've found lying around too.

RobertSamoilescu:
Can you also fix the links of the following form:
```
:py:meth:`alibi.explainers.anchor_tabular.TabularSampler.get_feature_index`
```

RobertSamoilescu:
maybe replace pdf with abs [here](https://github.com/mauicv/alibi/blob/cf6db2b9db833a87962df61290fb0f6d2b242793/doc/source/overview/high_level.md?plain=1#L104)

RobertSamoilescu:
Screenshot from 2022-06-15 12-05-26

RobertSamoilescu:
Some other minor corrections:

1. ** 4 ** should be (4) ?  KernelShap notebook [here](https://seldon--698.org.readthedocs.build/projects/alibi/en/698/examples/kernel_shap_adult_lr.html#Grouping-features-with-KernelShap).
2.  Remove empty code cell at the end in [this notebook](https://seldon--698.org.readthedocs.build/projects/alibi/en/698/examples/overview.html).

doc/source/overview/high_level.md
```diff
@@ -101,7 +101,7 @@ appropriate. However, human bias itself is still an element. Hence, if the model
 data, we are vulnerable to using explainability to justify relations in the data that may not be accurate. Consider:
 > _"Before launching the model, risk analysts are asked to review the Shapley value explanations to ensure that the
 > model exhibits expected behavior (i.e., the model uses the same features that a human would for the same task)."_
-> &mdash; <cite>[Explainable Machine Learning in Deployment](https://dl.acm.org/doi/pdf/10.1145/3351095.3375624)</cite>
+> &mdash; <cite>[Explainable Machine Learning in Deployment](https://arxiv.org/abs/1909.06342)</cite>
```
jklaise, line 104: Worth using this link instead? https://dl.acm.org/doi/abs/10.1145/3351095.3375624

jklaise (approved):

PR #701, opened 2022-06-15
Remove incorrect empty anchor justification

jklaise:
The second point in the current version of the docs should not be possible because if it was (empty anchor returned and the prediction always changes, regardless of perturbation), the empty anchor would have precision ~0 and hence would not be returned in the first place. In fact, if the prediction does change a lot regardless of the perturbation, the data point is likely near the decision boundary and the algorithm will try longer and longer anchors to find something that satisfies the precision threshold.


RobertSamoilescu (approved):
Looks good to me!

Issue #687, opened 2022-05-30
`fetch_imagenet_10` not using `requests`


jklaise:
Noticed that `requests` isn't actually used (we use it over `urllib` mostly to handle exceptions): https://github.com/SeldonIO/alibi/blob/b3b742364a68527e96a380e9d03208352c85bff6/alibi/datasets/default.py#L32-L72

This needs to be rewritten to use `resp.contents` similar to other examples.

PR #713, opened 2022-07-06
Use `requests` in `fetch_imagenet_10` and encode data as a compressed tarball of `.npy` files.

gipster:
* Addresses issue https://github.com/SeldonIO/alibi/issues/687 .

* Changed storage file format from joblib to tar.gz

notes:
In order to load .npy files from a tar archive, the changed function uses the method described [here](https://github.com/numpy/numpy/issues/7989#issuecomment-340921579).


alibi/datasets/default.py
```diff
@@ -26,7 +25,7 @@
                        'http://www.cs.cornell.edu/People/pabo/movie-review-data/rt-polaritydata.tar.gz']
 
 #  TODO change storage format.
-IMAGENET_URLS = ['https://storage.googleapis.com/seldon-datasets/imagenet10/imagenet10.joblib']
+IMAGENET_URLS = ['https://storage.googleapis.com/seldon-datasets/imagenet10/imagenet10.tar.gz']
 
 
 def fetch_imagenet_10(url_id: int = 0) -> Dict:
```
jklaise, line 31: Please add a docstring for `url_id`, i.e. `Index specifying which URL to use for downloading`

alibi/datasets/default.py
```diff
@@ -67,7 +66,45 @@ def fetch_imagenet_10(url_id: int = 0) -> Dict:
     except RequestException:
         logger.exception("Could not connect, URL may be out of service")
         raise
-    imagenet10 = joblib.load(urllib.request.urlopen(url))
+    tar = tarfile.open(fileobj=BytesIO(resp.content), mode="r:gz")
+    imagenet10 = {}
+
+    int_to_str_labels = json.load(tar.extractfile('imagenet10/int_to_str_labels.json'))  # type: ignore
```
jklaise, line 72: Can you narrow down the `type: ignore` statements using the error codes shown please? I.e. `type: ignore[error-code]`.

alibi/datasets/default.py
```diff
@@ -67,7 +66,45 @@ def fetch_imagenet_10(url_id: int = 0) -> Dict:
     except RequestException:
         logger.exception("Could not connect, URL may be out of service")
         raise
-    imagenet10 = joblib.load(urllib.request.urlopen(url))
+    tar = tarfile.open(fileobj=BytesIO(resp.content), mode="r:gz")
+    imagenet10 = {}
```
jklaise, line 70: I would move this to bottom where the return dictionary is created for clarity. Or consider even building it in one go, i.e.
```python
imagenet10 = {
    'trainset': (X_train, y_train),
    ...
}
```

jklaise (changes requested):
Thanks Giovanni, couple of minor comments but looks good.

gipster:
Hey, just pushed the new version with the changes requested.

alibi/datasets/default.py
```diff
@@ -49,7 +48,7 @@ def fetch_imagenet_10(url_id: int = 0) -> Dict:
 
     Parameters
     ----------
-    url_id
+    url_id: Index specifying which URL to use for downloading.
```
jklaise, line 51: This is not the `numpy` docstring format, description should be on a new line with an indent and no colon. If in doubt, recreate a new docstring which will auto-format (assuming Pycharm is set up with the `numpy` style docstrings) with all the parameters included.

jklaise (approved):
Thanks, just the docstring format needs to be fixed to follow `numpy` docstring conventions and then good to merge.

Issue #709, opened 2022-06-28
Optional-dependency kaggle error


mauicv:
The optional dependency pattern doesn't protect against incompatible preinstalled versions of libraries. For instance (see #707), if you install Alibi in Kaggle, you get the following error when on import:

```
...
/opt/conda/lib/python3.7/site-packages/alibi/explainers/shap_wrappers.py in <module>
      8 import pandas as pd
      9 import shap
---> 10 import shap.utils._legacy as shap_utils
     11 from scipy import sparse
     12 from scipy.special import expit

ModuleNotFoundError: No module named 'shap.utils'
```

This is because Kaggle comes packaged with `shap==0.35.0` whereas alibi requires `shap>=0.40.0`. Thus in the `import_optional` function when we import `shap` we get no error because a version of shap exists. In the next line we try to import `shap_utils` from `shap.utils._legacy` and this file doesn't exist in `0.35.0`. Previously Alibi managed Shap as an optional dependency by trying to import it into `explianers.__init__` and only adding it to the `__all__` if it succeeded. The optional dependency functionality however imports it and only replaces it with the `MissingDependency` object if we receive an import error from the `import shap` statement. Because of Shaps presence in the environment, this error is not thrown however the module `shap.utils._legacy` doesn't exist so we get a `ModuleNotFoundError` instead!

___

As a quick fix you can either uninstall shap or install the correct versions:

```py
pip install numba>=0.50.0 shap==0.40.0 alibi
```

___

In terms of fixing the issue within Alibi, we do actually check for `ModuleNotFoundError` errors in `import_optional` but we try to match it to the specific library and I think the file module name is throwing it off...

mauicv:
Note: Kaggle has since updated its environment to have `shap==0.41.00` as default!

PR #714, opened 2022-07-07
[WIP] Check compatible optional dependency versions

mauicv:
# What is this:

This draft PR addresses #709 and adds checks for compatibility of optional dependencies in `import_optional`. Note that at the moment this functionality just checks that the correct submodels are present in the imported dependency. This solves the issue described in #709 where Shap throws an error on account of the version installed in Kaggle having an incompatible module structure.

# Todo:

- [x] Rewrite warning to reflect incompatible versions not just uninstalled dependency

# Open Questions:

Should we be checking the version compatibility explicitly instead?

### DIFF
--- a/alibi/utils/missing_optional_dependency.py
+++ b/alibi/utils/missing_optional_dependency.py
@@ -12,10 +12,10 @@ from string import Template
 from importlib import import_module
 
 err_msg_template = Template((
-    "Attempted to use $object_name without the correct optional dependencies installed. To install "
-    + "the correct optional dependencies, run `pip install alibi[$missing_dependency]` "
-    + "from the command line. For more information, check the Installation documentation "
-    + "at https://docs.seldon.io/projects/alibi/en/latest/overview/getting_started.html."
+    "Attempted to use $object_name without the correct optional dependencies installed. This may be due to missing"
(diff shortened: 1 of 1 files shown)


jklaise:
LGTM.

I don't think we should explicitly check versions as that would introduce a lot of extra complexity and upkeep, the version requirements are clearly defined in the relevant files and it is the job of the `pip` resolver to bring the right ones in (I'm surprised that `pip install alibi` on the Kaggle kernel didn't update the `shap` version to a compatible one - perhaps it's some peculiarity with the Kaggle environment?).

mauicv:
> LGTM.
> 
> I don't think we should explicitly check versions as that would introduce a lot of extra complexity and upkeep, the version requirements are clearly defined in the relevant files and it is the job of the `pip` resolver to bring the right ones in (I'm surprised that `pip install alibi` on the Kaggle kernel didn't update the `shap` version to a compatible one - perhaps it's some peculiarity with the Kaggle environment?).

Yeah agreed re maintenance. I assumed that running `pip install alibi[shap]` would suffice but for some reason, it didn't seem to fix the issue. Although, just took a look and since the original error it seems like `shap==0.41.00` is now the default in Kaggle!

PR #721, opened 2022-07-21
Partial dependence plots 

RobertSamoilescu:
Implementation of the partial dependence (PD) and individual conditional expectation (ICE) leveraging`sklearn` implementation.
Some functionalities that it includes
 * PD and ICE for numerical features
 * PD and ICE for categorical features
 * PD and ICE for combinations of numerical and/or categorical features
 * Plots for all the above cases
 * Custom grids
 * Usage of any black-box model (i.e. not only restricted to `sklearn` estimators)

TODOs:

- [x] Method description notebook
- [x] Example usage notebook


ascillitoe:
@RobertSamoilescu I've attempted to fix the problem with the docs CI failing. It was due to the $f_S(x_{S})$ equation in `PartialDependence.ipynb`. There were a number of minor issues:

- The use of `$$` surrounding the `\begin{align}`. Our combination of `nbsphinx` and `myst-parser` means amsmath environments such as `align` are automatically interpreted as latex math. If you also add `$$` the html docs build will be OK, but it will screw with our latex docs build.
- The use of `|...|` for abs. Sphinx uses `|...|` for some sort of substitution functionality (see [here](https://stackoverflow.com/questions/41269319/why-the-substitution-version-is-not-working-on-sphinx)), so with `|S|` it was treating `S` as a substitution. Solution is to use `\lvert` and `\rvert` instead.
- Intends have a specific meaning to `myst-parser`. We need to make sure we don't indent the content in math environments.

With the above changes I think the equation should render correctly for html and pdf builds, but probably good to check!

ascillitoe:
P.s. It seems there is still an issue with the hyperlink format I advised you to use. I'll look into this now.

ascillitoe:
> P.s. It seems there is still an issue with the hyperlink format I advised you to use. I'll look into this now.

@RobertSamoilescu you are doing everything correct here. i.e. creating a header anchor in `overview/high-level.md`:

```markdown
(partial-dependence)=
#### Partial Dependence
```

(note you don't have to do this for heading levels 1 to 3 since our config setting `myst_heading_anchors = 3` means anchors are generated automatically for these). You should be able to reference this like you are doing in `PartialDependence.ipynb`:

```
[Partial Dependence](../overview/high_level.md#partial-dependence) 
```

But... I'm afraid this will not work due to our combination of `nbsphinx` for parsing `.ipynb` files and then `myst-parser` for doing the final rendering. Long story short `myst-parser` looks for `../overview/high_level.md#partial-dependence`, but `nbsphinx` has already converted it to `../overview/high_level.html#partial-dependence`. This would be fixed if we transition from `nbsphinx` to [myst-nb](https://myst-nb.readthedocs.io/en/latest/) one day, or better yet stop writing methods docs as jupyter-notebooks (my strong preference!).

jklaise:
I agree with moving towards `.md` format with `myst` syntax entirely for docs. Would need a boring PR that replaces all the `.ipynb` docs files with the `.md` + `myst` equivalents but not sure if the effort at the moment is justified.

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+            )
+
+        # reshape averaged_predictions to (n_outputs, n_values_feature_0, n_values_feature_1, ...)
+        averaged_predictions = averaged_predictions.reshape(
+            -1, *[val.shape[0] for val in values]
+        )
+
+        # special case when dealing with a binary classifier that uses predict_proba. sklearn removes
+        # the computation for the label 0 and only returns the predictions for label 1. To make it consistent with
+        # the ALE, we re-add the predictions for label 0.
```
jklaise, line 477: Would be good to add if there are concrete examples of such classifiers?

alibi/explainers/tests/test_partial_dependence.py
```diff
@@ -0,0 +1,579 @@
...
+
+    # compute pd with sklearn
+    exp_sklearn = partial_dependence(X=X_train, estimator=rf, features=features, **params)
+
+    assert np.allclose(exp_alibi.pd_values[0], exp_sklearn['average'])
+    if isinstance(features[0], numbers.Integral):
+        assert np.allclose(exp_alibi.feature_values[0], exp_sklearn['values'][0])
+    else:
+        for i in range(len(exp_sklearn['values'])):
+            assert np.allclose(exp_alibi.feature_values[0][i], exp_sklearn['values'][i])
```
jklaise, line 530: Does this mean that `exp_sklearn` always returns a "flat" list of values, whereas we return it in "two levels", i.e. the indexing for us here is `[0][i]` whilst for `sklearn` it is just `[i]`.

alibi/explainers/tests/test_partial_dependence.py
```diff
@@ -0,0 +1,579 @@
...
+@pytest.mark.parametrize('params', [
+    {
+        'response_method': 'auto',
+        'percentiles': (0, 1),
+        'grid_resolution': np.inf,
+        'method': 'auto',
+        'kind': 'average'
+    }
+])
+def test_sklearn_categorical(rf_classifier, adult_data, features, params):
```
jklaise, line 547: So `sklearn` can compute categorical PDs correctly as long as `grid_resolution` is high enough?

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+    ----------
+    ice_values
+        Array of ice_values of dimension (N, V), where N is the number of instances in the reference dataset,
+        and V is the number of feature values where the PD is computed.
+    n_ice
+        See :py:meth:`alibi.explainers.partial_dependence.plot_pd`.
+    seed
+        Seed to be used for sampling.
+    """
+    np.random.seed(seed)
```
jklaise, line 824: I think we should try to move away from setting seeds inside library functions as this could lead to user applications unexpectedly having their seeds reset by us. Is it possible to move this out and into the examples for plotting (i.e. call `np.random.seed(seed)` before calling `plot_pd`)? @ascillitoe @mauicv tagging for visibility

jklaise: Another option would be to check how we could set a temporary seed that does not interact with the global `numpy` seed.

jklaise (changes requested):
@RobertSamoilescu this is 1/3 of the review covering the main implementation and tests. There will be a follow-up review on the example and write-up and the plotting functionality.

Nice job on the extensive testing btw! Most of my comments are discussion points or asking for some clarification. Perhaps the main thing is deciding what to do with `predictor_kw` as at the moment I don't think it's intuitive for a user to specify something like `prediction_fn` in case they pass a fully black-box model.

jklaise:
@RobertSamoilescu btw are the 3rd and 4th files here redundant?
image

jklaise:
@RobertSamoilescu should we try to add a `tqdm` progress bar around the for loop over `features_list`? Can't remember if there were some issues in certain Python environments with this...

jklaise:
@RobertSamoilescu I think it would be great if we could have another example, preferably classification to show off interpretation of multiple targets and black-box to show that it works the same way as `sklearn` models. This could be a follow-up PR.

doc/source/examples/pdp_regression_bike.ipynb
```diff
@@ -0,0 +1,645 @@
+{
```
jklaise, line 1: <p>Prerequisites</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

jklaise, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.    !wget https://raw.githubusercontent.com/christophM/interpretable-ml-book/master/data/bike.csv -q
</pre><p>Mmm tricky, since there is no <code>wget</code> on Windows (and Mac?) would need to exclude this notebook from tests.</p><p> </p><p>If we keep <code>wget</code> then should remove the file at the end of the notebook.</p><p> </p><p>Perhaps also worth implementing <code>fetch_bike</code>  but that would add some unnecessary complexity as currently most of our <code>fetch*</code> functions return processed data (e.g. see <code>fetch_adult</code>) ready for consumption of models.     </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

jklaise, line 1: <p>"includes both casual and registered users" ?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

jklaise, line 1: <p>"We use ordinal encoding for the..."</p><p> </p><p>"and as value the corresponding feature values"</p><p> </p><p>Noting to resolve the TODO link later.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

jklaise, line 1: <p>I'm getting this when running this cell:</p><p> </p><p> </p><pre class="ql-syntax" spellcheck="false">A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy

A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
</pre><p> </p><p>categorial_names -&gt; categorical_names</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

jklaise, line 1: <p>Would be useful to clarify the one-hot encoding as we just said we label-encoded it in the previous cell (but for a different reason).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jklaise, line 1: <p>Wouldn't it be easier to just include the pre-processor as part of the model?</p><p> </p><p>We could have a separate section where the data is already pre-processed and included this example.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

jklaise, line 1: <p>they returned -&gt; they return</p><p> </p><p>Maybe worth clarifying what "different probability distributions" means and how it effects the PD values? Perhaps there is a link to a <code>sklearn</code> doc page? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

jklaise, line 1: <p>I would suggest computation and visualization to be in separate cells as best practice, especially in cases where computation can take some time and here the user may want to tweak the figure parameters without recomputing.</p><p> </p><p>Also noting that the plot for season does not start at 0 unlike the first 3 plots which can be misleading (looks like a huge drop).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

jklaise, line 1: <p>"center the plots at 0 by subtracting..."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

jklaise, line 1: <p>Same comments as for the previous plot.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

jklaise, line 1: <p>This paragraph probably needs rewriting. Would be cool to motivate the need for 2-way plots here (can refer to Molnar's exposition).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

jklaise, line 1: <p>Same comments (primarily cell separation) as the above 2 plots.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

jklaise, line 1: <p>the rental drop -&gt; the rentals drop</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

doc/source/methods/PartialDependence.ipynb
```diff
@@ -0,0 +1,265 @@
+{
```
jklaise, line 1: <p>Perhaps it would be better to present usage and interpretation first (after a short intro paragraph) and then dive into the definitions and theory? We want people to be able to see quickly what the method can do and the outputs (plots) and roughly how to interpret it. E.g. see</p><p><a href="https://seldon--721.org.readthedocs.build/projects/alibi/en/721/methods/ALE.html" target="_blank">https://seldon--721.org.readthedocs.build/projects/alibi/en/721/methods/ALE.html</a></p><p><a href="https://scikit-learn.org/stable/modules/partial_dependence.html#partial-dependence" target="_blank">https://scikit-learn.org/stable/modules/partial_dependence.html#partial-dependence</a></p><p> </p><p>I would suggest the following (doesn't have to be exact titles/sections):</p><ol><li>Short intro paragraph (like the first one you already have, maybe a bit more, <code>sklearn</code>  page for inspiration)</li><li>Example figure with a high-level description (not as detailed as in the example notebook)</li><li>Usage and API</li><li>Theoretical exposition</li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

jklaise, line 1: <p>I'm think we should enforce some consistency between this page and the ALE one, in particular perhaps denoting the PD function with <code>PD</code> etc.? Keen to hear your thoughts. </p><p> </p><p>Monte Carol :))</p><p> </p><p>"we query the model on synthetic instance" -&gt; "we query the model on synthetic instances"</p><p>"Already that from the computation" -&gt; "Already from the computation"</p><p>"i.e. feature not correlated" -&gt; "i.e. features are not correlated"</p><p>"will create unrealistic synthetic instance" -&gt; "will create an unrealistic synthetic instance"</p><p>"that do not correlated" -&gt; "that are not correlated"</p><p>"hight" -&gt; "height"</p><p>"weight of child" -&gt; "weight of child"</p><p>"Such limitation can be addressed by ALE" -&gt; "This limitation can be addressed by the Accumulated Local Effects (ALE) method" - add link to ALE docs</p><p>"the PD still have" -&gt; "the PD still has"</p><p>"PD by setting all data instance" -&gt; "PD by setting all data instances"</p><p>"there exists some methods" -&gt; "there exist some methods"</p><p>"ALE to be extended to categorical feature, for" -&gt; "ALE to be extended to categorical features for"</p><p>"the&nbsp;<a href="https://christophm.github.io/interpretable-ml-book/ale.html" target="_blank">ALE chapter</a>&nbsp;from&nbsp;<a href="https://christophm.github.io/interpretable-ml-book/" target="_blank">Interpretable machine learning</a>&nbsp;book, C. Molnar(2020) as further reading." - need to rewrite, e.g. "from the Interpretable Machine Learning (C. Molnar, 2000) book" - can also look into using bibtex with <code>sphinx</code> (check <code>alibi-detect</code> overview of drift detection for how Ashley has done it)  </p><p> </p><p>"We will continue to discuss the simple case of linear regression." - I would replace this with a sentence to motivate this section, e.g. something like "We illustrate PD for the simple case of interpreting a linear regression model and show that it correctly results in showing a linear relationship between the features and the response."</p><p> </p><p>"assume the the features" -&gt; "assume the features"</p><p> </p><p>Comma at the end of the equation isn't needed since a new sentence is started. On that note probably good to include a finishing sentence that concludes what you set out to show (i.e. linear relationship).</p><p> </p><p>Link for Partial Dependence in the last sentence not working? Also worth pointing to the relevant ALE section too?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

jklaise, line 1: <p>First sentence, add ":" to the end?</p><p>"meta data" -&gt; "metadata"</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+        predictor
+            New `sklearn` estimator.
+        """
+        self.predictor = predictor
+
+
+# No type check due to the generic explanation object
+@no_type_check
+def plot_pd(exp: Explanation,
+            features_list: Union[List[int], Literal['all']] = 'all',
```
jklaise, line 609: Similar to above discussion, probably just rename to `features`

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+    Plot partial dependence curves on matplotlib axes.
+
+    Parameters
+    ----------
+    exp
+        An `Explanation` object produced by a call to the
+        :py:meth:`alibi.explainers.partial_dependence.PartialDependence.explain` method.
+    features_list
+        A list of features for which to plot the partial dependence curves or ``'all'`` for all features.
+        Can be an integers denoting feature index denoting entries in `exp.feature_names`. Defaults to ``'all'``.
```
jklaise, line 636: Sentence needs rewriting. Also probably need to clarify that the indexing also includes the 2-way tuples.

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+    target_idx
+        Target index for which to plot the partial dependence (PD) curves. Can be a mix of integers denoting target
```
jklaise, line 638: If can be strings then need the type hint updated.

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+        self.predictor = predictor
+
+
+# No type check due to the generic explanation object
+@no_type_check
+def plot_pd(exp: Explanation,
+            features_list: Union[List[int], Literal['all']] = 'all',
+            target_idx: int = 0,
+            n_cols: int = 3,
+            n_ice: Union[str, int, List[int]] = 'all',
```
jklaise, line 612: `str` -> `'Literal['all']'`

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+            centered: bool = True,
```
jklaise, line 613: Nitpick: would prefer to use the imperative as best practice, i.e. `center`.

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+         - a string taking the value ``'all'`` to display the ICE curves for every instance in the reference dataset.
+
+         - an integer for which `n_ice` instances from the reference dataset will be sampled uniformly at random to
+           display their ICE curves.
+
+         - a list of integers, where each integer represents an index of an instance in the reference dataset to
+           display their ICE curves.
+
+    centered
+        Boolean flag to center the individual conditional expectation (ICE) curves.
```
jklaise, line 654: Would be useful to add another sentence why this is useful to do.

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+            sharey: str = 'all',
+            pd_num_kw: Optional[dict] = None,
+            ice_num_kw: Optional[dict] = None,
+            pd_cat_kw: Optional[dict] = None,
+            ice_cat_kw: Optional[dict] = None,
+            pd_num_num_kw: Optional[dict] = None,
+            pd_num_cat_kw: Optional[dict] = None,
+            pd_cat_cat_kw: Optional[dict] = None,
+            fig_kw: Optional[dict] = None,
+            seed: Optional[int] = None) -> 'np.ndarray':
```
jklaise, line 625: As discussed above, let's move `seed` outside library functions.

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+    import matplotlib.pyplot as plt
+    from matplotlib.gridspec import GridSpec
+
+    default_fig_kw = {'tight_layout': 'tight'}
+    if fig_kw is None:
+        fig_kw = {}
+    fig_kw = {**default_fig_kw, **fig_kw}
+
+    if features_list == 'all':
+        features_list = range(0, len(exp.feature_names))
```
jklaise, line 695: To clarify - once the explanation is returned then `exp.feature_names` is actually a combination of the original `feature_names` and some 2-way feature name combinations (if any 2-way PDs are requested)?

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+        See :py:meth:`alibi.explainers.partial_dependence.plot_pd`.
+    seed
+        Seed to be used for sampling.
+    """
+    np.random.seed(seed)
+
+    if n_ice == 'all':
+        return ice_values
+
+    _, N = ice_values.shape
```
jklaise, line 829: Nitpick: in the docstring the shape is written as `(N, V)` but here we use the variable `N` for the second entry.

jklaise (changes requested):
This last review covers some minor comments related to the plotting functionality.

setup.py
```diff
@@ -55,7 +55,8 @@ def readme():
               'typing-extensions>=3.7.4.3',
               'dill>=0.3.0, <0.4.0',
               'transformers>=4.7.0, <5.0.0',
-              'tqdm>=4.28.1, <5.0.0'
+              'tqdm>=4.28.1, <5.0.0',
+              'seaborn>=0.11.2, <1.0.0',
```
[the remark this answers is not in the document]
RobertSamoilescu: the `seaborn` is required for the two way PDP for the categorical-categorical features.

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+        target_names
+            A list of target/output names used for displaying results.
+        predictor_kw
+            Predictor identifier arguments when the predictor is a callable prediction function. The following
+            arguments must be provided:
+
+             - ``'predictor_type'`` : ``str`` - Type of the predictor. Available
+                values: ``'regressor'`` | ``'classifier'``.
+
+             - ``'prediction_fn'`` : ``str`` - Name of the prediction function. Available value for
```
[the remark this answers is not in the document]
RobertSamoilescu: It is true that we just need to call the function, but unfortunately the `sklearn` backend treats them a bit different in some cases. For example, in the binary classification setting, the `sklearn` remove the the proabability corresponding to the class 0. On the other hand, we decided offline that we should include that too as it would be consistent with the ALE. Although for probability we have the 'liberty' to choose whether we include or not include the probability for class 0, for decision function there is only one option. For a binary classifier, there is a single decision function. Hence the special treatment based on the type of predictor (i.e., `decision_function` or `probability`). Happy to discuss alternatives offline.

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+            A `sklearn` estimator or a prediction function which receives as input a `numpy` array and outputs a
+            `numpy` array.
+        feature_names
+            A list of feature names used for displaying results.
+        categorical_names
+            Dictionary where keys are feature columns and values are the categories for the feature. Necessary to
+            identify the categorical features in the dataset.
+        target_names
+            A list of target/output names used for displaying results.
+        predictor_kw
```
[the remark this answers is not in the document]
RobertSamoilescu: That case is already considered when initializing the wrapper. Check code [here](https://github.com/RobertSamoilescu/alibi/blob/0511b945c658b55c3994f1861e996d26a3ab81c2/alibi/explainers/partial_dependence.py#L1254-L1259).

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+            An `N x F` reference tabular dataset used to calculate partial dependence curves. This is typically the
+            training dataset or a representative sample.
+        features_list
+            An optional list of features or pairs of features for which to calculate the partial dependence for.
+            If not provided, the partial dependence will be computed for every single features in the dataset.
+        response_method
+            Specifies the prediction function to be used. For a classifier it specifies whether to use the
+            `predict_proba` or the `decision_function`. For a regressor, the parameter is ignored. If set to `auto`,
+            the `predict_proba` is tried first, and if not supported then it reverts to `decision_function`. Note
+            that if `method='recursion'`, the prediction function always uses `decision_function`.
```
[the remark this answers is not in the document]
RobertSamoilescu: It doesn't relate only to `sklearn` estimators. Once a blackbox model is wrapped, there is no longer a difference between the `sklearn` models and the bb.

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+            returned for each individual from the dataset. Otherwise, if set to ``'both'``, then both the PD and
+            the ICE are returned. Note that for the faster ``method='recursion'`` option the only compatible parameter
+            value is ``kind='average'``. To plot the ICE, consider using the more computation intensive
+            ``method='brute'``.
+        percentiles
+            Lower and upper percentiles used to create extreme values which can potential remove outliers in low
+            density regions. The values must be in [0, 1].
+        grid_resolution
+            Number of equidistant points to split the range of each target feature. Only applies if the number of
+            unique values of a target feature in the reference dataset `X` is less than the `grid_resolution` value.
```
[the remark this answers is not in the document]
RobertSamoilescu: you are correct.

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+            raise ValueError('The array X must be 2-dimensional.')
+        n_features = X.shape[1]
+
+        # set the features_names when the user did not provide the feature names
+        if self.feature_names is None:
+            self.feature_names = [f'f_{i}' for i in range(n_features)]
+
+        # set categorical_names when the user did not provide the category mapping
+        if self.categorical_names is None:
+            self.categorical_names = {}
```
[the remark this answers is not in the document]
RobertSamoilescu: yes.

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+
+        if not np.all(np.isin(list(grid_points.keys()),  np.arange(n_features))):
+            raise ValueError('The features provided in grid_points are not a subset of the dataset features.')
+
+        for f in grid_points:
+            if self._is_numerical(f):
+                grid_points[f] = np.sort(grid_points[f])
+
+            else:
+                grid_points[f] = np.unique(grid_points[f])
```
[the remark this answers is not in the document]
RobertSamoilescu: Can you elaborate on this?

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+                message = "The grid points provided for the categorical feature {} are invalid. "\
+                          "For categorical features, the grid points must be a subset of the features "\
+                          "values defined in categorical_names. Received an unknown value of '{}'."
+
+                # convert to label encoding if the grid is provided as strings
+                if grid_points[f].dtype.type is np.str_:
```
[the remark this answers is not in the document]
RobertSamoilescu: Won't fail as the `np.unique` transforms it into a `np.ndarray`. See line [here](https://github.com/RobertSamoilescu/alibi/blob/0511b945c658b55c3994f1861e996d26a3ab81c2/alibi/explainers/partial_dependence.py#L264).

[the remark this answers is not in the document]
RobertSamoilescu: Same is valid for numerica features where `np.sort` transforms them into `np.ndarray`. Probably should extend that grid points can be lists too.

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+                    raise ValueError(f'Current implementation of the partial dependence supports a maximum of two '
+                                     f'features at a time when a tuple is passed. Received {len(f)} features with the '
+                                     f'values {f}.')
+
+                check_feature(f[0])
+                check_feature(f[1])
+            else:
+                check_feature(f)
+
+    def _params_sanity_checks(self,
```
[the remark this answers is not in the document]
RobertSamoilescu: Let's discuss this offline.

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+            else:
+                values_f = [grid_points[f_indices[0]]]
+
+            features_indices.append(f_indices)
+            deciles.append(deciles_f)
+            values += values_f
+
+        # covers also the case of a single feature, just to ensure it has the right shape
+        features_indices = np.concatenate(features_indices, axis=0)
+        grid = cartesian(tuple([v.reshape(-1) for v in values]))
```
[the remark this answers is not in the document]
RobertSamoilescu: This is where we perform the cartesian product between the feature values when we have to compute 2-way pd. Probably the comment is a bit unclear, but just wanted to mention that it works also for 1-way pd.

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+            # get deciles for the current feature if the feature is numerical
+            deciles_f = get_quantiles(X_f, num_quantiles=11) if self._is_numerical(f) else None
+
+            if f not in grid_points:
+                # construct grid for feature f. Note that for categorical features we pass the
+                # grid resolution to be infinity because otherwise we risk to apply `linspace` to
+                # categorical values, which does not make sense.
+                _, values_f = _grid_from_X(X_f,
+                                           percentiles=percentiles,
+                                           grid_resolution=grid_resolution if self._is_numerical(f) else np.inf)
```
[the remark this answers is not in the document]
RobertSamoilescu: yes. This is the logic that allows us to control the grid resolution for numerical features and also treat correctly the categorical feature case. For example, consider that we want to compute the 2-way pdp for a numerical and a categorical feature. Now lets say that the numerical feature has 1000 feature values and the categorical one has 15 feature values. If we were to use directly the sklearn API and set the grid points to something less than 15, we would have ran into some problems. This is because the backend would have generated a grid point for the categorical values too. Instead of leaving them untouched (i.e. [0, 1, 2, ... , 14]) would have generated some float numbers (e.g. [0, 2.5, 7.5, 12.5]), which is non-sense.

alibi/explainers/tests/test_partial_dependence.py
```diff
@@ -0,0 +1,579 @@
...
+    assert len(exp.ice_values) == len(features)
+    assert len(exp.feature_values) == len(features)
+
+    for i, f in enumerate(features):
+        if isinstance(f, Tuple):
+            # check deciles
+            assert isinstance(exp.feature_deciles[i], List)
+            assert len(exp.feature_deciles[i]) == len(f)
+            assert len(exp.feature_deciles[i][0]) == 11
+            assert len(exp.feature_deciles[i][1]) == 11
```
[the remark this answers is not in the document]
RobertSamoilescu: Probably. Also we will lose some consistency with the ALE as far as I remember.

alibi/explainers/partial_dependence.py
RobertSamoilescu: Not sure I follow ... This happens for all classifiers that support `predict_proba` -- lots of classifiers. Can you elaborate?

alibi/explainers/tests/test_partial_dependence.py
RobertSamoilescu: The `partial_dependence` function from `sklearn` supports the computation of a single N-way PD. In our API, we allow for multiple explanation meaning that we are able to call the explain method for multiple N-way PDs. More concretely, we can call the explain on `[0, 1, (0, 1)]`. The `partial_dependence` function in `sklearn` does not allow that and we are required to make 3 calls: `partial_dependence([0])`, `partial_dependence([1])` and `partial_dependence([0, 1])`. Hope this explains why our explanation is a list and the `sklearn` explanation is not.

alibi/explainers/tests/test_partial_dependence.py
RobertSamoilescu: Yes. If you wonder why we can not leverage more directly the `sklearn` API, please see the comment above about the 2-way PD for categorical-numerical.

doc/source/examples/pdp_regression_bike.ipynb
RobertSamoilescu: <p>I would add the <code>fetch_bike</code> in another PR and update the notebook. </p>

RobertSamoilescu: <p>yes, it doesn't start at 0 because the <code>sharey="row"</code>. I could set it to <code>"all"</code> but then you would not see any drop because it is quite small. For me, as long the y axis shows the values I can understand it easily, but if you think that this is still confusing after clarifying why the second row doesn't start at 0, I will change it to <code>"all"</code>.      </p><p>  </p>

doc/source/methods/PartialDependence.ipynb
RobertSamoilescu: <p>Can you write the links to the ALE sections as I am not quite sure which ones you are referring to?</p>

RobertSamoilescu: <p>Modified the links. Should work after the next release.</p><p> </p>

setup.py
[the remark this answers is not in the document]
jklaise: OK would be keen to explore if we can get around having to depend on `seaborn` (i.e. can we make something comparable with `matplotlib`?).

alibi/explainers/partial_dependence.py
RobertSamoilescu: yes. For example `exp.feature_names = ['temp', ('temp', 'windspeed')]`.

alibi/api/defaults.py
```diff
@@ -283,3 +283,35 @@
...
+    "feature_values": None,
+    "feature_names": None,
+    "meta": {
+        "feature_names": None,
+        "categorical_names": None,
+        "target_names": None,
+        "response_method": None,
+        "method": None,
+        "kind": None,
+    }
```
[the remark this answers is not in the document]
RobertSamoilescu: Removed it.

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1294 @@
...
+        Parameters
+        ----------
+        feature
+            Feature to be checked.
+
+        Returns
+        -------
+        ``True`` if the feature is numerical. ``False`` otherwise.
+        """
+        return (self.categorical_names is None) or (feature not in self.categorical_names)
```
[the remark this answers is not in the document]
RobertSamoilescu: Not necessary as `self.categorical_names` is initialised as an empty dict.

doc/source/examples/pdp_regression_bike.ipynb
RobertSamoilescu: <p>Used Pipeline</p>

doc/source/methods/PartialDependence.ipynb
RobertSamoilescu: <p>Hope now it looks better</p><p> </p>

doc/source/examples/pdp_regression_bike.ipynb
```diff
@@ -0,0 +1,686 @@
...
...Hbx73OYOfuT7TbQkrJuLlJfTs1OCKz68829H4/QVeKvFwrVfXGGMyn+jWs+VrmURx4ntVZ+L/fM/VYXgk/+ufpmxIXVeH4ZFOnN9+xd8b96rSK9fJ498H/3bb76m1QNLD2GzJeHkV3zKJPMtmIuGoyBRqlK1SiMEopZRSnscYk+uHO9Nk2MPYbcl4eed8ajVj7EghZcMF3Stc0KLPpxAS5t4zdaiCV99rxuV3UkqpYsx+BQ93psmwh7EbGxZLzgd52e12JJNphIqtbD6dxkbbCAzKZT22KlIKYgEOpZQqanTRDeVSxpgsB4Blvr8dpGgkw/aUFMRacPeSmGDH319nk1BKKaWyo1OrKRczSC6S29wmz+4sOT4Ba8CVrz4HZFszbAxYLEXjvVJKKaUKirvXAOdW0egyLEaMMbkbEGdMtgmgJ0mOj8PqX3BLURexv9sqD7RuWCmlsmbH5PrhzjQZ9ji57BmmcHqGC3p+YYCU+IQCTYaVAq0bVkqpy9GaYeVSxthzlwzb7RTW3GoFPZPEgfhdeJXwv/yO2dHuX6UKjO9VnV0dgsfyv75ozo1fGE6c3+7qEIodexH7t1STYQ+TlBSHv39gjvdPTk7E6l00pguLO3iC1n06XvHxdpsdyaYmuIhUkyjlMolb5l1+J5WBb/0uxE94ydVheCT/O0YSXKK6q8PwSFGxe6/42KKVCmsy7HGMsedqarWkpHisPnkcdHYZhVEiAZAcGU1AyZArPz42Dt/ArMssitgHXaWUUqpAuHsNcG5pzXARl5SUgDUXi3RcKU9YbCMxKha/UK05VkoppdRF2jNcxG2NO0mViPIFdv7C6hW2Jacg1rzNAZwYFYtfcNa95FomoZRSSl1eUesZ1mS4iLMlJeLtV7BlEoXRK3z+0CECqpTN0zkkfi11KlTMp4iUUkqp4knnGVYucyU/fLbkRLx9CzYZLgw796+kccsGeTpH/Nl4gsIDstxexP5uK6WUUgVC5xlWLpOUkoiXV+5qXgsyGV5nThRarXDSqXOUKFsyT+eIORVLSNmgTLfZbXYs+rdBOVXwjygyC2+ISCUR+UNEdonIHhH5WER8nNtaishCEdkhIutE5FsReURE1jsfSSKyyfl8lCvit9ns3DzsLR5984t07aO+nUyr259KfX305Bnuf+VjbnjqDe596UOOnz6Xuu3YqbM8+Non9Hv
```
jklaise, line 669: This is causing issues (see the failing notebook execution test). It's the usual thing where the notebook throws an error because the environment `alibi` doesn't exist. It should be changed to `python3` or `ipykernel` (check other notebooks for examples - let's add this to Notion/CONTRIBUTING.md in a follow-up).

alibi/explainers/partial_dependence.py
```diff
@@ -103,62 +120,75 @@ def __init__(self,
...
+            PD estimates are less reliable in those extreme regions. The values must be in [0, 1]. Only used
+            with `grid_resolution`.
         grid_resolution
             Number of equidistant points to split the range of each target feature. Only applies if the number of
-            unique values of a target feature in the reference dataset `X` is less than the `grid_resolution` value.
+            unique values of a target feature in the reference dataset `X` is greater than the `grid_resolution` value.
+            For example, consider a case where a feature can take the following values:
+            ``[0.1, 0.3, 0.35, 0.351, 0.4, 0.41, 0.44, ..., 0.5, 0.54, 0.56, 0.6, 0.65, 0.7, 0.9]``, and we are not
+            interested in evaluating the marginal effect at every single point as it can become computationally costly
+            (assume hundreds/thousands of point) without providing any additional information for nearby points
```
jklaise, line 184: thousands of point -> thousands of points

alibi/explainers/partial_dependence.py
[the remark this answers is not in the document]
jklaise: What I meant was that there's a philosophical choice here - we either trust the users fully and expect them to pass correctly formatted arguments, however this leads to indecipherable errors when the arguments are malformed. On the other hand, we can do these sanity checks like sorting/taking unique values to ensure things go smoothly most of the the time with a (very small) performance hit.

I think the second one is preferable which is in the spirit of `sklearn` and shouldn't introduce a lot of overhead in most cases. Even if we wanted to validate and raise errors/warning on misformatted inputs it's likely true that the extra work for raising errors would already be done, so we may as well do the formatting internally.

doc/source/methods/PartialDependence.ipynb
jklaise: <p>Looks great, thanks!</p>

doc/source/examples/pdp_regression_bike.ipynb
jklaise: <p>Still to fix?</p>

RobertSamoilescu: <p>should be fixed already</p>

jklaise: I think there's a missing `e`

RobertSamoilescu: yeah ... I fixed it now ...

alibi/explainers/partial_dependence.py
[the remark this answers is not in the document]
jklaise: This is true, but I think we need more documentation about this, perhaps at the level of `predictor_kw` in the constructor and then cross-referencing between the `explain` and the constructor, as it's not fully clear in the black-box case.

This is similar to the previous discussion of having to specify `predictor_kw` like `prediction_fn` for black-box cases where it can be confusing what people are asked to do.

alibi/explainers/partial_dependence.py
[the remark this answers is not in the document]
jklaise: Let's leave arrays, less room for trouble.

alibi/explainers/partial_dependence.py
jklaise: Ah I understand, so any 2-class classifier with `predict_proba` has this behaviour.

alibi/explainers/partial_dependence.py
RobertSamoilescu: I've included an explanation.

alibi/explainers/partial_dependence.py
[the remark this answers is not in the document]
RobertSamoilescu: I've already modified it. Let me know if I should revert it. I think it's fine either way. Your call.

alibi/explainers/partial_dependence.py
```diff
@@ -676,26 +689,38 @@ def plot_pd(exp: Explanation,
...
+        two categorical features.
     fig_kw
-        Keyword arguments passed to the `fig.set` function.
+        Keyword arguments passed to the `matplotlib.figure.set`_ function.
+
+        .. _matplotlib.pyplot.plot:
+            https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.plot.html
+
+        .. _matplotlib.figure.set:
+            https://matplotlib.org/stable/api/figure_api.html
```
jklaise, line 719: This is not showing in hyperlinks, probably because the references are missing a `_`.

doc/source/examples/pdp_regression_bike.ipynb
```diff
@@ -0,0 +1,675 @@
+{
```
jklaise, line 1: <p>Btw <code>pandas</code> is currently a core dependency so no need to mention this or do the <code>!pip ...</code> call.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/721/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

alibi/explainers/partial_dependence.py
```diff
@@ -1216,13 +1241,14 @@ def _plot_two_pd_cat_cat(exp: Explanation,
     # plot heatmap
     default_pd_cat_cat_kw = {
         'annot': True,
-        'fmt': '.2f',
-        'linewidths': .5,
+        'fmt': '{x:.2f}',
```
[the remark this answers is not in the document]
jklaise: My bad, accidentally read this as an f-string.

setup.py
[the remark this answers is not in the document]
jklaise: Note for posterity - `pd_cat_cat_kw` are now passed to our heatmap wrapper function living in `alibi.utils.visualization.heatmap` so this issue is mitigated at the expense of the user having to look at our visualization API.

doc/source/examples/pdp_regression_bike.ipynb
jklaise: Resolving as the notebook now reads the csv file directly using `pandas`.

jklaise: <p>On second thought this is ok, I would just change the start of the sentence to "We continue with some examples...".</p>

alibi/explainers/partial_dependence.py
```diff
@@ -66,27 +77,36 @@ def __init__(self,
...
-                in analogy with the `sklearn` estimators API and the `response_method` used in
-                :py:meth:`alibi.explainers.partial_dependence.explain`.
+             - ``'prediction_fn'`` : ``str`` - Name of the prediction function. \
+             Available value for regression: ``'predict'``. \
+             Available values for classification: ``'predict_proba'`` | ``'decision_function'``. \
+             The choice should be considered in analogy with the `sklearn` estimators API. If ``'predict_proba'``
+             is used, the `predictor` should output an array of size `N x C`, where `N` is the number of instances
+             to compute the prediction for and `C` is the number of classes. Note that this setting includes the
+             binary classifier problem where `C` should be 2 (i.e., two columns corresponding to the negative
+             and positive class).
```
jklaise, line 100: When should `decision_function` be specified?

alibi/explainers/partial_dependence.py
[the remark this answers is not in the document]
jklaise: As discussed offline, any mistakes in wrapper would be caught in the tests `test_regression_wrapper` and `test_classification_wrapper`.

alibi/explainers/partial_dependence.py
```diff
@@ -0,0 +1,1340 @@
...
+        -------
+        Update parameters `(response_method, method, kind)`.
+        """
+        check_is_fitted(estimator)
+
+        if not (is_classifier(estimator) or is_regressor(estimator)):
+            raise ValueError('The predictor must be a fitted regressor or a fitted classifier.')
+
+        if is_classifier(estimator) and isinstance(estimator.classes_[0], np.ndarray):
+            raise ValueError('Multiclass-multioutput predictors are not supported.')
```
jklaise, line 389: Would like to see these in a separate validation method to be called in the constructor so we can fail quickly if the estimator isn't fitted/supported.

doc/source/examples/pdp_regression_bike.ipynb
jklaise, line 1: Please commit the generated symlink.

jklaise:
After an offline discussion we decided to copy the private methods from `sklearn` to compute PD using `brute` approach and use that directly on black-box models without wrapping into a `sklearn`-like wrapper. This allows us to get rid of the confusing `predictor_kw` kwarg making it easier and more transparent to the end-user. It also has the added benefit of us being more in control due to not relying on private `sklearn` functions.

On the flip side, this means for now we only support `numpy` arrays for both black-box and `sklearn` models (contrast with `sklearn` implementation which allows estimators to be fitted on `pandas` dataframes, sparse matrices etc. This is in line with our current approach of mainly supporting `numpy` array fitted models.

RobertSamoilescu:
Decided to remove the options `response_method='auto'`. This is because for a binary classifier the number of output targets can change based on the other parameters (e.g., `method='recursive'` and `kind='average'` vs `method='brute'` and `kind='both'`). If `method='recursive'` and `kind='average'` are set, then the decision function is used to compute the PD, which means that for a binary classifier the output will have just one column. On the other hand, if `method='brute'` and `kind='both'`, then `predict_proba` will be used instead, which in our implementation will result in two output columns.  The problem arises when plotting since the plots can use the `target_names` specified in the constuctor. For example, if `target_names=['output']`, then a plotting error will arise when calling the `explain` method with the second pair of parameters since in that case the output has 2 columns, but the user specified only one target name. Another failure can arrise if `target_names=[class_0, class_1]`. If the first pair of parameters is used, then the PD corresponding to the decison score will be labeled with `class_0'` which is not correct.

Also I decided to move the parameter `response_method` into the `__init__` since will fully specify the function to be used for the PD computation. Thus, we will avoid the plotting errors above. If the user wants to use anther function, they will have to create a new explainer object.

RobertSamoilescu:
Also decided to remove `method='auto'` option. The `sklearn` logic is the following:
```python
if method == Method.AUTO:
    if isinstance(self.predictor, BaseGradientBoosting) and self.predictor.init is None:
        method = Method.RECURSION.value
    elif isinstance(self.predictor, (BaseHistGradientBoosting, DecisionTreeRegressor, RandomForestRegressor)):
        method = Method.RECURSION.value
    else:
        method = Method.BRUTE.value

if method == Method.RECURSION:
    if not isinstance(self.predictor, (BaseGradientBoosting, BaseHistGradientBoosting, DecisionTreeRegressor,
                                       RandomForestRegressor)):
        supported_classes_recursion = (
            "GradientBoostingClassifier",
            "GradientBoostingRegressor",
            "HistGradientBoostingClassifier",
            "HistGradientBoostingRegressor",
            "HistGradientBoostingRegressor",
            "DecisionTreeRegressor",
            "RandomForestRegressor",
        )
        raise ValueError(f"Only the following estimators support the 'recursion' "
                         f"method: {supported_classes_recursion}. Try using method='{Method.BRUTE.value}'.")

    if response_method == ResponseMethod.AUTO:
        response_method = ResponseMethod.DECISION_FUNCTION.value

    if response_method != ResponseMethod.DECISION_FUNCTION:
        raise ValueError(f"With the '{method.RECURSION.value}' method, the response_method must be "
                         f"'{response_method.DECISION_FUNCTION.value}'. Got {response_method}.")
```
 
Removing the `method='auto'` option is a consequence of removing `response_method='auto'`.

Consider we have a model that supports both `decision_function` and `predict_proba` . Furthermore, assume it is a model that supports recursion method. Then we can have at least the following cases:

1) `kind='average'`, `response_method='decision_function'`, `method='auto'`
In this case, `method` will become `method='recursion'` because the model supports recursion option (first set of if). **Everything works well**.

2) `kind='average'`, `response_method='predict_proba'`, `method='auto'`
In this case, `method` will also become `method=recursion` because the model supports recursion (first set of if). But after that, **a value error will be thrown** because the `response_method` cannot be set or was not set to `decision_function`. The error would not have been thrown if `response_method='auto'` cause in that case `response_method` would have become `'decision_function'`. But since that option is no longer available, IMO it doesn't make sense to keep the option `method='auto'`.

alibi/explainers/partial_dependence.py
```diff
@@ -242,13 +242,21 @@ def explain(self,  # type: ignore[override]
...
             # set the `target_names` when the user did not provide the target names
             # we do it here to avoid checking model's type, prediction function etc.
-            key = Kind.AVERAGE if kind in [Kind.AVERAGE, Kind.BOTH] else Kind.INDIVIDUAL
-            n_targets = pds[0][key].shape[0]
             self.target_names = [f'c_{i}' for i in range(n_targets)]
 
+        elif len(self.target_names) != n_targets:
+            logger.warning('The length of `target_names` does not match the number of predicted outputs. '
+                           'Ensure that the lengths match, otherwise a call to the `plot_pd` method might '
+                           'raise an error or produce undesired labeling.')
```
RobertSamoilescu, line 257: I don't think there is a possibility to check whether the number of targets that we computed the PD/ICE for matches the length of `target_names` without a dummy call. Altough the user should be aware of that I decided to include a warning message in case that happens. Shall we go even further and raise an error?

jklaise:
Following an offline discussion it was decided to simplify the implementation and user interface by splitting the implementation into two distinct classes:
 - `PartialDependence` for use with black-box models calculating PD using a brute-force approach
 - `TreePartialDependence` for use with white-box models (currently only a small selection of `sklearn` estimators) that support a recursive algorithm for calculating PD which is faster than the brute-force approach

This allows us to remove all of the slightly confusing arguments discussed previously.

Also, @RobertSamoilescu checked that the recursive algorithm returns slightly different values than performing the brute-force PD on the same estimators which further justifies splitting the implementation into two public classes (similar to `KernelShap` and `TreeShap`.

RobertSamoilescu:
@jklaise, check the note at the end of this [section](https://scikit-learn.org/stable/modules/partial_dependence.html#computation-methods) which confirms that the two methods differ in the values they return

alibi/explainers/partial_dependence.py
```diff
@@ -874,46 +1138,114 @@ def _sample_ice(ice_values: np.ndarray, n_ice: Union[Literal['all'], int, List[i
...
+        ice_values = _sample_ice(ice_values=ice_values, n_ice=n_ice)  # type: ignore[arg-type]
+
+        # center ice values if necessary
+        if center:
+            ice_values = ice_values - ice_values[0:1, :]
+
+    return pd_values, ice_values
+
+
+def _compute_pd_limits(exp: Explanation,
```
jklaise, line 1201: `exp` is not used in the function body, not sure why `mypy` misses this...

alibi/explainers/partial_dependence.py
```diff
@@ -874,46 +1138,114 @@ def _sample_ice(ice_values: np.ndarray, n_ice: Union[Literal['all'], int, List[i
...
                      target_idx: int,
-                     center: bool = True,
-                     n_ice: Union[Literal['all'], int, List[int]] = 'all',
+                     center: bool = False,
+                     pd_limits: Optional[Tuple[float, float]] = None,
+                     n_ice: Union[Literal['all'], int, List[int]] = 100,
                      ax: Optional['plt.Axes'] = None,
                      pd_num_kw: Optional[dict] = None,
-                     ice_num_kw: Optional[dict] = None) -> 'plt.Axes':
+                     ice_num_kw: Optional[dict] = None) -> Tuple['plt.Axes', Optional[Tuple[float, float]]]:
```
jklaise, line 1238: Worth updating the docstrings for all these internal functions that now return more things.

jklaise (approved):
Nice work, looks great!

PR #728, opened 2022-07-29
Use pytest-custom_exit_code to not flag empty notebook test suite as …

### LINKED ISSUE #647: Suppress notebook CI test failure when all tests are skipped
In some circumstances, all the notebook CI tests are skipped. This causes a CI failure to be returned despite there being no problem with the notebook(s). For example, if a notebook excluded from some builds is changed, the [test_changed_notebooks](https://github.com/SeldonIO/alibi/blob/41db95e94d37d29a9da192cc72673f5bc15ccef1/.github/workflows/test_changed_notebooks.yml) CI is run, but then all tests are skipped on some builds, as is the case [here](https://github.com/SeldonIO/alibi/runs/6226721622?check_suite_focus=true).

We should prevent the CI from being flagged as failing when `pytest` exits with code 5 (indicating all tests skipped). The [pytest documentation](https://docs.pytest.org/en/7.1.x/reference/exit-codes.html) recommends using the [pytest-custom_exit_code](https://github.com/yashtodi94/pytest-custom_exit_code) plugin for this.

jklaise:
…a CI failure

Resolves #647. @ascillitoe if happy with implementation I'll open an identical PR for alibi-detect.

### DIFF
--- a/.github/workflows/test_all_notebooks.yml
+++ b/.github/workflows/test_all_notebooks.yml
@@ -22,9 +22,9 @@ jobs:
         os: [ubuntu-latest]
         python-version: [ '3.7', '3.8', '3.9', '3.10' ]
         include: # Run macos and windows tests on only one python version
-          - os: windows-latest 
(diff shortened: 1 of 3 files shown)


ascillitoe (approved):
LGTM!

PR #731, opened 2022-08-01
ALE custom grid points update.

RobertSamoilescu:
Changes behavior from:
```
Grid points outside the feature range. Consider the following example: `O O O X X O X O X O O`, \
where 3 grid-points are smaller than the minimum value in `f`, and 2 grid-points are larger than the maximum \
value in `f`. Grid-points outside the feature value range are clipped between the minimum and maximum \
values of `f`. The grid-points considered will be: `(O|X) X O X O (X|O)`.
```

to:
```
Grid points outside the feature range. Consider the following example: `O O O X X O X O X O O`, \
where 3 grid-points are smaller than the minimum value in `f`, and 2 grid-points are larger than the maximum \
value in `f`. The empty leading and ending bins are removed. The grid-points considered
will be: `O X X O X O X O`.
```

Note that the new approach is less strict in terms of the lower & upper bound of the grid. This further allows
custom lower/upper grid values to be considered (previously the grid points were clamped to the min/max feature values) for which the ALE is still well defined.

The PR also includes a stress test for custom grid points.

### DIFF
--- a/alibi/explainers/ale.py
+++ b/alibi/explainers/ale.py
@@ -138,8 +138,8 @@ class ALE(Explainer):
 
          - Grid points outside the feature range. Consider the following example: `O O O X X O X O X O O`, \
         where 3 grid-points are smaller than the minimum value in `f`, and 2 grid-points are larger than the maximum \
-        value in `f`. Grid-points outside the feature value range are clipped between the minimum and maximum \
-        values of `f`. The grid-points considered will be: `(O|X) X O X O (X|O)`.
+        value in `f`. The empty leading and ending bins ware removed. The grid-points considered
+        will be: `O X X O X O X O`.
 
          - Grid points that do not cover the entire feature range. Consider the following example: \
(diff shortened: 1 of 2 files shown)


alibi/explainers/ale.py
```diff
@@ -138,8 +139,8 @@ def explain(self,
 
          - Grid points outside the feature range. Consider the following example: `O O O X X O X O X O O`, \
         where 3 grid-points are smaller than the minimum value in `f`, and 2 grid-points are larger than the maximum \
-        value in `f`. Grid-points outside the feature value range are clipped between the minimum and maximum \
-        values of `f`. The grid-points considered will be: `(O|X) X O X O (X|O)`.
+        value in `f`. The empty leading and ending bins were removed. The grid-points considered
```
jklaise, line 142: were -> are

jklaise (approved):
Nice!

Issue #317, opened 2020-11-03
IndexError: index -1 is out of bounds for axis 0 with size 0


charlie9526:
```
~/.local/lib/python3.6/site-packages/alibi/explainers/anchor_tabular.py in perturbation(self, anchor, num_samples)
    278         )
    279         nb_partial_anchors = np.array([len(n_records) for n_records in reversed(partial_anchor_rows)])
--> 280         coverage = nb_partial_anchors[-1] / self.n_records
    281 
    282         # if there are enough train records containing the anchor, replace the original records and return...

IndexError: index -1 is out of bounds for axis 0 with size 0
```

jklaise:
Looks like the the variable `nb_partial_anchors` is empty, this could theoretically happen if there are no rows in the training set where the proposed anchor holds. This could be a bug on our side as we should catch that error if that is what is happening.

Can you run the code in a debugger and print out the variable values? Probably starting with `allowed_rows` in this method: https://github.com/SeldonIO/alibi/blob/ce4d695285236fe1b638446bedcabed5e8122d1f/alibi/explainers/anchor_tabular.py#L237-L306

charlie9526:
Thank you for the quick reply. In the current situation, I am not using a debugger. I will upload them later.
Thank you!

enricorotundo:
Fixed in `alibi>=0.5.6` ?

https://github.com/SeldonIO/alibi/blob/9af1d73d046a2239ce73c12c2e1f835fecc89780/alibi/explainers/anchor_tabular.py#L281

jklaise:
@enricorotundo I don't think so because you would get the same error regardless if the index was 0 or -1.

I think the issue may originate because the list is empty (meaning `partial_anchor_rows` is empty). Is this something you've also experienced?

enricorotundo:
Yes but I've workaround it by tweaking my dataset a bit so I get different anchors and it wouldn't fail.

jklaise:
@enricorotundo can you elaborate a little bit? Maybe we can get to the bottom of this and implement appropriate handling.

enricorotundo:
@jklaise I'm working with a customer dataset using AnchorTabular via Seldon so I'm not so sure how to provide a reproducible example. Anyhow, replacing `AnchorTabular(clf.predict(x))` with `AnchorTabular(clf.predict_proba(x))` magically works.

jklaise:
[...]

PR #732, opened 2022-08-02
Add `NotFittedError` and use it in `AnchorTabular`

jklaise:
Whilst debugging #317 a few times I called the `explain` method without having called the `fit` method first which resulted in a cryptic stack trace:

```
  File "/opt/pycharm-professional/plugins/python/helpers/pydev/pydevd.py", line 1483, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "/opt/pycharm-professional/plugins/python/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/home/janis/PycharmProjects/public-fork-alibi/dev_notebooks/#317-minimal.py", line 37, in <module>
    explanation = explainer.explain(bad_instance)  # breaks here
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/anchors/anchor_tabular.py", line 815, in explain
    self.instance_label = self.samplers[0].instance_label
IndexError: list index out of range
```

This PR introduces a `NotFittedError` and a private `_fitted` attribute to `AnchorTabular` to determine when to raise this error resulting in a much nicer error message:

```
  File "/opt/pycharm-professional/plugins/python/helpers/pydev/pydevd.py", line 1483, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "/opt/pycharm-professional/plugins/python/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/home/janis/PycharmProjects/public-fork-alibi/dev_notebooks/#317-minimal.py", line 37, in <module>
    explanation = explainer.explain(bad_instance)  # breaks here
  File "/home/janis/PycharmProjects/public-fork-alibi/alibi/explainers/anchors/anchor_tabular.py", line 801, in explain
    raise NotFittedError(msg)
alibi.exceptions.NotFittedError: This AnchorTabular instance is not fitted yet. Call 'fit' with appropriate arguments first.
```

The reason for going with `NotFittedError` instead of `AlibiNotFittedError` is that the pre-fix `Alibi` makes for more verbose errors and should be unnecessary as all exceptions derive from `AlibiException` so that they can all be caught at once. My thinking is to open another PR to rename the 2 current exceptions with the `Alibi` prefix to have no prefix (but keep the old classes and inherit from them to prevent immediate breakage of existing user code).

### DIFF
--- a/alibi/exceptions.py
+++ b/alibi/exceptions.py
@@ -41,3 +41,9 @@ class PredictorReturnTypeError(AlibiException, AlibiPredictorReturnTypeError):
     an unexpected or unsupported type.
     """
     pass
+
+class NotFittedError(AlibiException):
+    """
+    This exception is raised whenever a compulsory call to a `fit` method has not been carried out.
+    """
+    pass
--- a/alibi/exceptions.py
+++ b/alibi/exceptions.py
@@ -42,6 +42,7 @@ class PredictorReturnTypeError(AlibiException, AlibiPredictorReturnTypeError):
     """
     pass
 
+
 class NotFittedError(AlibiException):
     """
     This exception is raised whenever a compulsory call to a `fit` method has not been carried out.
--- a/alibi/exceptions.py
+++ b/alibi/exceptions.py
@@ -47,4 +47,8 @@ class NotFittedError(AlibiException):
     """
     This exception is raised whenever a compulsory call to a `fit` method has not been carried out.
     """
-    pass
+
+    def __init__(self, object_name: str):
+        super().__init__(
+            f"This {object_name} instance is not fitted yet. Call 'fit' with appropriate arguments first."
+        )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 [Full Changelog](https://github.com/SeldonIO/alibi/compare/v0.7.0...v0.7.1)
 
 ## Added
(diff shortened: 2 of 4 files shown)


alibi/exceptions.py
```diff
@@ -27,3 +27,14 @@ class AlibiPredictorReturnTypeError(AlibiException):
...
+
+class NotFittedError(AlibiException):
+    """
+    This exception is raised whenever a compulsory call to a `fit` method has not been carried out.
+    """
+
+    def __init__(self, object_name: str):
+        super().__init__(
+            f"This {object_name} instance is not fitted yet. Call 'fit' with appropriate arguments first."
+        )
```
jklaise, line 40: I've decided to put the error message here and allow it to be parametrized with the object name. The message is generic enough to apply to any object that as a compulsory `fit` method so I don't think it's necessary to define messages by hand in the code (which will inevitably become different from each other).

mauicv:
This LGTM. 

I guess one further idea would be to extend the behaviour to all explainers using the `FitMixin` interface. So something like:

```py
class FitMixin(abc.ABC):
    def __init__(self):
        self._fitted=False

    def fit(self, X: Any) -> "Explainer":
        self._fitted=True
        self._fit(X)

    @abc.abstractmethod
    def _fit(self, X: Any) -> "Explainer":
         pass
```

Can go further and run the `_fitted` check in the `explain` method on the base class as well. Possibly this would require further thought though and perhaps makes sense in a later PR.

mauicv (approved):

PR #733, opened 2022-08-02
Rename exception classes to be more concise

jklaise:
This PR renames exception classes to be more concise, specifically removing the `Alibi` prefix. This should not be necessary as all exceptions already inherit from `AlibiException` and the traceback also shows that the specific exceptions are defined in `alibi.exceptions`.

I've kept the old classes around for backwards compatibility just in case any application code already relies on the previously defined exceptions.

This is an accompaniment to #732.

### DIFF
--- a/alibi/exceptions.py
+++ b/alibi/exceptions.py
@@ -14,14 +14,28 @@ class AlibiException(Exception, ABC):
         super().__init__(message)
 
 
-class AlibiPredictorCallException(AlibiException):
+# Legacy exception classes starting with the prefix `Alibi`.
+# These have been kept around for backwards compatibility.
+# Any new exception classes should not start with the `Alibi` prefix.
+class AlibiPredictorCallException:
+    pass
+
+
+class AlibiPredictorReturnTypeError:
+    pass
+
+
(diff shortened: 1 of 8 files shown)


alibi/exceptions.py
```diff
@@ -14,14 +14,28 @@ def __init__(self, message: str) -> None:
...
+
+
+class AlibiPredictorReturnTypeError:
+    pass
+
+
+# Exception classes. These should only inherit from `AlibiException`. Ones inheriting from
+# other classes beginning with the prefix `Alibi` are to do with backwards compatibility as
+# exception classes used to all start with the prefix `Alibi`.`
+class PredictorCallError(AlibiException, AlibiPredictorCallException):
```
jklaise, line 31: I've also standardised to end the exception with `Error` rather than `Exception` which is common practice for specific Python exceptions.

mauicv:
LGTM

Btw, Are we updating the change log on each PR btw or just on release?

mauicv (approved):

jklaise:
Happy to start the trend on updating the changelog on each PR.

PR #736, opened 2022-08-09
Clarify usage of magic numbers in the KL-LUCB bandit algorithm

jklaise:
The constants come from PAC lower bounds proved in the paper: http://proceedings.mlr.press/v30/Kaufmann13.html

### DIFF
--- a/alibi/explainers/anchors/anchor_base.py
+++ b/alibi/explainers/anchors/anchor_base.py
@@ -165,7 +165,10 @@ class AnchorBaseBeam:
         -------
         Level used to update upper and lower precision bounds.
         """
-        # TODO: where do magic numbers come from?
+        # The following constants are defined and used in the paper introducing the KL-LUCB bandit algorithm
+        # (http://proceedings.mlr.press/v30/Kaufmann13.html). Specifically, Theorem 1 proves the lower bounds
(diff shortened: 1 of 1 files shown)


jklaise:
I chose not to factor out the constants at the module level (even if left private) because it's unclear whether there would be any benefit in changing these for practical purposes. This can of course be revisited if at any point we would like to run very detailed experiments, but I wanted to keep this PR purely focused on documentation.

RobertSamoilescu:
Looks good to me. Was wondering if we should also clarify in this PR the following [line](https://github.com/SeldonIO/alibi/blob/65e4edc4f9f19edb47a79bdb045df3d3823a99a2/alibi/explainers/anchors/anchor_base.py#L106)?

jklaise:
@RobertSamoilescu I've tracked down the loop usage to the original Anchors repo by the author, curiously it has been removed as deemed superfluous: https://github.com/marcotcr/anchor/commit/ff0924e6bcaaa7149e2940303cd0b22994112157.

jklaise:
@RobertSamoilescu I'd propose postponing investigating this for a future PR. The motivation for the original bisection is not very clear to me and the loop removal results in different numerical outcomes from this function (I suspect for the application it was not necessary to do the bisection in the first place).

Issue #689, opened 2022-05-31
Intermittent `test_save_cfrl` failure


mauicv:
The `test_save_cfrl` test in `alibi/tests/saving.py` has started failing intermittently. This seems to have started occurring as a result of a change in a dependency as the error could not be recreated locally until the pip environment was updated to be in line with [CI] (https://github.com/SeldonIO/alibi/runs/6640017227?check_suite_focus=true) dependencies.

Currently, this is "fixed" by #688 which just raises the tolerance of the `asset_allclose` method.

mauicv:
__Note__: Error only arises after reloading the CFRL_Explainer which makes me think that we change the accuracy of the weight in saving or loading.

mauicv:
difference arises first [here](https://github.com/SeldonIO/alibi/blob/7e8b45d41cf8b7bc0dbf52db7ac9c861ebb901fd/alibi/explainers/backends/tensorflow/cfrl_base.py#L318)

ascillitoe:
@mauicv I imagine you've already tried lots of this, but some things I can think to try based on a similar experience with `alibi-detect` are:

1. Save an explainer. Then re-open it subsequent times. Compare explanations on each load. Are these deterministic? This would rule out if there is some randomness in the loading procedure, versus differences between pre-saved and loaded.
2. If explanations are the same after each load, save multiple explainers. When each of these are re-loaded, are there differences between their explanations? If so, this would suggest there is randomness in saving perhaps.
3.  "change the accuracy of the weight in saving or loading" - this makes me think it might be a `dtype` issue? I would have thought everything is already `float32`, but maybe something defaults to `float64` during re-loading or something?

mauicv:
These are good ideas. I actually did a test in which instead of running the input into the model via the `__call__` method I just ran it through the original and reloaded models layer by layer and check if there was a difference for a particular layer. Both of these results were the same so the difference seems to arise because of the `__call__` method. Internally of `__call__` it does some stuff and then calls `call` which Keras expects you to overide with your layer ops. I'm assuming it's something it does in `__call__`. 

I'd spent quite a while on it by this point and it seems like it's not really a big issue, I'll return to it at some point though.

ascillitoe:
Yeh agreed, the differences are very very small so not a high priority issue!

PR #737, opened 2022-08-17
Fix intermittency of tests

mauicv:
# What is this

`test_save_cfrl` intermittently fails as per [this](https://github.com/SeldonIO/alibi/issues/689) issue. (See issue notes there). I've set the `rtol` to 0, and the `atol` to `1e-06`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,6 +76,9 @@ target/
 # PyCharm
 .idea/
 
+# VSCode
+.vscode
+
 # pyenv
 .python-version
 
--- a/alibi/tests/test_saving.py
+++ b/alibi/tests/test_saving.py
@@ -448,8 +448,8 @@ def test_save_cfrl(cfrl_explainer, rf_classifier, iris_data):
         assert exp0.meta == exp1.meta
 
         # cfrl is determinstic
-        assert_allclose(exp0.cf["X"].astype(np.float32), exp1.cf["X"].astype(np.float32), rtol=1e-06)
-        assert_allclose(exp0.cf["class"].astype(np.float32), exp1.cf["class"].astype(np.float32), rtol=1e-06)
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -76,6 +76,9 @@ target/
 # PyCharm
 .idea/
 
+# VSCode
+.vscode
```
ascillitoe, line 80: Not happy to see this 😉

alibi/tests/test_saving.py
```diff
@@ -448,8 +448,8 @@ def test_save_cfrl(cfrl_explainer, rf_classifier, iris_data):
         assert exp0.meta == exp1.meta
 
         # cfrl is determinstic
-        assert_allclose(exp0.cf["X"].astype(np.float32), exp1.cf["X"].astype(np.float32), rtol=1e-06)
-        assert_allclose(exp0.cf["class"].astype(np.float32), exp1.cf["class"].astype(np.float32), rtol=1e-06)
+        assert_allclose(exp0.cf["X"].astype(np.float32), exp1.cf["X"].astype(np.float32), atol=1e-06, rtol=0)
```
ascillitoe, line 451: Just for context, what are the approx magnitudes of `exp1.cf["X"]` and `exp1.cf["class"]`? I ask in order to get an idea of how much larger `atol=1e-6` is compared to `rtol=1e-6`, given that `atol + rtol * abs(desired)` is the comparision in `allclose(actual, desired, ...)`.

ascillitoe (approved):
Assuming `atol` is not too big, I agree it makes sense to increase the tolerances for now so that CI passes, as long as we keep #689 open to revisit later...

alibi/tests/test_saving.py
mauicv: between ~1.5 and ~-1.5 ish

ascillitoe: OK `atol` makes a lot of sense if `desired` is going negative, as I guess `rtol` gives an exceedingly small tolerance when `desired` hovers around zero?

PR #744, opened 2022-09-06
Enforce utf-8 when reading README.md in setup.py

ascillitoe:
Alibi equivalent of https://github.com/SeldonIO/alibi-detect/pull/605.


mauicv (approved):
LGTM!

PR #747, opened 2022-09-08
Convert string keys to int keys when returning from fetch_imagenet10

### LINKED ISSUE #746: Patch `fetch_imagenet10` to convert `str` keys to `int` keys
Currently the notebook `similarity_explanations_imagenet.ipynb` is failing on `master` because the file `int_to_str_labels.json` has string keys but the code expects it to have `int` keys, this is because `json` only supports string keys. A fix would be to manually convert the keys to `int` in the `fetch_imagenet10` function before returning.

jklaise:
Fixes #746.

### DIFF
--- a/alibi/datasets/default.py
+++ b/alibi/datasets/default.py
@@ -16,7 +16,6 @@ from alibi.utils.data import Bunch
 
 logger = logging.getLogger(__name__)
 
-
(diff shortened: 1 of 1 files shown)


RobertSamoilescu:
Thanks! Looks good to me and the similarity-based explanation for ImageNet notebook example works now.

PR #750, opened 2022-09-09
Fixed cfproto_cat_adult_ohe example

RobertSamoilescu:
Replaced OHE sparse with OHE dense and fixed display function.


jklaise:
Testing this locally, this is not triggered with `numpy` version `1.20.3` but it is with version `1.22.4`. Since the error originates from a call to `np.percentile` it appears that earlier `numpy` versions may have been able to handle `np.matrix` (which is the result of conversion from sparse to dense matrix in the original notebook) but no longer so.

jklaise (approved):
Thanks Robert, have tested this locally as well.

Issue #748, opened 2022-09-08
AnchorImage  slic start_label=1  by default


RobertSamoilescu:
The current `AnchorImage` implementation assumes that the segmentation labels start at 0. The logic can be seen [here](https://github.com/SeldonIO/alibi/blob/711bc1697cc5b6876b95254116c49ea9665a791d/alibi/explainers/anchors/anchor_image.py#L251-L257):

* first, the algorithm identifies the superpixels to be perturbed by checking where the `mask` is 0. This will return an array of values between 0 and K - 1 (including 0 and K-1), where K is the number of superpixels. The results is stored in `to_perturb`.
* the values in `to_perturbed` are compared with the values in the segmentation mask (i.e., `segments == superpixel`, where `superpixels` is an iterator over `to_perturb`). This assumes that the values of `segments`  are also between 0 and K-1.

It seems that from version 0.19.x the default value of the `start_label`  in the segmentation mask returned by `slic` became 1 which contradicts the anchor sampler assumptions. I believe this means that one superpixel will not be perturbed.  For more info see the PR [here](https://github.com/scikit-image/scikit-image/pull/6242).

There are two possible solution:
* update the `slic` `kwargs` and linclude `'start_label': 0`. This will require to bump the version of `skimage` to 0.17.x 
* update the internal logic to be agnostic to the values in the segmentation mask.

jklaise:
We should also check if it's true that other segmentation methods we use from `skimage` already return segments labelled starting with `1` as this may have an effect of which superpixels we ultimately perturb dependent on the segmentation function (we could have an off-by-one error if other methods return labelling `1,...,K` as we wouldn't then perturb the `K`th segment).

RobertSamoilescu:
The other two methods [here](https://github.com/SeldonIO/alibi/blob/711bc1697cc5b6876b95254116c49ea9665a791d/alibi/explainers/anchors/anchor_image.py#L20-L24) return segments labels starting from 0. 

I guess the only problem is when a segmentation function, possible user specified, does not respect the expected labelling convention.

jklaise:
Closed via #753.

PR #753, opened 2022-09-09
Included 'start_label' for 'slic' segmentation.

RobertSamoilescu:
The latest releases of `skimage` updated the `slic` default indexing to start from 1 instead of 0 for convention reasons. This conflicts with the `alibi` expectation which assumes that the indexing starts from 0.

This PR addresses this issue by forcing the `slic` segmentation indexing to start from 0 through the argument `start_label=0`. The default `slic` kwargs were updated with `start_label=0` as well as the arguments in the example notebook. The docstring were updated as well to describe `alibi` expectation of the segmentation indexing.  Those updates required a bump in the `sklearn` verison to a lower bound version of 0.17.2 (the stable version of 0.17.X).

Addresses issue #748


jklaise (approved):
LGTM!

PR #754, opened 2022-09-09
Fixed pdp num-num y-axis rug.

RobertSamoilescu:
The PR includes a fix to the PDP num-num which didn't display the y-axis rug.
Also updates the method description figures and increases their resolution.
Thanks @jklaise for noticing the bug and proposing the solution.


jklaise (approved):
Nice one!

PR #756, opened 2022-09-13
Reinstate ray installation during Python 3.10 CI

ascillitoe:
Ray [now supports Python 3.10](https://github.com/ray-project/ray/blob/96cceb08e8bf73df990437002e25883c5a72d30c/python/setup.py#L23), so adding it back into the Python 3.10 CI. 

Also note that any tests requiring `ray` were not running at all due to an issue with the syntax of this version check in the github action yaml files (see https://github.com/SeldonIO/alibi-detect/pull/615). This should now be fixed since the version does not need to be checked.


jklaise (approved):
LGTM!

Issue #661, opened 2022-05-13
Enable `test_changed_notebooks` CI workflow for PRs to feature branches


jklaise:
Currently it's only triggered for PRs to `master`.

PR #762, opened 2022-09-15
Run test_changed_notebooks on push/PR to any branch and manual dispatch

jklaise:
Addresses #661.

If happy will do the equivalent for `alibi-detect`.


ascillitoe (approved):
LGTM!

PR #767, opened 2022-09-16
Update licences

ascillitoe:
Changes should be noted in below lists, with ✔️/❌ added depending on if the change is acceptable/unacceptable.

### New deps

- [contourpy](https://github.com/contourpy/contourpy). BSD license. ✔️


ascillitoe:
Curious where `contourpy` is coming from...

Issue #718, opened 2022-07-19
One hot encoding, multicollinearity, dummy trap


RobertSamoilescu:
A standard practice when transforming the label encoding to one-hot encoding (ohe) is to drop the first column to address multicollinearity (dummy trap). Currently there exist some explainers that know how to deal with one-hot encoded features such as: [AnchorTabular](https://github.com/SeldonIO/alibi/blob/e790f52a2b19236cb3beb8be90962f2116057f8d/alibi/explainers/anchors/anchor_tabular.py#L589), [Counterfactuals](https://github.com/SeldonIO/alibi/blob/e790f52a2b19236cb3beb8be90962f2116057f8d/alibi/explainers/cfproto.py#L47). Some utility function used by those are available here. The limitation of using the `ohe` flag (i.e. passing ohe dataset) is that the explainers don't know how to deal with the case where the ohe uses `drop='first'` (see [sklearn docs](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OneHotEncoder.html)), which seems to be a standard procedure/best practice. 

It would be good to see if we can extend the capabilities of our explainers and utility functions to deal with all the arguments available in [sklearn](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OneHotEncoder.html). Otherwise, it would be preferable to mention explicitly in our documentation what "format" the ohe is expected to be in (i.e. full representation).

PR #773, opened 2022-09-20
Fix examples tabular preprocessing ohe drop first.

RobertSamoilescu:
Fixes incorrect suggestion to drop first column for one-hot encoding for data preprocessing.
Updates the notebook and method for the PDP.

Check this [gist](https://gist.github.com/RobertSamoilescu/c609ded901c6c50388b6648a73066fed) to see how dropping first column can affect the model performance.


jklaise:
Thanks Robert for the investigation. Would be great if you could publish the gist on your profile and link to it from this issue (and the `alibi-detect` issue) for posterity.

Btw does `handle_unknown='ignore'` is used to potentially avoid unknown categories due to the random data split?

RobertSamoilescu:
Yes. In the case that the test set contains a category that was not seen when fitted the one-hot encoder preprocessor, then that value would be transformed to `[0, 0, ..., 0]` (i.e., only zeros) if we use`handle_unknown='ignore'`. There are also other behaviours that would throw an error if the category is unknown. For our purposes of a notebook example, I would say that this is good enough. 

I also want to mention that in general would be preferable for binary variables to drop the first column. I did not include that because and I kept the overparametrisation for more consistency reasons. Note that [ord_to_ohe](https://github.com/SeldonIO/alibi/blob/f0aed3a4970e3c39e8ca3b91fbc2e7b13cb6bb65/alibi/utils/mapping.py#L86) is not aware of that and this issue was mentioned in #718. Also in `CFRL`, binary features are overparametrised.  Anyway, I don't think that the overparametrisation for binary features is an issue.

jklaise (approved):

PR #776, opened 2022-09-21
Update shap example notebooks.

jklaise:
- Add cell installing `matplotlib==3.5.3` and explaining why this is necessary to run `shap.summary_plot`
 - Replace `plot_confusion_matrix` with `ConfusionMatrixDisplay.from_estimator` due to upcoming deprecation. This also requires an update of the minimal version of `scikit-learn` to `1.0.0`.
 - Replace `get_feature_names` with `get_feature_names_out` due to upcoming deprecation

Note that I re-ran some notebooks end-to-end and the summary plots are now more narrow than previously, I believe this is due to plotting changes in `shap==0.41`.

Also note that I have not re-run `KernelShap` examples involving regularization as it pollutes the output with `sklearn` warnings, see here: https://github.com/slundberg/shap/pull/2641


RobertSamoilescu:
Looks good to me!

PR #784, opened 2022-09-23
Refactored 'predict' function for BB torch models.

RobertSamoilescu:
The refactoring ensures that the gradients are not computes and the model is in evaluation mode for torch models.

In the previous version, the `detach` function was used. This means that the gradients were computed, but detached afterwards. Note that there is no need to compute the gradients at all as it adds extra computation complexity and memory consumption especially for large models. Furthermore we should try to emphasize the model should be in `eval` mode.


jklaise (approved):
Looks good, thanks!

PR #796, opened 2022-10-03
Update CONTRIBUTING.md with tmate instructions

ascillitoe:
Updates `CONTRIBUTING.md` which was a TODO from #770.


jklaise (approved):
Thanks!

PR #798, opened 2022-10-12
Permutation importance

RobertSamoilescu:
This PR implements the `Permutation Importance` to compute the importance of a feature. The implementation is based on the following papers:
* https://link.springer.com/article/10.1023/A:1010933404324
* https://arxiv.org/abs/1801.01489
 
with an emphasis on the second one since it proposed an unbiased version of the original proposal (first one).
The PR also includes a plotting functionality to visualise the importance of the explained features.

To discuss:
* the current implementation supports permuting tuples of columns which acts similarly with the `KernelSHAP` grouping. The second paper actually mentioned that one can choose to permute multiple columns at once and not only a single column. Furthermore, it can help dealing with OHE until we provide explicit support.
* currently, the loss function has to be specified by the user, and the loss function must receive `y_true`, `y_pred`, and optionally `sample_weights`. Furthermore, the user has to ensure that the output of the model and the loss function (i.e. expected `y_pred`) are compatible. This compatibility must be addressed by the user because we are dealing with a black-box model. The `sklearn` implementation uses a special type of object called `Scorer` (see more [here](https://scikit-learn.org/stable/modules/model_evaluation.html)), but their implementation is again limited to their `sklearn` models. The `sklearn` implementation use [scores](https://github.com/scikit-learn/scikit-learn/blob/36958fb240fbe435673a9e3c52e769f01f36bec0/sklearn/inspection/_permutation_importance.py#L74) (i.e. where higher is better) instead of the loss (i.e. lower is better as in the paper formulation). 
* another thing to consider is to allow the user to specify the loss through a string name as `sklearn` does. This means that we will have to implement internally some loss functions and use them according to the string passed by the user. 
* replace kind `exact -> switch`, `estimate -> divide` to match the paper formulation?


jklaise:
> * the current implementation supports permuting tuples of columns which acts similarly with the `KernelSHAP` grouping. The second paper actually mentioned that one can choose to permute multiple columns at once and not only a single column. Furthermore, it can help dealing with OHE until we provide explicit support.

This looks good as it allow easy extension. Documentation is key here, especially if at first we want to focus on permuting single features, so might want to de-emphasize that it's possible to do more untill we have convincing examples on why you might want to do that.

> * currently, the loss function has to be specified by the user, and the loss function must receive `y_true`, `y_pred`, and optionally `sample_weights`. Furthermore, the user has to ensure that the output of the model and the loss function (i.e. expected `y_pred`) are compatible. This compatibility must be addressed by the user because we are dealing with a black-box model. The `sklearn` implementation uses a special type of object called `Scorer` (see more [here](https://scikit-learn.org/stable/modules/model_evaluation.html)), but their implementation is again limited to their `sklearn` models. The `sklearn` implementation use [scores](https://github.com/scikit-learn/scikit-learn/blob/36958fb240fbe435673a9e3c52e769f01f36bec0/sklearn/inspection/_permutation_importance.py#L74) (i.e. where higher is better) instead of the loss (i.e. lower is better as in the paper formulation).

I have quite a few thoughts about loss function specification in general:
 - Think these should be specified in the constructor so that at `explain` time everything is "set up" and you don't have to think about it anymore.
 - I think being too strict about the argument names (running introspection) might not be great as it adds friction if a user has a custom loss function but they have named their arguments differently (see code comment).
 - Main thing to discuss imo is if we want to keep the loss function formulation or actually follow the `sklearn` formulation and use metrics instead:
   - Pro for loss functions: keeps to the original problem definition
   - Pro for metrics: same usage as `sklearn`
   - Pro for metrics: likely more interpetable. E.g. an increase in loss may not be very interpretable/meaningful for a user, but a drop in a metric is more likely to be.
   - Pro for metrics: there is only one loss that a model is trained on and often times it's "implicit" in the framework, i.e. a user may have to do work to investigate what loss was used and then define a callable that calculates the same loss. Metrics are more intuitive and usually multiple metrics are used to assess a model naturally so also semantically makes sense to talk about multiple metrics rather than multiple losses.
   - Pro for metrics: useful "loss functions" are not actually loss functions, e.g. in the example `1-accuracy` is pretty much `1-0` loss, but really accuracy is probably what is more interpretable anyway (and the model wouldn't be trained with `1-0` loss). Also naming conventions - naming a `1-accuracy` loss as `'accuracy'` is misleading, but naming it something else like `'1-0'` loss is awkward/unfamiliar (what about naming `1-f1` "loss"?).
   - As you can see I'm kind of in favour of metrics :) would be good to also see other existing implementations of this.

> * another thing to consider is to allow the user to specify the loss through a string name as `sklearn` does. This means that we will have to implement internally some loss functions and use them according to the string passed by the user.

I think we should, this would enable easy specification of an explainer (e.g. in a config file) and most people would be happy to use pre-sets. I think the implementation of the functions would be straightforward as they can be generic Python functions, perhaps we can even reuse metric functions from `sklearn`.

> * replace kind `exact -> switch`, `estimate -> divide` to match the paper formulation?

Can see arguments for both, `exact` and `estimate` are quite nice as they are intuitive, the other two  not so much if you don't know the paper, so would lean towards keeping these (or do we forsee other methods in the future that would compute exact/estimate in which case we have a potential name collision?).

alibi/explainers/permutation_importance.py
```diff
@@ -0,0 +1,659 @@
...
+        Loss value.
+        """
+        # get scoring function arguments
+        args = inspect.getfullargspec(loss_fn).args
+
+        if 'y_true' not in args:
+            raise ValueError('The `scoring` function must have the argument `y_true` in its definition.')
+
+        if 'y_pred' not in args:
+            raise ValueError('The `scoring` function must have the argument `y_pred` in its definition.')
```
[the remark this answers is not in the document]
RobertSamoilescu: The alternative would be to loose IMO. We would have the rely on the ordering of the parameters which might lead to errors. For example there are loss functions that are not symmetric in their arguments (e.g. KL). If the ordering of the arguments does not match our expectation this can lead to incorrect results.

RobertSamoilescu:
Don't really see that your argument about searching for the loss that was used is relevant. The loss is used generically to denote that lower is better. 
`1 - accuracy` can be name `error rate`. Also `1 - f1` is used in the Interpretable ML book example.

alibi/explainers/permutation_importance.py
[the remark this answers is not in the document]
jklaise: ok I can see the advantage here, happy to leave this validation in. Since we'll provide built-in loss functions/metrics most people would be able to use those with no effort and custom ones would still be supported, though people may need to write a short wrapper around any existing function to ensure the argument names are as expected by `alibi`.

alibi/explainers/tests/test_permutation_importance.py
```diff
@@ -0,0 +1,360 @@
+import pytest
```
jklaise, line 1: Should we have more "functional" tests like in the PDVariance case, e.g. checking that for synthetic problem an uninformative feature gets a ~1/0 score for ratio/difference and an informative features gets a non-zero score?

Also thinking if we should have a convergence test checking whether the estimate method is close to the exact method with some number of repeats.

jklaise (changes requested):
Nice work Robert!

doc/source/examples/permutation_importance_classification_leave.ipynb
```diff
@@ -0,0 +1,933 @@
+{
```
jklaise, line 1: <p>Since <code>seaborn</code> isn't a required dependency, I would add a sentence about requiring it for the example and a cell calling <code>!pip install seaborn</code>  <span class="ql-cursor">﻿</span></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

jklaise, line 1: <p>of the data instance -&gt; of the data instances</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

jklaise, line 1: <p>Rephrase. "some distributions?" And it's unclear what the difference is wrt to, need a name for the base case for the unconditional label distribution (population label distribution?)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

jklaise, line 1: <p>Need a comment for the reader somewhere in this notebook about what <code>sales</code> means as otherwise it's quite illegible - I suspect it's a feature naming issue. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

jklaise, line 1: <p>defining prediction function -&gt; defining a prediction function</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

jklaise, line 1: <p>"which in our case since it means that they are not relevant for the model" - rewrite, e.g. "which means that they are not relevant for the model"</p><p> </p><p>can simplify last sentence, e.g. "It is worth emphasizing that this does not always happen - using a different metric or loss function can give different results".</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='36'/>

jklaise, line 1: <p>If using type hints, it's good practice to also include the return type hint (check other functions in this notebook too please).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='38'/>

jklaise, line 1: <p>Is it worth mentioning the change of scale for the importance scores as compared to the previous example using a metric? I.e. here we have scores that are generally an order of magnitude higher than previously, although the only change is replacing a metric with a <code>1-metric</code>.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='41'/>

jklaise, line 1: <p>e new -&gt; a new</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='42'/>

doc/source/methods/PermutationImportance.ipynb
```diff
@@ -0,0 +1,463 @@
+{
```
jklaise, line 1: <p>propose -&gt; proposed</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

jklaise, line 1: <p>I think we need to make explicit the difference between <code>score_fns</code> and <code>loss_fns</code> and why there are both options available, i.e. higher-is-better vs lower-is-better and the context of the PFI score. Perhaps  it's even worth using a callout <code>Note</code> box for this. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

jklaise, line 1: <p>Rename <code>metric_fns</code> to <code>metric_fn</code> since it's a single one?  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

jklaise, line 1: <p>as method -&gt; as a method</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

jklaise, line 1: <p>Worth using a callout <code>Note</code> box for this? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

jklaise, line 1: <p>consistency of notation: e_switched -&gt; e_switch ? (check all occurences)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

jklaise, line 1: <p>ration -&gt; ratio</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

jklaise, line 1: <p>ration -&gt; ratio</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

jklaise, line 1: <p>Nice work on the exposition!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi/pull/798/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

doc/source/overview/algorithms.md
```diff
@@ -11,20 +11,21 @@ only requirement is to have access to a prediction function (which could be an A
...
-|[Anchors](../methods/Anchors.ipynb)|BB|local|✔| |✔|✔|✔|✔|For Tabular| |
-|[CEM](../methods/CEM.ipynb)|BB* TF/Keras|local|✔| |✔| |✔| |Optional| |
-|[Counterfactuals](../methods/CF.ipynb)|BB* TF/Keras|local|✔| |✔| |✔| |No| |
-|[Prototype Counterfactuals](../methods/CFProto.ipynb)|BB* TF/Keras|local|✔| |✔| |✔|✔|Optional| |
-|[Counterfactuals with RL](https://docs.seldon.io/projects/alibi/en/stable/methods/CFRL.html)|BB|local|✔| |✔| |✔|✔|✔| |
-|[Integrated Gradients](../methods/IntegratedGradients.ipynb)|TF/Keras|local|✔|✔|✔|✔|✔|✔|Optional| |
-|[Kernel SHAP](../methods/KernelSHAP.ipynb)|BB|local  global|✔|✔|✔| | |✔|✔|✔|
-|[Tree SHAP](../methods/TreeSHAP.ipynb)|WB|local  global|✔|✔|✔| | |✔|Optional| | |
-|[Similarity explanations](../methods/Similarity.ipynb)|WB|local|✔|✔|✔|✔|✔|✔|✔| | |
+| Method                                                                                       |Models|Exp. types|Classification|Regression|Tabular|Text|Image|Cat. data| Train or test |Dist.|
```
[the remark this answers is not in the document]
RobertSamoilescu: I was a bit confused why the ALE had the box checked. I guess it was a mistake there?

alibi/explainers/permutation_importance.py
```diff
@@ -0,0 +1,960 @@
...
+        -------
+        A dictionary having as keys the metric names and as values the permutation feature importance associated
+        with the corresponding metrics.
+        """
+        y_hat = []
+        weights: Optional[List[np.ndarray]] = [] if sample_weight else None
+
+        for i in range(len(X)):
+            # create dataset
+            X_tmp = np.tile(X[i:i+1], reps=(len(X) - 1, 1))
```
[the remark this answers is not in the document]
RobertSamoilescu: It is only quadratic in the response, but not the actual input, thus I am not entirely sure we need to go into those details in the docs. Regarding the size, I would say that is able to handle easily in order of thousands. That's also how `sklearn` does with the PDP and others.

alibi/explainers/permutation_importance.py
```diff
@@ -0,0 +1,960 @@
...
+        loss_feature_importance = PermutationImportance._compute_importances(metric_orig=loss_orig,
+                                                                             metric_permuted=loss_permuted,
+                                                                             kind=kind,
+                                                                             lower_is_better=True)
+
+        # compute feature importance for the score functions
+        score_feature_importance = PermutationImportance._compute_importances(metric_orig=score_orig,
+                                                                              metric_permuted=score_permuted,
+                                                                              kind=kind,
+                                                                              lower_is_better=False)
```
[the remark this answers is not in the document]
RobertSamoilescu: Fair enough ... Was just trying to avoid duplicated code ...

alibi/explainers/tests/test_permutation_importance.py
RobertSamoilescu: I think the following test are sufficient as they are sanity checks on some synthetic datasets:
* [test_compute_exact](https://github.com/RobertSamoilescu/alibi/blob/3dbcdd523229638fdef27f767f945c2d4250e165/alibi/explainers/tests/test_permutation_importance.py#L241)
* [test_compute_estimate](https://github.com/RobertSamoilescu/alibi/blob/3dbcdd523229638fdef27f767f945c2d4250e165/alibi/explainers/tests/test_permutation_importance.py#L287)
* [test_explain_exact](https://github.com/RobertSamoilescu/alibi/blob/3dbcdd523229638fdef27f767f945c2d4250e165/alibi/explainers/tests/test_permutation_importance.py#L332)

I tried to avoid using some randomly generate datasets because of the randomness in the results. Same thing for convergence. Tried to avoid introducing randomness since some tests may occasionally fail if the consistency threshold is too small. I haven't tried, but I remember we had some issues with other explainers. Of course, one can argue that for a large sample size this is very unlikely to happen ....

Let me know if you still consider that we need more tests.

doc/source/examples/permutation_importance_classification_leave.ipynb
RobertSamoilescu: <p>Not sure what comment to add. I rephrased it so I hope now is more intelligible.</p>

doc/source/overview/algorithms.md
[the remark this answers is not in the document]
jklaise: Ah yes, I would say it's a mistake and best to remove the tick for all these global explanation methods that don't have a `fit` stage (as you have done).

alibi/explainers/permutation_importance.py
```diff
@@ -0,0 +1,960 @@
...
+            values the metric functions.
+        y
+            Ground truth targets.
+        y_hat
+            Predicted outcome as returned by the classifier.
+        sample_weight
+            Weight of each sample instance.
+        metrics
+            An optional dictionary of metrics, having as keys the name of the metrics and as value the evaluation of
+            the metrics.
```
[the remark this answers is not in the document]
RobertSamoilescu: I refactored the function so it doesn't have side effect.

alibi/explainers/permutation_importance.py
```diff
@@ -0,0 +1,960 @@
...
+    ESTIMATE = 'estimate'
+
+
+class Kind(str, Enum):
+    """ Enumeration of supported kind. """
+    DIFFERENCE = 'difference'
+    RATIO = 'ratio'
+
+
+METRIC_FNS = {
```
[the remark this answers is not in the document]
RobertSamoilescu: Let me know if referencing to this dictionary is ok now

alibi/explainers/permutation_importance.py
```diff
@@ -153,18 +153,18 @@ class PermutationImportance(Explainer):
 
     def __init__(self,
                  predictor: Callable[[np.ndarray], np.ndarray],
-                 loss_fns: Optional[
+                 loss_fns: Optional[  # type: ignore[misc, valid-type]
                      Union[
-                         str,
-                         List[str],
+                         Literal[tuple(METRIC_FNS["loss"].keys())],
+                         List[Literal[tuple(METRIC_FNS["loss"].keys())]],
```
jklaise, line 159: I don't think `mypy` actually can check this given that the values here are dynamic instead of static (hence the required type-ignore's). On the other hand, [sphinx docs render nicely](https://seldon--798.org.readthedocs.build/projects/alibi/en/798/api/alibi.explainers.html#alibi.explainers.PermutationImportance), docstring on the help message (i.e. `PermutationImportance?`) is properly formatted (!) and we avoid duplicating runtime info and type info. @ascillitoe any thoughts on this?

ascillitoe: Difficult one... I agree `mypy` will probably not be checking this properly (hence the `type: ignore`'s as you say). That being said the docs look nice and it avoids duplication. 

Is there a reason we can't replace `METRIC_FNS` with a separate Enum for the `loss` and `score` functions?

jklaise: Are you suggesting to use enums to also type the arguments? I've been generally avoiding this as it's not as nice to the user (who would presumably see `score_fns: SCORE_FNS` as a type and have to look up our enum) as opposed to `Literal's

ascillitoe: Yes that was what I was thinking. But that's a good point it would be more cumbersome when viewing the api docs.

alibi/explainers/permutation_importance.py
```diff
@@ -41,11 +41,9 @@ class Kind(str, Enum):
 METRIC_FNS = {
     "loss": {
         # regression
-        "max_error": sklearn.metrics.max_error,
         "mean_absolute_error": sklearn.metrics.mean_absolute_error,
         "mean_squared_error": sklearn.metrics.mean_squared_error,
         "mean_squared_log_error": sklearn.metrics.mean_squared_log_error,
-        "median_absolute_error": sklearn.metrics.median_absolute_error,
```
jklaise: Why where these removed?

alibi/explainers/permutation_importance.py
jklaise: Seems like duplication is the only option to satisfy both static analysis and nicely rendered docstrings: https://stackoverflow.com/a/67292548

jklaise: @ascillitoe what do you think about the solution in the link, basically defining the `Literal` type statically (thus duplicating keys) but keeping an `assert` statement to ensure the keys are in sync at runtime? Seems like a reasonable workaround to get both static type-checking working and nice `Literal` docstrings.

ascillitoe: Sorry missed this. My main concern with the duplication was maintainability (wrt to the keys getting out of sync) rather than extra lines of code, so that solution seems like a really nice idea.

alibi/explainers/permutation_importance.py
RobertSamoilescu: See discussion on notion page.

alibi/explainers/permutation_importance.py
jklaise: ok let's do this @RobertSamoilescu

doc/source/methods/PermutationImportance.ipynb
```diff
@@ -142,18 +142,30 @@
...
+   "id": "afb070b1",
+   "metadata": {},
+   "source": [
+    "<div class=\"alert alert-info\">\n",
+    "<b>Note</b> \n",
+    "\n",
+    "As mentioned in the parameter description, depending on the loss or score functions used to measure the model performance, the estimate might not converge to the exact metric value, regardless of the number of times the feature values are permuted.\n",
+    "</div>"
+   ]
+  },
```
jklaise, line 168: Seems like this should be below `n_repeats` instead of below `sample_weight` ?

alibi/explainers/permutation_importance.py
```diff
@@ -38,105 +38,125 @@ class Kind(str, Enum):
...
 
-            .. _sklearn.metrics.mean_squared_error:
-               https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mean_squared_error.html#sklearn.metrics.mean_squared_error
+        .. _sklearn.metrics.log_loss:
+            https://scikit-learn.org/stable/modules/generated/sklearn.metrics.log_loss.html#sklearn.metrics.log_loss
+"""
 
-            .. _sklearn.metrics.mean_squared_log_error:
-                https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mean_squared_log_error.html#sklearn.metrics.mean_squared_log_error
+LossFnsValues = Literal[
```
jklaise, line 85: Nitpick - I would stick to singular for type definitions and also change it to `Name` from `Value` as it's more natural, i.e. `LossFnName` and `ScoreFnName` respectively.

jklaise (approved):

Issue #760, opened 2022-09-15
Address drop in testing coverage due to excluding test files


jklaise:
After excluding test files themselves from test coverage reporting in #751 it has become clear that our actual coverage is smaller than expected. This task is to first list the areas where our testing can be improved and second to start addressing those.

PR #799, opened 2022-10-14
Included PyTorch model tests for CFRL

RobertSamoilescu:
This PR includes the `PyTorch` model test for `CFRL` to increase coverage (i.e. addresses https://github.com/SeldonIO/alibi/issues/760)


alibi/models/pytorch/metrics.py
```diff
@@ -113,7 +113,7 @@ def update_state(self, values: np.ndarray):
 
     def result(self) -> Dict[str, float]:
         """
-        Computes the result according the the reduction procedure.
+        Computes the result according the reduction procedure.
```
jklaise, line 116: the -> to the

jklaise:
@RobertSamoilescu  there's a binary file called `.pymon` in this PR which should be removed.

alibi/models/pytorch/tests/test_cfrl_models.py
```diff
@@ -0,0 +1,104 @@
...
+from alibi.models.pytorch import AE, HeAE
+from alibi.models.pytorch.cfrl_models import (ADULTDecoder, ADULTEncoder,
+                                              MNISTClassifier, MNISTDecoder,
+                                              MNISTEncoder)
+
+
+@pytest.mark.parametrize('output_dim', [10, 20])
+def test_mnist_classifier(output_dim):
+    """ Tests whether the `MNISTClassifier` has a deterministic output when is in the
+     evaluation model and if the output shape matches the expectation. """
```
jklaise, line 12: "when it is in evaluation mode" ?

alibi/models/pytorch/tests/test_cfrl_models.py
```diff
@@ -0,0 +1,104 @@
...
+    x = torch.randn(1, 1, 28, 28)
+    model = MNISTClassifier(output_dim=output_dim).cpu()
```
jklaise, line 14: Btw are these calls to `cpu` to avoid using any local GPUs when running the test suite?

alibi/models/pytorch/tests/test_model.py
```diff
@@ -0,0 +1,618 @@
...
+from alibi.models.pytorch.model import Model
+
+
+class UnimodalModel(Model):
+    """ Simple uni-modal output model. """
+
+    def __init__(self, input_dim: int, output_dim: int):
+        super().__init__()
+        self.fc1 = nn.Linear(input_dim, output_dim, bias=False)
+        self.to(self.device)
```
jklaise, line 24: What's the reason for doing this here (contrasting with other tests where `cpu()` was called explicitly after model instance creation).

alibi/models/pytorch/tests/test_model.py
```diff
@@ -0,0 +1,618 @@
...
+
+
+@pytest.mark.parametrize('multimodal_model', [
+    {
+        'input_dim': input_dim,
+        'output_dims': output_dims,
+    }
+], indirect=True)
+def test_compile_multimodal_mismatch(multimodal_model):
+    """ Test compile function raises and error when multiple loss functions are passed but the number \
```
jklaise, line 165: and error -> an error

alibi/models/pytorch/tests/test_model.py
```diff
@@ -0,0 +1,618 @@
...
+        'compile': True,
+        'optimizer': optimizer_class,
+        'loss': nn.MSELoss(reduction='mean'),
+    }
+], indirect=True)
+def test_train_step(unimodal_model, dataset):
+    """ Test if the train step return the appropriate statistics. """
+    # copy the model `state_dict` as it will be modified by the `train_step`. Not really necessary now,
+    # but to avoid any errors in the future if it is the case.
+    state_dict_cpy = deepcopy(unimodal_model.state_dict())
```
jklaise, line 376: Is the concern here because of fixture sharing between test functions? Could it be done in a more idiomatic way using "yield" fixtures and/or scoping the model fixture at a test-function level instead of module level?

alibi/models/pytorch/tests/test_model.py
```diff
@@ -0,0 +1,618 @@
...
+    # perform test step
+    with reset_model(unimodal_model):
+        results = unimodal_model.test_step(x, y_true)
+
+    # check if gradients are not computed
+    for param in unimodal_model.parameters():
+        assert (param.grad is None) or (torch.allclose(param, torch.zeros_like(param)))
+
+    # compute prediction
+    unimodal_model.eval()
```
jklaise, line 430: Should this be reverted before exiting the test (also see previous comment on fixtures)?

alibi/models/pytorch/tests/test_model.py
```diff
@@ -0,0 +1,618 @@
...
+        assert np.isclose(multimodal_model.metrics[key].total, 0)
+        assert np.isclose(multimodal_model.metrics[key].count, 0)
+
+
+def test_saving():
+    """ Test saving functionality. """
+    model1 = UnimodalModel(input_dim=input_dim, output_dim=output_dim)
+    model2 = UnimodalModel(input_dim=input_dim, output_dim=output_dim)
+
+    with tempfile.TemporaryDirectory() as temp_dir:
```
jklaise, line 612: `pytest` actually ships with a `tmp_path` fixture which can be used, saving some code and imports :) https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

jklaise (changes requested):
Thanks, nice tests! Left a few comments, particularly around fixture management.

alibi/models/pytorch/tests/test_cfrl_models.py
RobertSamoilescu: There is no need for the `cpu` when the CI is running since it only has CPU. By default, on my local machine was running on the GPU and the problem was that the tensors were not sent to the GPU. I forgot to delete them after I set the env variable to avoid running on the GPU.

alibi/models/pytorch/tests/test_model.py
RobertSamoilescu: See comment above. Probably can be deleted and was just added for some consistency. Guess I will just leave it there?

alibi/models/pytorch/tests/test_model.py
```diff
@@ -0,0 +1,618 @@
...
+        return [fc(x) for fc in self.fcs]
+
+
+@contextlib.contextmanager
+def reset_model(model: Model):
+    model._reset_loss()
+    model._reset_metrics()
+    yield
+    model._reset_loss()
+    model._reset_metrics()
```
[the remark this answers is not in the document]
RobertSamoilescu: Yes. That's correct.

alibi/models/pytorch/tests/test_model.py
RobertSamoilescu: yes, the concern is that the model is shared between tests. To be honest, I am not sure how I can re-write it with "yield". Changing the scope to function level will do the job, but what I tried to do is avoid creating a new object again and again for each test. I guess if we decide to move the fixture to function level, then I think there will be no need for the context manager to reset the metrics and the loss. Not entirely sure what's the best approach ... Any suggestions?

alibi/models/pytorch/tests/test_model.py
RobertSamoilescu: Doesn't really matter for those tests since the model has the same behavior in train and eval. Now that I think about it, probably it makes sense to set the scope to function level ...

alibi/models/pytorch/tests/test_model.py
jklaise: I think just initializing a model isn't a huge overhead so think function scope makes sense and we have better guarantees for test isolation.

alibi/models/pytorch/tests/test_model.py
```diff
@@ -318,17 +299,15 @@ def test_compute_loss_multimodal(multimodal_model, dataset):
         'compile': True,
         'optimizer': optimizer_class,
         'loss': loss_fn,
-        'metrics': metric
+        'metrics': [AccuracyMetric()]
```
jklaise, line 302: Since you've changed these to be explicit, should we remove the global variables defined here? https://github.com/SeldonIO/alibi/blob/ea108699258c50284059da8f0d8c684590a0f2b0/alibi/models/pytorch/tests/test_model.py#L48-L49

jklaise (changes requested):
Looks good, just one more comment.

jklaise (approved):

PR #810, opened 2022-10-31
Fix Optional tensor error in similarity base.py

mauicv:
## What is this:

Fixes typing error in `similarity/backend/base.py` that arises due to `tensor.grad` being `None` prior to the backward call. Solved by adding `# type: ignore [arg-type]` to the offending line. The other option was to add an if statement like so:

```py
return np.concatenate([_PytorchBackend.to_numpy(param.grad).reshape(-1)
                               for param in model.parameters() if param.grad is not None])
```
However, doing this means extra overhead that isn't necessary as `param.grad` is accessed after the backward call and so is never `None`.


alibi/tests/test_saving.py
```diff
@@ -340,6 +341,7 @@ def test_save_AnchorImage(ai_explainer, mnist_predictor):
         assert exp0.meta == exp1.meta
 
 
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="Test fails intermittently on windows platform")
```
ascillitoe, line 344: Is this a separate issue? If so, perhaps best to do this in a separate PR so that we can easily revert once fixed.

ascillitoe (approved):
LGTM, pending clarification of https://github.com/SeldonIO/alibi/pull/810#discussion_r1011542450 and CI passing 👍🏻

alibi/tests/test_saving.py
ascillitoe: Done in #812

PR #819, opened 2022-11-10
Fix/pd polt test coverage

RobertSamoilescu:
This PR includes tests for the potting functionality of the `PartialDependence` explainer to increase coverage (i.e. addresses #760 ).


alibi/explainers/tests/test_partial_dependence.py
```diff
@@ -281,36 +297,52 @@ def test_grid_points_error(adult_data, use_int):
...
-        np.testing.assert_allclose(ice_vals, ice_sampled_vals)
-
-    elif n_ice == 'list':
-        size = np.random.randint(1, n_samples)
-        # needs to be sorted because of the np.unique applied inside the _sample_ice
-        n_ice = np.sort(np.random.choice(n_samples, size=size, replace=False)).tolist()
-        ice_sampled_vals = _sample_ice(ice_values=ice_vals, n_ice=n_ice)
-        np.testing.assert_allclose(ice_vals[:, n_ice], ice_sampled_vals)
-
+def compare_feature_values(exp_alibi: Explanation, exp_sklearn: Bunch):
```
jklaise, line 300: I would name this starting with `assert` (e.g. `assert_feature_values_equal`) similar to `numpy.testing` utility functions as it makes clear to the developer that the function performs assertions.

alibi/explainers/tests/test_partial_dependence.py
```diff
@@ -467,21 +487,471 @@ def test_sklearn_blackbox(rf_classifier, adult_data, features, params):
...
 
     # compare explanations
-    assert np.allclose(exp_alibi.pd_values[0][1], exp_sklearn['average'])
-    assert np.allclose(exp_alibi.ice_values[0][1], exp_sklearn['individual'])
+    assert np.allclose(exp_alibi.data['ice_values'][0][1], exp_sklearn['individual'])
+    assert np.allclose(exp_alibi.data['pd_values'][0][1], exp_sklearn['average'])
+    compare_feature_values(exp_alibi=exp_alibi, exp_sklearn=exp_sklearn)
+
+
+@pytest.fixture(scope='function')
```
jklaise, line 505: This seems to be just static data, is there a reason this has to be function scoped?

jklaise: Ah I see it is modified by test functions. Perhaps leave a comment here saying why this needs to be function scoped. Although probably would be better to use a parametrised fixture.

jklaise: Can we just use this pattern? https://docs.pytest.org/en/6.2.x/fixture.html#parametrizing-fixtures

jklaise (changes requested):
Thanks Robert, looks good! I would just prefer if we can parametrize the fixture instead of manually changing it's return value, plus rename any functions that make `assert` statements to start with `assert_`.

jklaise (approved):

PR #820, opened 2022-11-11
Fix/pd variance plot test coverage

RobertSamoilescu:
This PR includes tests for the potting functionality of the `PartialDependenceVariance` explainer to increase coverage (i.e. addresses https://github.com/SeldonIO/alibi/issues/760 ).


alibi/explainers/tests/test_pd_variance.py
```diff
@@ -243,25 +251,384 @@ def test_interaction_num_cat(rf_classifier, pd_pdv_explainers, features, adult_d
...
+            np.array([[19.919, 10.809], [14.462, 50.703], [70.373, 35.073]]),
+            np.array([[65.962, 18.605], [11.311, 14.507], [21.231, 24.657]])
+        ]
+    ]
+    return Explanation(meta=meta, data=data)
+
+
+@pytest.mark.parametrize('n_axes', [4, 5, 6, 7])
+@pytest.mark.parametrize('n_cols', [2, 3, 4])
+@pytest.mark.parametrize('exp', [lazy_fixture('explanation_importance')])
```
jklaise, line 412: Not sure why this needs to be a lazy fixture since it doesn't have it any parameters, can't test functions just depend on this fixture directly as usual?

jklaise (commented):
LGTM! Just not sure we need to use `lazy_fixture` at all.

alibi/explainers/tests/test_pd_variance.py
RobertSamoilescu: Yes ... I think initially I intended to use both types of explanations in the test ... I will update it.

jklaise (approved):

PR #823, opened 2022-11-22
Fix PFI multidimensional ground-truth labels for `exact` method.

RobertSamoilescu:
This PR fixes the computation of the `exact` PFI to handle multidimensional ground-truth labels. The previous implementation assumed that the ground truth label is a unidimensional array.

### DIFF
--- a/alibi/explainers/permutation_importance.py
+++ b/alibi/explainers/permutation_importance.py
@@ -558,16 +558,20 @@ class PermutationImportance(Explainer):
         A dictionary having as keys the metric names and as values the permutation feature importance associated
         with the corresponding metrics.
         """
-        y_hat = []
+        y_perm, y_perm_hat = [], []
         weights: Optional[List[np.ndarray]] = [] if sample_weight else None
 
         for i in range(len(X)):
-            # create dataset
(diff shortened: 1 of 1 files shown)


alibi/explainers/permutation_importance.py
```diff
@@ -558,39 +558,43 @@ def _compute_exact(self,
...
 
         for i in range(len(X)):
-            # create dataset
-            X_tmp = np.tile(X[i:i+1], reps=(len(X) - 1, 1))
+            # create input features dataset
+            X_tmp = np.tile(X[i:i+1], reps=(len(X) - 1, ) + (1, ) * (len(X.shape) - 1))
             X_tmp[:, features] = np.delete(arr=X[:, features], obj=i, axis=0)
 
+            # create ground-truth labels
+            y_tmp = np.tile(y[i:i+1], reps=(len(y) - 1, ) + (1, ) * (len(y.shape) - 1))
```
jklaise, line 570: Is it worth adding shapes in comments for these arrays? Otherwise hard to see what's going on. Would also appreciate a short explanation of the failure case for the previous implementation (an example of multi-dimensional ground truth).

jklaise (approved):

PR #826, opened 2022-11-24
Modularized computation of protoypes importance scores

RobertSamoilescu:
This PR modularises the computation of prototypes importance scores from the visualisation function. This gives the option to compute the importance and use a different tool to plot the 2D representation of the prototypes.


alibi/prototypes/protoselect.py
```diff
@@ -617,10 +612,69 @@ def visualize_image_prototypes(summary: 'Explanation',
...
+    classified according to the 1-KNN classifier (Bien and Tibshirani (2012): https://arxiv.org/abs/1202.5933).
+
+    Parameters
+    ----------
+    reducer
+        2D reducer. Reduces the input feature representation to 2D. Note that the reducer operates directly on the
+        input instances if ``preprocess_fn=None``. If the `preprocess_fn` is specified, the reducer will be called
+        on the feature representation obtained after passing the input instances through the `preprocess_fn`.
+    summary, trainset, preprocess_fn, knn_kw
+        See :py:meth:`alibi.prototypes.protoselect.compute_prototypes_importance` method.
```
jklaise, line 645: I would say this needs a full docstring as will be the main user-facing function. I believe the new `compute_...` function we would not advertise as much? In any case, if both are public, I would just duplicate the docstring.

alibi/prototypes/protoselect.py
```diff
@@ -568,32 +564,31 @@ def visualize_image_prototypes(summary: 'Explanation',
...
-    zoom_ub
-        Zoom upper bound. The zoom will be scaled linearly between `[zoom_lb, zoom_ub]`.
+
+    Returns
+    -------
+    A dictionary containing:
+
+     - ``'prototypes_indices'`` - an array of the prototypes indices.
+
+     - ``'prototypes_importance'`` - an array of prototypes importance.
```
jklaise, line 582: I think more grammatically correct would be to call these `prototype_indices` and `prototype_importances`?

alibi/prototypes/protoselect.py
```diff
@@ -568,32 +564,31 @@ def visualize_image_prototypes(summary: 'Explanation',
...
+
+     - ``'X_protos'`` - an array of raw prototypes.
+
+     - ``'X_protos_ft'`` - an array of preprocessed prototypes. If the ``preprocess_fn=None``, no preprocessing \
```
jklaise, line 586: Should this be `None` (or empty) if there is no pre-processing? Basically do we avoid duplicating the same data

alibi/prototypes/protoselect.py
jklaise: Similar for the name of the function.

jklaise (changes requested):

alibi/prototypes/protoselect.py
```diff
@@ -546,6 +547,79 @@ def _imscatterplot(x: np.ndarray,
     return ax
 
 
+def compute_prototype_importances(summary: 'Explanation',
+                                  trainset: Tuple[np.ndarray, np.ndarray],
+                                  preprocess_fn: Optional[Callable[[np.ndarray], np.ndarray]] = None,
+                                  knn_kw: Optional[dict] = None):
```
jklaise, line 553: Just noticed - missing return type.

jklaise (approved):

PR #829, opened 2022-11-30
Filter gradients of IndexedSlices in tf grad sim backend, fixes #828

### LINKED ISSUE #828: Embedding Layers error in Similarity `_TensorFlowBackend`
The following raises the error `'IndexedSlices' object has no attribute 'numpy'`:

```py
model = tf.keras.models.Sequential([
        tf.keras.layers.Embedding(10, 4, input_shape = (5,)),
        tf.keras.layers.Flatten(), 
        tf.keras.layers.Dense(1)
    ])
    X = tf.random.uniform(shape = (1, 5), minval = 0, maxval = 10, dtype = tf.float32)
    Y = tf.random.uniform(shape = (1, 1), minval = 0, maxval = 10, dtype = tf.float32)
    loss_fn = tf.keras.losses.MeanSquaredError()
    tf_grads = _TensorFlowBackend.get_grads(model, X, Y, loss_fn)
```

<s>This is because the gradient through an embedding layer is not computable.</s> The gradients get returned as `IndexedSlices` which are sparse representations. These raise an error in `explainers.similarity.backends.tensorflow.base` in the following code:

```py
        with tf.device(_TensorFlowBackend.device):
            with tf.GradientTape() as tape:
                output = model(X, training=False)
                loss = loss_fn(Y, output)

            # compute gradients of the loss w.r.t the weights
            grad_X_train = tape.gradient(loss, model.trainable_weights)
            grad_X_train = np.concatenate([w.numpy().reshape(-1) for w in grad_X_train])
```

because if `w` is an instance of `IndexedSlices` it has no `numpy` method.

mauicv:
## What is this:

Bugfix for #828.

## Implementation:

Having spoken to @RobertSamoilescu we've agreed on the following plan of action:

- [x] If the tensor is in a sparse format we try to convert it to dense and then to `np.ndarray`. This will also need to be done for sparse representations in pytorch.
- [x] Finally we filter layers with `trainable=False` 
- [x]  If a user passes a model that has some layers set to `trainable=False` then we raise warning to make it clear that those layers won't be used in the computation. Note this should only happen in the model init call!
- [x] Check non-trainable warning behaviour for batch layers...
- [x] Raise informative error if specific layer doesn't convert to numpy
- [x] Instead of filtering for batch norm layers in trainable layers check, just list non-trainable layers in warning.

## More Todo:

- [x] Ensure correct formatting of module and variable names in docstrings: numpy -> `'numpy'`
- [x] Ensure public methods all have correct docstrings with return types etc...
- [x] Add note to FAQ explaining warnings

## Issues:

1. If we fail to convert a sparse tensor to dense or the gradient tensor has no numpy attribute the original plan was to raise a warning informing the user that some tensors grads are not used.
    - Having thought further about this I'm not sure about this? Seems like we'd end up catching a broad range of errors. In principle any model passed should have gradients computable for all layers so it's not clear that we should discard errors that might arise here!?
    - If an error due to something like the above arises we can suggest the user set trainable to false on the layer in question while we figure out a fix? I'm open to doing this differently though!
    - __UPDATE__: having spoken to @RobertSamoilescu, we've agreed it makes sense to raise error here if we can't the gradient to a numpy array. Other errors will not be attempted to be caught. The error will also suggest tot he user they mark the gradinet as non-trainable so they can still get results.

2. tensorflow has two representations of sparsity, 1. `IndexedSlice` which arises as gradients of embedding layers as we've seen and 2. `SparceTensor`. They have different methods of converting to dense representations for instance the below needs to be used:

    ```py
    if isinstance(grad, tf.SparseTensor):
      grad = tf.sparse.to_dense(grad)
    if isinstance(grad, tf.IndexedSlices):
      grad = tf.convert_to_tensor(grad)
    ```
    
    The issue I have is that I can't find any indication that tensorflow layers would ever have sparse gradients other than the `IndexedSlices`  that results from `tf.gather` in the embedding layer. So I'm not sure the above case is needed. You certainly can't take gradients w.r.t. `SparseTensors`. Doing so yeilds:
    
    ```
    ValueError: Type SparseTensor is not supported as a gradient source or gradient target.
    ```
    
    @jklaise and @RobertSamoilescu can either of you think of any situation in which this would arise? For now, I'm going to remove that logic. If an error due to something like the above arises we can suggest the user set trainable to false on the layer in question while we figure out a fix

3. Checking weather or not models contain non-trainable layers is a little non trivial. For instance the `batch_normalization` layer has non-trainable weights and so if a user includes this in there model naive solutions would trigger this warning. I've set it to filter the batch norm layer as according to [the tensorflow docs](https://www.tensorflow.org/guide/keras/transfer_learning#:~:text=The%20only%20built%2Din%20layer,writing%20new%20layers%20from%20scratch.) this is the only layer that is non-trainable. A user might create a custom layer which would throw warnings but I think that's ok as they should be alerted in this case...
  __UPDATE__:: having spoken to @RobertSamoilescu we've agreed to remove the checks for batch normalization as this kind of approach is inherently non-exhaustive and instead replace with a more informative warning that lists all layers not included in grad sim computation!

### DIFF
--- a/alibi/explainers/similarity/backends/pytorch/base.py
+++ b/alibi/explainers/similarity/backends/pytorch/base.py
@@ -91,7 +91,7 @@ class _PytorchBackend:
         elif isinstance(device, torch.device):
             _PytorchBackend.device = device
         elif device is not None:
-            raise TypeError(("`device` must be a ``None``, ``string``, ``integer`` or "
+            raise TypeError(("`device` must be a ``None``, `string`, `integer` or "
                             f"`torch.device` object. Got {type(device)} instead."))
 
     @staticmethod
--- a/alibi/explainers/similarity/backends/pytorch/base.py
+++ b/alibi/explainers/similarity/backends/pytorch/base.py
@@ -96,10 +96,9 @@ class _PytorchBackend:
         return torch.argmax(X, dim=dim)
 
     @staticmethod
-    def check_all_layers_trainable(model: nn.Module) -> None:
+    def check_all_layers_trainable(model: nn.Module) -> bool:
         """Checks that all layers in a model are trainable."""
         for param in model.parameters():
             if not param.requires_grad:
                 return False
         return True
-                
--- a/alibi/explainers/similarity/backends/pytorch/base.py
+++ b/alibi/explainers/similarity/backends/pytorch/base.py
@@ -57,7 +57,11 @@ class _PytorchBackend:
 
     @staticmethod
     def _grad_to_numpy(grad: torch.Tensor) -> torch.Tensor:
-        """Flattens a gradient tensor."""
+        """Convert graidient to numpy array.
+
+        Converts gradient tensor to flat numpy array. If the gradient is a sparse tensor, it is converted to a dense \
+        tensor first.
+        """
         if grad.is_sparse:
             grad = grad.to_dense()
         grad = grad.reshape(-1).detach().cpu()
--- a/alibi/explainers/similarity/backends/pytorch/base.py
+++ b/alibi/explainers/similarity/backends/pytorch/base.py
@@ -59,7 +59,7 @@ class _PytorchBackend:
     def _grad_to_numpy(grad: torch.Tensor, name: Optional[str] = None) -> torch.Tensor:
         """Convert graidient to numpy array.
 
-        Converts gradient tensor to flat numpy array. If the gradient is a sparse tensor, it is converted to a dense \
+        Converts gradient tensor to flat numpy array. If the gradient is a sparse tensor, it is converted to a dense
         tensor first.
         """
         if grad.is_sparse:
--- a/alibi/explainers/similarity/backends/pytorch/base.py
+++ b/alibi/explainers/similarity/backends/pytorch/base.py
@@ -67,6 +67,7 @@ class _PytorchBackend:
 
         if not hasattr(grad, 'numpy'):
             name = f' for the named tensor: {name}' if name else ''
+            # add comment for devs
             raise TypeError((f'Could not convert gradient to numpy array{name}. To ignore these '
                              'gradients in the similarity computation use `requires_grad=False`.'))
         return grad.reshape(-1).cpu().numpy()
@@ -104,7 +105,15 @@ class _PytorchBackend:
         return torch.argmax(X, dim=dim)
 
     @staticmethod
-    def get_non_trainable(model: nn.Module) -> List[Union[int, str]]:
+    def get_non_trainable(model: nn.Module) -> List[Optional[str]]:
         """Checks that all layers in a model are trainable."""
-        return [name if name else i for i, (name, param) in enumerate(model.named_parameters())
-                if not param.requires_grad]
+
+        params = [name if name else None for name, param in model.named_parameters()
+                  if not param.requires_grad]
+
+        if len(params) == len(list(model.parameters())):
+            raise ValueError('The model has no trainable parameters. This method requires at least'
+                             'one trainable parameter to compute the gradients for. '
+                             'Set `.requires_grad_(True)` on the model')
+
+        return params
--- a/alibi/explainers/similarity/backends/pytorch/base.py
+++ b/alibi/explainers/similarity/backends/pytorch/base.py
@@ -106,7 +106,11 @@ class _PytorchBackend:
 
     @staticmethod
     def get_non_trainable(model: nn.Module) -> List[Optional[str]]:
-        """Checks that all layers in a model are trainable."""
+        """Returns a list of non trainable parameters.
+
+        Returns a list of names of parameters that are non trainable. If no trainable parameter exist we raise
+        a ValueError.
+        """
 
         params = [name if name else None for name, param in model.named_parameters()
                   if not param.requires_grad]
--- a/alibi/explainers/similarity/backends/pytorch/base.py
+++ b/alibi/explainers/similarity/backends/pytorch/base.py
@@ -108,7 +108,7 @@ class _PytorchBackend:
     def _get_non_trainable(model: nn.Module) -> List[Optional[str]]:
         """Returns a list of non trainable parameters.
 
-        Returns a list of names of parameters that are non trainable. If no trainable parameter exist we raise
+        Returns a list of names of parameters that are non trainable. If no trainable parameter exists we raise
         a `ValueError`.
         """
 
--- a/alibi/explainers/similarity/backends/pytorch/base.py
+++ b/alibi/explainers/similarity/backends/pytorch/base.py
@@ -4,7 +4,7 @@ Methods unique to the `pytorch` backend are defined here. The interface this cla
 backend in order to ensure that the similarity methods only require to match this interface.
 """
 
-from typing import Callable, Union, Optional
+from typing import Callable, Union, Optional, List
 
 import numpy as np
 import torch.nn as nn
@@ -104,9 +104,7 @@ class _PytorchBackend:
         return torch.argmax(X, dim=dim)
 
     @staticmethod
-    def check_all_layers_trainable(model: nn.Module) -> bool:
+    def get_non_trainable(model: nn.Module) -> List[Union[int, str]]:
         """Checks that all layers in a model are trainable."""
-        for param in model.parameters():
-            if not param.requires_grad:
-                return False
-        return True
+        return [name if name else i for i, (name, param) in enumerate(model.named_parameters())
+                if not param.requires_grad]
--- a/alibi/explainers/similarity/backends/pytorch/base.py
+++ b/alibi/explainers/similarity/backends/pytorch/base.py
@@ -51,12 +51,12 @@ class _PytorchBackend:
         loss.backward()
         model.train(initial_model_state)
 
-        return np.concatenate([_PytorchBackend._grad_to_numpy(param.grad)  # type: ignore [arg-type] # see #810
-                               for param in model.parameters()
+        return np.concatenate([_PytorchBackend._grad_to_numpy(grad=param.grad, name=name)
+                               for name, param in model.named_parameters()
                                if param.grad is not None])
 
     @staticmethod
-    def _grad_to_numpy(grad: torch.Tensor) -> torch.Tensor:
+    def _grad_to_numpy(grad: torch.Tensor, name: Optional[str] = None) -> torch.Tensor:
         """Convert graidient to numpy array.
 
         Converts gradient tensor to flat numpy array. If the gradient is a sparse tensor, it is converted to a dense \
@@ -64,8 +64,12 @@ class _PytorchBackend:
         """
         if grad.is_sparse:
             grad = grad.to_dense()
-        grad = grad.reshape(-1).detach().cpu()
-        return grad.numpy()
+
+        if not hasattr(grad, 'numpy'):
+            name = f' for the named tensor: {name}' if name else ''
+            raise TypeError((f'Could not convert gradient to numpy array{name}. To ignore these '
+                             'gradients in the similarity computation use `requires_grad=False`.'))
+        return grad.reshape(-1).cpu().numpy()
 
     @staticmethod
     def to_tensor(X: np.ndarray) -> torch.Tensor:
--- a/alibi/explainers/similarity/backends/pytorch/base.py
+++ b/alibi/explainers/similarity/backends/pytorch/base.py
@@ -56,8 +56,8 @@ class _PytorchBackend:
                                if param.grad is not None])
 
     @staticmethod
-    def _grad_to_numpy(grad: torch.Tensor, name: Optional[str] = None) -> torch.Tensor:
-        """Convert graidient to `np.ndarray`.
+    def _grad_to_numpy(grad: torch.Tensor, name: Optional[str] = None) -> np.ndarray:
+        """Convert gradient to `np.ndarray`.
 
         Converts gradient tensor to flat `numpy` array. If the gradient is a sparse tensor, it is converted to a dense
         tensor first.
--- a/alibi/explainers/similarity/backends/pytorch/base.py
+++ b/alibi/explainers/similarity/backends/pytorch/base.py
@@ -94,3 +94,12 @@ class _PytorchBackend:
     def argmax(X: torch.Tensor, dim=-1) -> torch.Tensor:
         """Returns the index of the maximum value in a tensor."""
         return torch.argmax(X, dim=dim)
+
+    @staticmethod
+    def check_all_layers_trainable(model: nn.Module) -> None:
+        """Checks that all layers in a model are trainable."""
+        for param in model.parameters():
+            if not param.requires_grad:
+                return False
+        return True
+                
--- a/alibi/explainers/similarity/backends/pytorch/base.py
+++ b/alibi/explainers/similarity/backends/pytorch/base.py
@@ -50,8 +50,18 @@ class _PytorchBackend:
         loss = loss_fn(output, Y)
         loss.backward()
         model.train(initial_model_state)
-        return np.concatenate([_PytorchBackend.to_numpy(param.grad).reshape(-1)  # type: ignore [arg-type] # see #810
-                               for param in model.parameters()])
+
+        return np.concatenate([_PytorchBackend._grad_to_numpy(param.grad)  # type: ignore [arg-type] # see #810
+                               for param in model.parameters()
+                               if param.grad is not None])
+
+    @staticmethod
+    def _grad_to_numpy(grad: torch.Tensor) -> torch.Tensor:
+        """Flattens a gradient tensor."""
+        if grad.is_sparse:
+            grad = grad.to_dense()
+        grad = grad.reshape(-1).detach().cpu()
+        return grad.numpy()
 
     @staticmethod
     def to_tensor(X: np.ndarray) -> torch.Tensor:
--- a/alibi/explainers/similarity/backends/pytorch/base.py
+++ b/alibi/explainers/similarity/backends/pytorch/base.py
@@ -105,7 +105,7 @@ class _PytorchBackend:
         return torch.argmax(X, dim=dim)
 
     @staticmethod
-    def get_non_trainable(model: nn.Module) -> List[Optional[str]]:
+    def _get_non_trainable(model: nn.Module) -> List[Optional[str]]:
         """Returns a list of non trainable parameters.
 
         Returns a list of names of parameters that are non trainable. If no trainable parameter exist we raise
--- a/alibi/explainers/similarity/backends/pytorch/base.py
+++ b/alibi/explainers/similarity/backends/pytorch/base.py
@@ -68,7 +68,7 @@ class _PytorchBackend:
         if not hasattr(grad, 'numpy'):
             name = f' for the named tensor: {name}' if name else ''
             raise TypeError((f'Could not convert gradient to `numpy` array{name}. To ignore these '
-                             'gradients in the similarity computation set `requires_grad=False` on the '
+                             'gradients in the similarity computation set ``requires_grad=False`` on the '
                              'corresponding parameter.'))
         return grad.reshape(-1).cpu().numpy()
 
@@ -91,7 +91,7 @@ class _PytorchBackend:
         elif isinstance(device, torch.device):
             _PytorchBackend.device = device
         elif device is not None:
-            raise TypeError(("`device` must be a `None`, `string`, `integer` or "
+            raise TypeError(("`device` must be a ``None``, ``string``, ``integer`` or "
                             f"`torch.device` object. Got {type(device)} instead."))
 
     @staticmethod
@@ -116,8 +116,8 @@ class _PytorchBackend:
                   if not param.requires_grad]
 
         if len(params) == len(list(model.parameters())):
-            raise ValueError('The model has no trainable parameters. This method requires at least'
-                             'one trainable parameter to compute the gradients for. '
-                             "Try setting `.requires_grad_(True)` on the model or one of it's parameters")
+            raise ValueError("The model has no trainable parameters. This method requires at least "
+                             "one trainable parameter to compute the gradients for. "
+                             "Try setting ``.requires_grad_(True)`` on the model or one of its parameters.")
 
         return params
--- a/alibi/explainers/similarity/backends/pytorch/base.py
+++ b/alibi/explainers/similarity/backends/pytorch/base.py
@@ -67,9 +67,9 @@ class _PytorchBackend:
 
         if not hasattr(grad, 'numpy'):
             name = f' for the named tensor: {name}' if name else ''
-            # add comment for devs
             raise TypeError((f'Could not convert gradient to numpy array{name}. To ignore these '
-                             'gradients in the similarity computation use `requires_grad=False`.'))
+                             'gradients in the similarity computation set `requires_grad=False` on the '
+                             'corresponding parameter.'))
         return grad.reshape(-1).cpu().numpy()
 
     @staticmethod
--- a/alibi/explainers/similarity/backends/pytorch/base.py
+++ b/alibi/explainers/similarity/backends/pytorch/base.py
@@ -4,7 +4,7 @@ Methods unique to the `pytorch` backend are defined here. The interface this cla
 backend in order to ensure that the similarity methods only require to match this interface.
 """
 
-from typing import Callable, Union, Optional, List
+from typing import Callable, Union, Optional
 
 import numpy as np
 import torch.nn as nn
@@ -105,19 +105,18 @@ class _PytorchBackend:
         return torch.argmax(X, dim=dim)
 
     @staticmethod
-    def _get_non_trainable(model: nn.Module) -> List[Optional[str]]:
-        """Returns a list of non trainable parameters.
+    def _count_non_trainable(model: nn.Module) -> int:
+        """Returns number of non trainable parameters.
 
-        Returns a list of names of parameters that are non trainable. If no trainable parameter exists we raise
+        Returns the number of parameters that are non trainable. If no trainable parameter exists we raise
         a `ValueError`.
         """
 
-        params = [name if name else None for name, param in model.named_parameters()
-                  if not param.requires_grad]
+        num_non_trainable_params = len([param for param in model.parameters() if not param.requires_grad])
 
-        if len(params) == len(list(model.parameters())):
+        if num_non_trainable_params == len(list(model.parameters())):
             raise ValueError("The model has no trainable parameters. This method requires at least "
                              "one trainable parameter to compute the gradients for. "
                              "Try setting ``.requires_grad_(True)`` on the model or one of its parameters.")
 
-        return params
+        return num_non_trainable_params
--- a/alibi/explainers/similarity/backends/pytorch/base.py
+++ b/alibi/explainers/similarity/backends/pytorch/base.py
@@ -57,9 +57,9 @@ class _PytorchBackend:
 
     @staticmethod
     def _grad_to_numpy(grad: torch.Tensor, name: Optional[str] = None) -> torch.Tensor:
-        """Convert graidient to numpy array.
+        """Convert graidient to `np.ndarray`.
 
-        Converts gradient tensor to flat numpy array. If the gradient is a sparse tensor, it is converted to a dense
+        Converts gradient tensor to flat `numpy` array. If the gradient is a sparse tensor, it is converted to a dense
         tensor first.
         """
         if grad.is_sparse:
@@ -67,7 +67,7 @@ class _PytorchBackend:
 
         if not hasattr(grad, 'numpy'):
             name = f' for the named tensor: {name}' if name else ''
-            raise TypeError((f'Could not convert gradient to numpy array{name}. To ignore these '
+            raise TypeError((f'Could not convert gradient to `numpy` array{name}. To ignore these '
                              'gradients in the similarity computation set `requires_grad=False` on the '
                              'corresponding parameter.'))
         return grad.reshape(-1).cpu().numpy()
@@ -91,12 +91,12 @@ class _PytorchBackend:
         elif isinstance(device, torch.device):
             _PytorchBackend.device = device
         elif device is not None:
-            raise TypeError(("`device` must be a None, string, integer or "
-                            f"torch.device object. Got {type(device)} instead."))
+            raise TypeError(("`device` must be a `None`, `string`, `integer` or "
+                            f"`torch.device` object. Got {type(device)} instead."))
 
     @staticmethod
     def to_numpy(X: torch.Tensor) -> np.ndarray:
-        """Maps a `pytorch` tensor to a `numpy` array."""
+        """Maps a `pytorch` tensor to `np.ndarray`."""
         return X.detach().cpu().numpy()
 
     @staticmethod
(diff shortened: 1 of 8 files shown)


jklaise:
Before merging this we should answer a few questions.

As a background, this error was raised when trying to calculate `GradientSimilarity` for a transformer model on text instances with respect to token ids (as opposed to embeddings). The error is to be expected since the layer mapping token ids to embeddings is not differentiable; however, this solution proved to be an effective workaround - just get rid of the results which correspond to non differentiable layers.

On the other hand, one could make the claim that the original (error) is expected behaviour as the user should be aware that the model they're passing in has non-differentiable layers. However, fixing the model so that gradients are only taken from the embedding layer onwards proved to be very difficult in the specific case of `transformers` models which included workarounds splitting the main model pipeline in several steps.

Therefore, the proposed solution here seems like a good alternative. To discuss and investigate:
1. Can there be any potential issues when simply discarding results of non-differentiable layers like in this PR?
2. Should we raise a warning if this happens?
3. Is there a more principled way of filtering the results other than looking at `IndexSlices` (e.g. looking at the existence of gradients)?
4. The `pytorch` backend should have the same behaviour if we are to implement this PR.
@RobertSamoilescu @mauicv

mauicv:
Having looked into this further it's not actually the case that the embedding layer is non-differentiable (see [this](https://biswajitsahoo1111.github.io/post/indexedslices-in-tensorflow/) explanation). It's just that the `tf.keras.layers.Embedding` represents the gradients as Sparse tensors in the form `tf.IndexedSlices` and this is where the error arises as these don't have `.numpy()` method.

Moving forward it makes sense to either 
1. filter for `IndexedSlices` or on something like `hasattr(grad, 'numpy')`. 
2. Or find a way to convert them to the numpy gradinet tensor...

jklaise:
Nice find. Would it be more correct to convert all sparse layers to dense and take gradients into account rather than filtering them out? On the other hand, are we risking memory blowing up?

mauicv:
Having spoken to @RobertSamoilescu we've agreed on the following plan of action:

- [ ] If the tensor is in a sparse format we try to convert it to dense and then to `np.ndarray`. This will also need to be done for sparse representations in pytorch.
- [ ] If the above fails or the tensor has no numpy attribute we raise a warning informing the user that some tensors grads are not used.
- [ ] Finally we filter layers with `trainable=False` and raise warning to user if this occurs.

mauicv:
> Nice find. Would it be more correct to convert all sparse layers to dense and take gradients into account rather than filtering them out? On the other hand, are we risking memory blowing up?

w.r.t. memory usage, agreed this is potentially an issue if the vocabulary or the embedding dimension is very big. I think the correct solution to this is to maintain the sparse representation of the layer grads while computing the similarity. So the `get_grads` method would return `sparse_grads` and `dense_grads`. This is a bit trickier though so we decided to consider it for a later PR. In the meantime, if a user wants to reduce the memory footprint they can set trainable to False on layers...

mauicv:
tensorflow has two representations of sparsity, 1. `IndexedSlice` which arises as gradients of embedding layers as we've seen and `SparceTensor`. They have different methods of converting to dense representations for instance the bellow needs to be used:

```py
if isinstance(grad, tf.SparseTensor):
  grad = tf.sparse.to_dense(grad)
if isinstance(grad, tf.IndexedSlices):
  grad = tf.convert_to_tensor(grad)
```

The issue I have is that I can't find any indication that tensorflow layers would ever have sparse gradients other than the `IndexedSlices`  that results from `tf.gather` in the embedding layer. So I'm not sure the above case is needed. You certainly can't take gradients w.r.t. `SparseTensors`. Doing so yeilds:

```
ValueError: Type SparseTensor is not supported as a gradient source or gradient target.
```

Can either of you think of any situation in which this would arise? For now, I'm going to remove that logic and issue a warning if the layer can't be converted to numpy as planned.

alibi/explainers/similarity/backends/tensorflow/base.py
```diff
@@ -79,3 +98,11 @@ def argmax(X: tf.Tensor, dim=-1) -> tf.Tensor:
         """Returns the index of the maximum value in a tensor."""
         X = tf.math.argmax(X, axis=dim)
         return X
+
+    @staticmethod
+    def get_non_trainable(model: keras.Model) -> List[Union[int, str]]:
+        """Checks if all layers in a model are trainable.
```
RobertSamoilescu, line 104: Update docstring as for `pytorch` backend.

alibi/explainers/similarity/backends/tensorflow/base.py
```diff
@@ -79,3 +98,11 @@ def argmax(X: tf.Tensor, dim=-1) -> tf.Tensor:
...
+
+        Note: batch normalization layers are ignored as they are not trainable by default.
```
RobertSamoilescu, line 106: I don't understand the note. Batch normalization has some trainable parameters. Can you clarify what 'are ignored' means in this context?

alibi/explainers/similarity/backends/tensorflow/base.py
```diff
@@ -79,3 +98,11 @@ def argmax(X: tf.Tensor, dim=-1) -> tf.Tensor:
...
+        """
+        return [getattr(layer, 'name', i) for i, layer in enumerate(model.layers) if not layer.trainable]
```
RobertSamoilescu, line 108: This approach does not work as for the `pytorch` backend.
For example, consider the model:

```python
model = keras.Sequential([
    keras.layers.Dense(10),
    keras.layers.Dense(20),
    keras.Sequential([
        keras.layers.Dense(30),
        keras.layers.Dense(40),
    ])    
])

model.build((None, 10))
```

Calling `model.layers` will return 

```
[<keras.layers.core.dense.Dense at 0x7f54d4332430>,
 <keras.layers.core.dense.Dense at 0x7f54d43323d0>,
 <keras.engine.sequential.Sequential at 0x7f54d44dea30>]
```

Setting on layer inside the sequential layer to non-trainable:
``` python
model.layers[-1].layers[0].trainable = False
```

If we call your function, it returns an empty list, which is incorrect.
``` python
get_non_trainable(model)
```

If we call it with the sequential layer, then it returns the correct answer.
```python
get_non_trainable(model.layers[-1])
```

This shows that the function does not work for nested layers, which are very common. Not sure if I am missing something and if the problem is addressed somewhere else?

RobertSamoilescu (changes requested):

mauicv:
- [ ] Ensure correct formatting of module and variable names in docstrings: numpy -> `'numpy'`
- [ ] Ensure public methods all have correct docstirngs with return types etc...

alibi/explainers/similarity/backends/pytorch/base.py
```diff
@@ -50,8 +50,26 @@ def get_grads(
...
+                               for name, param in model.named_parameters()
+                               if param.grad is not None])
+
+    @staticmethod
+    def _grad_to_numpy(grad: torch.Tensor, name: Optional[str] = None) -> torch.Tensor:
+        """Convert graidient to numpy array.
+
+        Converts gradient tensor to flat numpy array. If the gradient is a sparse tensor, it is converted to a dense \
+        tensor first.
+        """
```
jklaise, line 64: graidient -> gradient

nitpick: remove the `\` in the docstring and just use regular line breaks (not sure if this is the IDE doing it for you)?

Same comment for the `tensorflow` backend.

alibi/explainers/similarity/grad.py
```diff
@@ -128,6 +129,14 @@ def __init__(self,
             task_name=task
         )
 
+        non_trainable_layers = self.backend.get_non_trainable(self.predictor)
+        if non_trainable_layers:
+            layers_msg = 'The following layers are not trainable: '
+            layers = ", ".join([f"'{layer}'" for layer in non_trainable_layers if layer is not None])
```
jklaise, line 135: `if layer is not None` -> are there instances where the returned list containe `None` elements?

alibi/explainers/similarity/backends/tensorflow/base.py
mauicv: Yeah, good catch! :man_facepalming: It's better if I just iterate through the `model.non_trainable_weights` here!

alibi/explainers/similarity/grad.py
```diff
@@ -128,6 +129,14 @@ def __init__(self,
...
+            warning_msg = ("Some layers in the model are not trainable. These layer gradients will not be "
```
jklaise, line 136: `These layer gradients...` - presumably there are no gradients in the first place because the layers are not trainable? Perhaps need a small rephrase...

alibi/explainers/similarity/backends/tensorflow/base.py
jklaise: It's a similar issue for `tensorflow` models which have nesting, e.g. any transformer model (see the model for our text app as an example). I think we should try a recursive approach.

mauicv: I think @RobertSamoilescu misspoke no? The above is a `Keras` model?

alibi/explainers/similarity/backends/pytorch/base.py
```diff
@@ -50,8 +50,26 @@ def get_grads(
...
         model.train(initial_model_state)
-        return np.concatenate([_PytorchBackend.to_numpy(param.grad).reshape(-1)  # type: ignore [arg-type] # see #810
-                               for param in model.parameters()])
+
+        return np.concatenate([_PytorchBackend._grad_to_numpy(grad=param.grad, name=name)
+                               for name, param in model.named_parameters()
+                               if param.grad is not None])
+
+    @staticmethod
+    def _grad_to_numpy(grad: torch.Tensor, name: Optional[str] = None) -> torch.Tensor:
```
jklaise, line 59: `grad` type should probably be a union here since it can be sparse?

alibi/explainers/similarity/backends/pytorch/base.py
```diff
@@ -50,8 +50,26 @@ def get_grads(
...
+        Converts gradient tensor to flat numpy array. If the gradient is a sparse tensor, it is converted to a dense \
+        tensor first.
+        """
+        if grad.is_sparse:
+            grad = grad.to_dense()
+
+        if not hasattr(grad, 'numpy'):
+            name = f' for the named tensor: {name}' if name else ''
+            raise TypeError((f'Could not convert gradient to numpy array{name}. To ignore these '
+                             'gradients in the similarity computation use `requires_grad=False`.'))
```
jklaise, line 71: Is it possible for the user to easily do this? I'm thinking of the situation where a layer has multiple gradient tensors, e.g. weights and bias. Here we would be reporting the tensors not the layers?

A related point might be that this could be annoying if many such layers are present, as the error would be thrown for the first one, and then once it's fixed the next one etc. Should we consider checking the `numpy` attribute of all tensors first (after converting to dense) and then raise one error if there are any that don't have that attribute? Perhaps this is a minor point if in realistic scenarios we don't expect there to be more than a handful of such layers?

jklaise: So it sounds like this check is more of a fail-safe and shouldn't really happen unless something is quite wrong? In which case I think it's fine as we at least have a readable error rather than a non-readable one.

alibi/explainers/similarity/backends/pytorch/base.py
jklaise: Ignore this, was thinking of the `tensorflow` backend.

alibi/explainers/similarity/backends/tensorflow/base.py
```diff
@@ -50,9 +50,28 @@ def get_grads(
 
             # compute gradients of the loss w.r.t the weights
             grad_X_train = tape.gradient(loss, model.trainable_weights)
-            grad_X_train = np.concatenate([w.numpy().reshape(-1) for w in grad_X_train])
+            grad_X_train = np.concatenate([_TensorFlowBackend._grad_to_numpy(w, getattr(w, 'name', None))
+                                           for w in grad_X_train])
         return grad_X_train
 
+    @staticmethod
+    def _grad_to_numpy(grad: tf.Tensor, name: Optional[str] = None) -> tf.Tensor:
```
jklaise, line 58: Type hint for `grad` should be a union including `tf.IndexedSlices`.

alibi/explainers/tests/test_simiarlity/conftest.py
```diff
@@ -143,18 +144,26 @@ def linear_models(request):
...
     """
     Constructs a linear model for `tensorflow`.
     """
-    return keras.Sequential([
-        keras.layers.InputLayer(input_shape=input_shape),
-        keras.layers.Dense(output_shape),
-        keras.layers.Softmax()
+    layers = [
+        tf.keras.layers.InputLayer(input_shape=input_shape),
+        tf.keras.layers.Flatten(),
```
jklaise, line 153: Assume this was added intentionally?

alibi/explainers/tests/test_simiarlity/conftest.py
```diff
@@ -143,18 +144,26 @@ def linear_models(request):
...
+    ]
+    if batch_norm:
+        layers.append(tf.keras.layers.BatchNormalization())
+    layers.extend([
+        tf.keras.layers.Dense(output_shape),
+        tf.keras.layers.Dense(output_shape),
```
jklaise, line 159: Assume the duplication was unintentional?

jklaise: Ah I see the same for `pytorch`, was an extra layer needed for extra tests?

alibi/explainers/tests/test_simiarlity/test_backends.py
```diff
@@ -41,3 +42,90 @@ def test_backends(random_cls_dataset, linear_models):
     torch_grads = np.sort(torch_grads)
     tf_grads = np.sort(tf_grads)
     np.testing.assert_allclose(torch_grads, tf_grads, rtol=1e-04)
+
+
+@pytest.mark.parametrize('trainable_emd, grads_shape', [(True, (61, )), (False, (21, ))])
+def test_tf_embedding_similarity(trainable_emd, grads_shape):
+    """Test GradSim explainer correctly handles sparcity and non-trainable layers for tensorflow.
```
jklaise, line 49: sparcity  -> sparsity

alibi/explainers/tests/test_simiarlity/test_backends.py
```diff
@@ -41,3 +42,90 @@ def test_backends(random_cls_dataset, linear_models):
...
+        tf.keras.layers.Embedding(10, 4, input_shape=(5,), trainable=trainable_emd),
+        tf.keras.layers.Flatten(),
+        tf.keras.layers.Dense(1)
+    ])
+
+    X = tf.random.uniform(shape=(1, 5), minval=0, maxval=10, dtype=tf.float32)
+    Y = tf.random.uniform(shape=(1, 1), minval=0, maxval=10, dtype=tf.float32)
+    loss_fn = tf.keras.losses.MeanSquaredError()
+    tf_grads = _TensorFlowBackend.get_grads(model, X, Y, loss_fn)
+    assert tf_grads.shape == grads_shape  # (4 * 10) + (5 * 4) + 1
```
jklaise, line 65: nitpick: perhaps change to `(4 * 10) * trainable_emd + ...` to emphasize that the first term may be absent so the equation is correct for both tests

alibi/explainers/tests/test_simiarlity/test_backends.py
```diff
@@ -41,3 +42,90 @@ def test_backends(random_cls_dataset, linear_models):
...
+
+
+@pytest.mark.parametrize('trainable_emd, grads_shape', [(True, (61, )), (False, (21, ))])
+@pytest.mark.parametrize('sparse', [True, False])
+def test_pytorch_embedding_similarity(trainable_emd, grads_shape, sparse):
+    """Test GradSim explainer correctly handles sparcity and non-trainable layers for pytorch.
```
jklaise, line 71: sparcity -> sparsity

alibi/explainers/tests/test_simiarlity/test_backends.py
```diff
@@ -41,3 +42,90 @@ def test_backends(random_cls_dataset, linear_models):
...
+        torch.nn.LazyLinear(1)
+    )
+
+    model[0].weight.requires_grad = trainable_emd
+
+    X = torch.randint(0, 10, (1, 5))
+    Y = torch.randint(0, 10, (1, 1), dtype=torch.float32)
+    loss_fn = torch.nn.MSELoss()
+    tf_grads = _PytorchBackend.get_grads(model, X, Y, loss_fn)
+    assert tf_grads.shape == grads_shape  # (4 * 10) + (5 * 4) + 1
```
jklaise, line 89: Same comment as above

alibi/explainers/tests/test_simiarlity/conftest.py
mauicv: Yeah, the issue was there was only one layer that could be set to non-trainable and I felt we probably wanted some weights that were still trainable to make sure that the setting was doing what we want it to. I've replaced it in what I'm currently doing though because the model for checking for non-trainable layers is sufficiently complex now that it doesn't make sense to use the above fixture. i.e. I've added nested sequential layers etc...

alibi/explainers/similarity/backends/pytorch/base.py
mauicv: No your correct, torch also has `SparseTensors`

alibi/explainers/tests/test_simiarlity/test_grad_methods_integration.py
```diff
@@ -138,6 +139,7 @@ def test_correct_grad_dot_sim_result_tf(seed, normed_ds):
     `tensorflow` backend.
     """
     model = keras.Sequential([keras.layers.Dense(1, use_bias=False)])
+    model.build((None, 2))
```
jklaise, line 142: What was the reason for adding these? Since we're hard-coding the input dimension, probably should mention it in a comment or something?

alibi/explainers/tests/test_simiarlity/test_grad_methods_integration.py
```diff
@@ -303,3 +311,62 @@ def test_multiple_test_instances_stored_grads_asym_dot(precompute_grads):
...
+
+@pytest.mark.parametrize('linear_cls_model',
+                         [
+                             ({'framework': 'pytorch', 'input_shape': (10,), 'output_shape': 10, 'batch_norm': True}),
+                             ({'framework': 'tensorflow', 'input_shape': (10,), 'output_shape': 10, 'batch_norm': True})
+                         ],
+                         indirect=True, ids=['torch-model', 'tf-model'])
+def test_no_warnings_batch_norm(linear_cls_model):
+    """
+    Test that no warnings are raised when using batch norm layers.
```
jklaise, line 324: Think need to add another sentence for future developers as not sure I follow the reasoning why specifically batch norm layers shouldn't trigger warnings.

alibi/explainers/tests/test_simiarlity/test_grad_methods_integration.py
```diff
@@ -303,3 +311,62 @@ def test_multiple_test_instances_stored_grads_asym_dot(precompute_grads):
...
+
+@pytest.mark.parametrize('linear_cls_model',
+                         [
+                             ({'framework': 'pytorch', 'input_shape': (10,), 'output_shape': 10}),
+                             ({'framework': 'tensorflow', 'input_shape': (10,), 'output_shape': 10})
+                         ],
+                         indirect=True, ids=['torch-model', 'tf-model'])
+def test_non_trainable_layer_warnings(linear_cls_model):
+    """
+    Test that warnings are raised when user passes non-trainbable layers to grad sim method.
```
jklaise, line 345: non-trainbable -> non-trainable

alibi/explainers/similarity/grad.py
mauicv: I haven't been able to figure out this behaviour. `EagerTensors` don't have name attributes, so those layer names returned would be `None`. I'm leaving this in defensively so that if this behaviour does happen then an error won't occur but the user still receives an error message.

I've also added an error message for the case where no parameter in the model is trainable.

alibi/explainers/similarity/backends/tensorflow/base.py
mauicv: Marking resolved as we discussed offline and agreed to no longer filter for non_trainable weights set by default like in batch norm layers. Instead, we just warn the user by printing a list of such parameters.

alibi/explainers/similarity/backends/pytorch/base.py
mauicv: Ignore me, torch handles sparse tensors by setting a layout attribute on normal tensors!

alibi/explainers/tests/test_simiarlity/conftest.py
mauicv: Sort of, changes were made that have now been reverted. I think I left the above. I've now reverted it also.

alibi/explainers/tests/test_simiarlity/test_grad_methods_integration.py
mauicv: A newly instantialized model is built either explicitly as above or when the user passes input to it. Previously this was done in the grad method but I've since added these warnings that require first checking whether the model weights are trainable or not. If the model hasn't been built by this point an error is thrown. 

I figure the above behaviour is alright as users should only be passing build models to the method and if they aren't the error is descriptive enough for them to figure it out without us adding anything. I'll add a comment explaining it in the tests though.

alibi/explainers/similarity/backends/pytorch/base.py
```diff
@@ -50,8 +50,27 @@ def get_grads(
         loss = loss_fn(output, Y)
         loss.backward()
         model.train(initial_model_state)
-        return np.concatenate([_PytorchBackend.to_numpy(param.grad).reshape(-1)  # type: ignore [arg-type] # see #810
-                               for param in model.parameters()])
+
+        return np.concatenate([_PytorchBackend._grad_to_numpy(grad=param.grad, name=name)
+                               for name, param in model.named_parameters()
+                               if param.grad is not None])
```
mauicv, line 56: Note that torch tensors have a `grad` attribute by default set to `None` so we don't have to check it exist first!

doc/source/overview/faq.md
```diff
@@ -64,3 +64,19 @@ Unfortunately, running this line means it's impossible to run explainers based o
...
+
+
+## Similarity explanations
+
+### I'm using the [GradientSimilarity](../methods/Similarity.ipynb) method on a large model and it runs very slow. If I use `precompute_grads` I get out of memory errors?
+
+Large models with many parameters result in the similarity method running very slow and using `precompute_grads=True` may not be an option due to the memory cost. The best solutions for this problem are:
+
+- Use the explainer on a reduced dataset. You can use [Prototype Methods](../methods/ProtoSelect.ipynb) to obtain a smaller representative sample.
+- Freeze some parameters in the model so that when computing the gradients the simialrity method excludes them. If using [tensorflow](https://www.tensorflow.org/guide/keras/transfer_learning) you can do this by setting `trainable` to false on layers or specific parameters. For [torch](https://pytorch.org/docs/master/notes/autograd.html#locally-disabling-gradient-computation) we can set `requires_grad=False` on the relevent model parameters.
```
jklaise, line 76: Nice explanations!

Just a nitpick - would be nice to be slightly more consistent, e.g. `trainable=False`, and perhaps `precompute_grads=True` in the question also. I would also change `torch` -> `pytorch`.

alibi/explainers/tests/test_simiarlity/test_backends.py
```diff
@@ -41,3 +42,94 @@ def test_backends(random_cls_dataset, linear_models):
     torch_grads = np.sort(torch_grads)
     tf_grads = np.sort(tf_grads)
     np.testing.assert_allclose(torch_grads, tf_grads, rtol=1e-04)
+
+
+@pytest.mark.parametrize('trainable_emd, grads_shape', [(True, (61, )), (False, (21, ))])
+def test_tf_embedding_similarity(trainable_emd, grads_shape):
+    """Test GradSim explainer correctly handles sparsity and non-trainable layers for tensorflow.
```
RobertSamoilescu, line 49: GradSim -> \`GradientSimilarity\`
tensorflow -> `tensorflow`

Although those won't show up in the docs.
If decide to update, please update everywhere

alibi/explainers/tests/test_simiarlity/test_backends.py
```diff
@@ -41,3 +42,94 @@ def test_backends(random_cls_dataset, linear_models):
...
+    Y = torch.randint(0, 10, (1, 1), dtype=torch.float32)
+    loss_fn = torch.nn.MSELoss()
+    pt_grads = _PytorchBackend.get_grads(model, X, Y, loss_fn)
+    assert pt_grads.shape == grads_shape  # (4 * 10) * trainable_emd + (5 * 4) + 1
+
+
+def test_non_numpy_grads_pytorch():
+    """Test that the `pytorch` backend handles non-numpy gradients correctly.
+
+    _PytorchBackend should throw an error if the gradients cannot be converted to numpy arrays.
```
RobertSamoilescu, line 95: _PytorchBackend -> \`_PytorchBackend\`
numpy -> \`numpy\`

alibi/explainers/similarity/backends/pytorch/base.py
```diff
@@ -57,18 +57,19 @@ def get_grads(
 
     @staticmethod
     def _grad_to_numpy(grad: torch.Tensor, name: Optional[str] = None) -> torch.Tensor:
-        """Convert graidient to numpy array.
+        """Convert graidient to `np.ndarray`.
```
jklaise, line 60: graidient -> gradient

alibi/explainers/tests/test_simiarlity/test_grad_methods_integration.py
```diff
@@ -303,3 +334,147 @@ def test_multiple_test_instances_stored_grads_asym_dot(precompute_grads):
     scores = np.array([[50, 50, 2*50**2]], dtype=np.float32) / (denoms + 1e-7)
     scores.sort()
     np.testing.assert_allclose(explanation.scores, scores[:, ::-1], atol=1e-4)
+
+
+def test_non_trainable_layer_warnings_tf():
+    """Test Non trainable layer warnings tensorflow.
```
RobertSamoilescu, line 340: Non trainable -> non-trainable?
I would suggest to follow the docstring conventions for the tests too: full name of the explainer, proper use of single/double ticks etc. I know this won't show up anywhere, but ...

alibi/explainers/tests/test_simiarlity/test_grad_methods_integration.py
```diff
@@ -303,3 +334,147 @@ def test_multiple_test_instances_stored_grads_asym_dot(precompute_grads):
...
+            model,
+            task='regression',
+            loss_fn=loss_tf,
+            sim_fn='grad_dot',
+            backend='pytorch'
+        )
+
+    assert err.value.args[0] == ('The model has no trainable parameters. This method requires at least'
+                                 'one trainable parameter to compute the gradients for. '
+                                 "Try setting `.requires_grad_(True)` on the model or one of it's parameters")
```
RobertSamoilescu, line 480: same as above

alibi/explainers/similarity/backends/tensorflow/base.py
```diff
@@ -55,10 +55,10 @@ def get_grads(
         return grad_X_train
 
     @staticmethod
-    def _grad_to_numpy(grad: tf.Tensor, name: Optional[str] = None) -> tf.Tensor:
-        """Convert graidient to numpy array.
+    def _grad_to_numpy(grad: Union[tf.IndexedSlices, tf.Tensor], name: Optional[str] = None) -> tf.Tensor:
+        """Convert graidient to `np.ndarray`.
```
jklaise, line 59: graidient -> gradient

jklaise: The return type hint says `tf.Tensor`...

alibi/explainers/similarity/backends/pytorch/base.py
jklaise: Return type hint says `torch.Tensor`...

alibi/explainers/similarity/backends/tensorflow/base.py
```diff
@@ -86,11 +87,11 @@ def set_device(device: Union[str, None] = None) -> None:
         if device is None or isinstance(device, str):
             _TensorFlowBackend.device = device
         else:
-            raise TypeError(f"`device` must be a string or None. Got {type(device)} instead.")
+            raise TypeError(f"`device` must be a `string` or `None`. Got {type(device)} instead.")
 
     @staticmethod
     def to_numpy(X: tf.Tensor) -> tf.Tensor:
```
jklaise, line 93: Update return type hint.

alibi/explainers/similarity/grad.py
```diff
@@ -129,21 +129,22 @@ def __init__(self,
...
 
     def fit(self,
             X_train: np.ndarray,
             Y_train: np.ndarray) -> "Explainer":
         """Fit the explainer.
 
-        The GradientSimilarity explainer requires the model gradients over the training data. In the explain method it
-        compares them to the model gradients for the test instance(s). If ``store_grads`` was set to ``True`` on
+        The ``GradientSimilarity`` explainer requires the model gradients over the training data. In the explain method
+        it compares them to the model gradients for the test instance(s). If ``store_grads`` was set to ``True`` on
```
jklaise, line 147: Just noticed that the public parameter name is `precompute_grads` not `store_grads`.

jklaise (changes requested):
Looks good! Left a few minor comments, plus the one we discussed offline here:
 - Testing this on a language model with all the transformer layers set to non-trainable results in a very large warning off the screen as we list all non-trainable parameters by name. Although the names could potentially be useful, we're likely better off changing the warning to output the number of non-trainable parameters detected and adding a few short sentences about the 2 possibilities (user having set them manually or model having some in by default e.g. batch norm).

RobertSamoilescu (changes requested):

alibi/explainers/tests/test_simiarlity/test_backends.py
mauicv: I've updated GradSim -> GradientSimilarity but have not added backticks in the tests

alibi/explainers/tests/test_simiarlity/test_grad_methods_integration.py
```diff
@@ -356,17 +356,17 @@ def test_non_trainable_layer_warnings_tf():
...
     model.layers[1].trainable = False
     model.layers[-1].layers[1].trainable = False
 
-    tensor_names = [tensor.name for tensor in model.layers[1].non_trainable_weights]
-    tensor_names.extend([tensor.name for tensor in model.layers[-1].layers[1].non_trainable_weights])
-    tensor_names.extend([tensor.name for tensor in model.layers[2].non_trainable_weights])
-
+    # tensor_names = [tensor.name for tensor in model.layers[1].non_trainable_weights]
+    # tensor_names.extend([tensor.name for tensor in model.layers[-1].layers[1].non_trainable_weights])
+    # tensor_names.extend([tensor.name for tensor in model.layers[2].non_trainable_weights])
```
jklaise, line 368: These lines can be deleted?

jklaise (approved):

PR #834, opened 2022-12-09
Replace Boston with Diabetes dataset for testing

jklaise:
Due to `scikit-learn=1.2` deprecating the Boston housing dataset we're replacing our test functions to use the Diabetes dataset instead.

As a follow-up to this, we also need to replace 2 examples (most likely using the California housing datset) plus any other mentions of the Boston dataset:
```
examples/ale_regression_boston.ipynb
examples/cfproto_housing.ipynb
```


ascillitoe (approved):
LGTM, subject to CI passing.

Issue #840, opened 2023-01-03
Report test duration in CI


ascillitoe:
To identify long-running tests during CI, it would be helpful for overly long-running tests to report their duration. This could be achieved by:

1. Add a timeout with `pytest-timeout`, so that tests that take longer than a certain (relatively generous) duration would simply be killed. Long-running tests would then be reported as failed.
2. Add a pytest hook to report duration after each test. Something like [this](https://stackoverflow.com/questions/61644097/pytest-hook-to-access-durations).

As noted by @jklaise , with option 2. a minimum duration threshold would be required to avoid cluttering the output.

jklaise:
Just for context, we should also do the command line argument `--durations=50` (report the top 50 long running tests at the end of the test suite). The reason we're considering a hook (or use of a timeout) is to identify long running tests before the test suite has finished. This is required because our MacOS CI tests frequently time out and take almost 6h to run (compared to ~1h on other OSes).

jk:
@ascillitoe I'm not @jklaise

ascillitoe:
> @ascillitoe I'm not @jklaise

Apologies!

jklaise:
Closing as implemented in #849.

PR #849, opened 2023-01-10
Add test execution times and a timeout of 600s

jklaise:
Addresses #840.

This is primarily to see what is causing mac-os CI tests to periodically run to ~6h before getting killed.


jklaise:
@ascillitoe I would advocate including the `--durations=50 --durations-min=1.0` flags by default for `alibi` and `alibi-detect`, whilst the `pytest-timeout` flag should be temporary in case we have frequently hanging tests like we've observed for `mac-os` on CI here. Otherwise the routine usage of `pytest-timeout` in a test suite is not particularly recommended, e.g. https://github.com/pytest-dev/pytest-timeout#thread (nevertheless, note that we currently do use it for notebook tests, the overhead hear would be minimal compared to the time it takes to execute the notebooks).

ascillitoe:
@jklaise including the `--durations=50 --durations-min=1.0` by default makes sense to me, as does only using `pytest-timeout` temporarily given the large number of tests we set-up/teardown.

Just to check, by `pytest-timeout` being temporary does that mean you will be commenting it out once the CI has run (before merge)?

ascillitoe:
p.s. just noticed you didn't include `pytest-timeout` in https://github.com/SeldonIO/alibi-detect/pull/712/files. Maybe it makes sense to include it but have it commented out with a note saying when it should be uncommented?

jklaise:
@ascillitoe it seems I might have to merge this to `master` to investigate the mac-os failing as we seem unable to replicate it on regular PRs, but generally yes, the `pytest-timout` shouldn't be used routinely.

I didn't include it for `alibi-detect` on purpose as we haven't seen any similar failings there (it's also unwieldy, as for the whole CI to pass need to add to all the `tox` environments too.

ascillitoe:
Ah I didn't consider `tox`. Makes sense to leave out in general then!

PR #870, opened 2023-02-01
Replace Python2 type comments with type annotations

### LINKED ISSUE #863: Replace Python 2 type comments with type annotations
Replace Python 2 type comments with type annotations for consistency with https://github.com/SeldonIO/alibi-detect/pull/725.  https://github.com/SeldonIO/alibi/pull/825 should be mergeable afterwards.

mauicv:
## What is this:

Removes python2 type comments with type annotations, fixes #863. Required before we can merge #825.

### DIFF
--- a/alibi/explainers/anchors/anchor_base.py
+++ b/alibi/explainers/anchors/anchor_base.py
@@ -22,7 +22,7 @@ class AnchorBaseBeam:
         """
 
         self.sample_fcn = samplers[0]
-        self.samplers = None  # type: Optional[List[Callable]]
+        self.samplers: Optional[List[Callable]] = None
         # Initial size (in batches) of data/raw data samples cache.
         self.sample_cache_size = kwargs.get('sample_cache_size', 10000)
         # when only the max of self.margin or batch size remain emptpy, the cache is
@@ -44,7 +44,7 @@ class AnchorBaseBeam:
 
         prealloc_size = batch_size * self.sample_cache_size
         # t_ indicates that the attribute is a dictionary with entries for each anchor
-        self.state = {
+        self.state: dict = {
             't_coverage': defaultdict(lambda: 0.),  # anchors' coverage
             't_coverage_idx': defaultdict(set),  # index of anchors in coverage set
             't_covered_true': defaultdict(None),  # samples with same pred as instance where t_ applies
@@ -60,7 +60,7 @@ class AnchorBaseBeam:
             'current_idx': 0,
             'n_features': coverage_data.shape[1],  # data set dim after encoding
             'coverage_data': coverage_data,  # coverage data
-        }  # type: dict
+        }
         self.state['t_order'][()] = ()  # Trivial order for the empty result
 
     @staticmethod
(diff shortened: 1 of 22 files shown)


alibi/explainers/anchors/anchor_base.py
```diff
@@ -351,7 +351,9 @@ def draw_samples(self, anchors: list, batch_size: int) -> Tuple[tuple, tuple]:
             if anchor not in self.state['t_order']:
                 self.state['t_order'][anchor] = list(anchor)
 
-        sample_stats, pos, total = [], (), ()  # type: List, Tuple, Tuple
+        sample_stats: List = []
+        pos: Tuple = tuple()
```
ascillitoe, line 355: Is the `tuple()` preferred to `()`?

ascillitoe: Just noticed we tend to mostly do `tuple()` elsewhere...

mauicv: I tend to use tuple because `()` is ambiguous and can create errors in certain contexts.

alibi/explainers/anchors/language_model_text_sampler.py
```diff
@@ -60,8 +60,10 @@ def __init__(self, model: LanguageModel, perturb_opts: dict, ):
                 self.subwords_mask[vocab[token]] = True
 
         # define head, tail part of the text
-        self.head, self.tail = '', ''  # type: str, str
-        self.head_tokens, self.tail_tokens = [], []  # type: List[str], List[str]
+        self.head: str = ''
```
ascillitoe, line 63: Since this crops up in a few places, we probably need to decide on a convention for handling these single-line multi-definition bits with type annotations. We could do something like:

```python
self.head: str = ""; self.tail: str = ''"
```

alibi/explainers/cfproto.py
```diff
@@ -768,7 +768,7 @@ def fit(self,
                                                  update_feature_range=False)
 
                 # combine abdm and mvdm
-                self.d_abs = {}  # type: Dict
+                self.d_abs: Any = {}
```
ascillitoe, line 771: Why the change from `Dict` to `Any`?

alibi/utils/distance.py
```diff
@@ -122,8 +122,8 @@ def abdm(X: np.ndarray,
     # combine dict for categorical with binned features
     cat_vars_combined = {**cat_vars, **cat_vars_bin}
 
-    d_pair = {}  # type: Dict
-    X_cat_eq = {}  # type: Dict
+    d_pair: dict = {}
```
ascillitoe, line 125: Just noting that elsewhere we've done `dict` -> `Dict` but here it is `Dict -> dict`

ascillitoe (approved):
A few comments but generally LGTM!

alibi/explainers/anchors/language_model_text_sampler.py
mauicv: I'm happy with the above. This [PEP](https://peps.python.org/pep-0526/#global-and-local-variable-annotations) suggest placing the type hints before the multiple assignment like so:

```py
self.head: str
self.tail: str
self.head, self.tail = "", ""
```

alibi/explainers/cfproto.py
mauicv: A mistake! Good catch! Thanks

alibi/explainers/anchors/language_model_text_sampler.py
ascillitoe: I might be being silly, but I can't actually see where they discuss multiple assignment in that PEP? Only a bit about tuple unpacking and conditional branches...

In any case, I sort of think what you posted above is more readable than what I suggested?  Although it doesn't save as many lines...

ascillitoe: @mauicv @jklaise what was the decision with this? Just to leave as separate lines?

jklaise: Didn't seem to be very important, separate lines is nice as corresponds closely to how you might define types in a dataclass.

PR #872, opened 2023-02-07
Replaced `np.object` with `object` in `plot_permutation_importance`.

RobertSamoilescu:
`np.object` is deprecated in latest release of `numpy` (1.24).


mauicv:
Note that I think the latest version the [ci is run on](https://github.com/SeldonIO/alibi/actions/runs/4114696375/jobs/7102408523#step:4:909) is `numpy==1.23.5`. I think other libraries must be pinning the NumPy version.

jklaise:
Can we check if this is backwards compatible with older versions of `numpy`?

jklaise:
Seems like should be fine from at least `1.20`: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

PR #874, opened 2023-02-14
Remove unused `type: ignore` statements

jklaise:
This PR adds the `mypy` options `warn_unused_ignores = True` and `warn_redundant_casts = True` and fixes the resulting errors. This is motivated because `type: ignore` statements can become stale as newer versions of `mypy` fix bugs and libraries against which we typecheck ship typing stubs and fix bugs therein (e.g. `numpy`), see more here: https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-warn-unused-ignores

Will open a similar PR for `alibi-detect` after https://github.com/SeldonIO/alibi-detect/pull/738.

@ascillitoe @mauicv FYI

### DIFF
--- a/alibi/confidence/trustscore.py
+++ b/alibi/confidence/trustscore.py
@@ -123,7 +123,7 @@ class TrustScore:
 
         # make sure Y represents predicted classes, not one-hot encodings
         if len(Y.shape) > 1:
-            Y = np.argmax(Y, axis=1)  # type: ignore
+            Y = np.argmax(Y, axis=1)
 
         if self.filter == 'probability_knn':
             X_filter, Y_filter = self.filter_by_probability_knn(X, Y)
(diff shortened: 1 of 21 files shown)


ascillitoe:
Makes sense!

jklaise:
Not entirely unexpected, but there's some complications with running this on older `python` versions which have older dependencies (e.g. `numpy`) which may not allow removing `type: ignore` or, indeed, suggest removing ones that shouldn't be removed in newer `python` versions... https://github.com/SeldonIO/alibi/actions/runs/4173451223/jobs/7225790896

ascillitoe:
Perhaps controversial, but do we actually care about `mypy` and `flake8` passing for older Python versions. One could argue that its better for our codebase to be fully mypy and flake8 compliant on our most recent supported Python version, rather than include additional complexity to get these checks passing for Python versions that are close to EoL? We could factor out the `mypy`/`flake8` checks to a separate workflow that is only run on `ubuntu-latest, 3.10` for example...

jklaise:
I was thinking similarly, need to think a bit more about implications.

PR #880, opened 2023-02-27
Reset predictor for SHAP.

RobertSamoilescu:
This PR fixes the `reset_predictor` functionality within `KernelShap` and `TreeShap`. The current implementation does not reset the predictor within the wrapped `shap` object. This means that if an explainer is loaded and a different predictor ispassed as argument to `load_explainer` function (i.e., basically just keeping the configuration of the explainer but changing the predictor), still the predictor used in the initialization of the explainer will be used.


jklaise (approved):
LGTM!

PR #896, opened 2023-04-05
Bump tensorflow to 2.12

ascillitoe:
See https://github.com/SeldonIO/alibi-detect/pull/764#issue-1655749490 for summary of important changes in 2.12.


jklaise:
Looks like 2.12 pulls in both `jax` and `ml-dtypes` which is curious. Also noting that there is no licence change for `alibi-detect` (meaning those two were already pulled in previously?).

ascillitoe:
> Looks like 2.12 pulls in both `jax` and `ml-dtypes` which is curious. Also noting that there is no licence change for `alibi-detect` (meaning those two were already pulled in previously?).

I think `jax` uses `ml-dtypes`, and I know tf 2.7 added something to _"Add experimental API experimental_from_jax to support conversion from Jax models to TensorFlow Lite"_, but confused why the `jax` licence is only being added now...

PR #912, opened 2023-04-18
`GradientSimilarity` - allow models of any input type

RobertSamoilescu:
This PR extends the support for the `GradientSimilarity` explainer to support other data types apart from `np.ndarray`, `torch.Tensor` and `tf.Tensor`.

This extension is useful when a user defines a custom model for text modality, for example, for which the input is a list of strings. Note that for `tensorflow` we can get around this issue by casting the input to a `numpy` array of string. This is because `tensorflow` can convert from `numpy` string to `tensorflow` string tensors. Unfortunately, for `pytorch` the string type is not supported.

### DIFF
--- a/alibi/explainers/similarity/base.py
+++ b/alibi/explainers/similarity/base.py
@@ -1,5 +1,5 @@
 from abc import ABC
-from typing import TYPE_CHECKING, Any, Callable, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any, Callable, List, Optional, Union
 
 import numpy as np
 from alibi.api.interfaces import Explainer
@@ -82,11 +82,11 @@ class BaseSimilarityExplainer(Explainer, ABC):
         self
             Returns self.
         """
-        self.X_train: Union[np.ndarray, List[Any]] = X_train
-        self.Y_train: np.ndarray = Y_train
-        self.X_dims: Optional[Tuple] = self.X_train.shape[1:] if isinstance(self.X_train, np.ndarray) else None
-        self.Y_dims: Tuple = self.Y_train.shape[1:]
-        self.grad_X_train: np.ndarray = np.array([])
+        self.X_train = X_train
+        self.Y_train = Y_train
+        self.X_dims = self.X_train.shape[1:] if isinstance(self.X_train, np.ndarray) else None
+        self.Y_dims = self.Y_train.shape[1:]
+        self.grad_X_train = np.array([])
 
         # compute and store gradients
         if self.precompute_grads:
@@ -100,7 +100,7 @@ class BaseSimilarityExplainer(Explainer, ABC):
         return self
 
     @staticmethod
-    def _is_tensor(x: Any):
+    def _is_tensor(x: Any) -> bool:
         """Checks if an obejct is a tensor."""
         if has_tensorflow and isinstance(x, _TfTensor):
             return True


mauicv:
Can you illustrate the use case with a simple example? Is it just that if you have a model with a text embedding layer you might want `X_train` to be a set of strings instead of processed tokens?

alibi/explainers/similarity/grad.py
```diff
@@ -249,7 +250,8 @@ def explain(
         X, Y = self._preprocess_args(X, Y)
```
mauicv, line 250: It may be easier to have a separate `_preprocess_args` function dependent on data type and dispatch to that rather than having lots of conditional statements throughout.

alibi/explainers/similarity/base.py
```diff
@@ -149,18 +156,19 @@ def _compute_adhoc_similarity(self, grad_X: np.ndarray) -> np.ndarray:
         grad_X
             Gradients of the test instances.
         """
-        scores = np.zeros((grad_X.shape[0], self.X_train.shape[0]))
+        scores = np.zeros((len(grad_X), len(self.X_train)))
+        X: Union[np.ndarray, List[Any]]
         for i, (X, Y) in tqdm(enumerate(zip(self.X_train, self.Y_train)), disable=not self.verbose):
-            grad_X_train = self._compute_grad(X[None], Y[None])
+            X = X[None] if hasattr(self.X_train, 'shape') else [X]  # type: ignore[call-overload]
```
mauicv, line 162: Is there a reason you check if `X_train` has a shape attribute here instead of checking its a `np.ndarray`?

mauicv:
LGTM for the most part. I have some minor questions but once they're answered I'm happy to merge.

One thing is that a lot of the logic is conditional statements of the form `if instance(x, np.ndarray) ... else ...` which if may be possible to refactor into something more modular. If the logic is really generic then can we just add a method that deals with each input type:

```py
def _format(self, x):
    if isinstance(x, Tensors):
        # other formatting that makes sense here
        return x[None]
    elif isinstance(x, List):
        return [x] 
```

This way if we add new cases later then it's just a case of extending this function rather than making lots of small changes... I'm not sure the above is possible though, just something to think about!

mauicv (commented):
LGTM for the most part. I have some minor questions but once they're answered I'm happy to merge.

One thing is that a lot of the logic is conditional statements of the form `if instance(x, np.ndarray) ... else ...` which if may be possible to refactor into something more modular. If the logic is really generic then can we just add a method that deals with each input type:

```py
def _format(self, x):
    if isinstance(x, Tensors):
        # other formatting that makes sense here
        return x[None]
    elif isinstance(x, List):
        return [x] 
```

This way if we add new cases later then it's just a case of extending this function rather than making lots of small changes... I'm not sure the above is possible though, just something to think about!

alibi/explainers/similarity/base.py
```diff
@@ -131,12 +134,16 @@ def _match_shape_to_data(self,
             If the shape of `data` does not match the shape of the training data, or fit has not been called prior to
             calling this method.
         """
-        target_shape = getattr(self, f'{target_type}_dims')
-        if data.shape == target_shape:
-            data = data[None]
-        if data.shape[1:] != target_shape:
-            raise ValueError((f'Input `{target_type}` has shape {data.shape[1:]}'
-                              f' but training data has shape {target_shape}'))
+        if hasattr(data, 'shape'):
```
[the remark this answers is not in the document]
RobertSamoilescu: The code still works either if you are passing a single instance or a batch of instances. Your logic hasn't changed. I only added the case in which `data` is not a tensor. In that case, if it is a list, all is fine. If it is not a list, it is wrapped in a list. Similar to what you are doing by adding a batch dimension if necessary.

alibi/explainers/similarity/base.py
```diff
@@ -79,18 +79,21 @@ def fit(self,
         self
             Returns self.
         """
-        self.X_train: np.ndarray = X_train
+        self.X_train: Union[np.ndarray, List[Any]] = X_train
         self.Y_train: np.ndarray = Y_train
-        self.X_dims: Tuple = self.X_train.shape[1:]
+        self.X_dims: Optional[Tuple] = self.X_train.shape[1:] if isinstance(self.X_train, np.ndarray) else None
```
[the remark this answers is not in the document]
RobertSamoilescu: `self.X_dims` represents the dimension of the data, excluding the batch dimension. If you provide any other data type which doesn't have shape (e.g. text data), then the shape is not defined anymore.

alibi/explainers/similarity/base.py
RobertSamoilescu: This is just a very quick prototype and I choose to go checking the `shape` attribute for consistency. There is another place in the code in which I need to check if it's either `np.ndarray` or `tf.Tensor` or `pt.Tensor`. I think the problem with checking whether is either `tf.Tensor` or `pt.Tensor` is whether or not the dependencies are installed. So I think I need to check that first.

alibi/explainers/similarity/base.py
```diff
@@ -149,18 +156,19 @@ def _compute_adhoc_similarity(self, grad_X: np.ndarray) -> np.ndarray:
...
         return scores
 
     def _compute_grad(self,
-                      X: 'Union[np.ndarray, tensorflow.Tensor, torch.Tensor]',
+                      X: 'Union[np.ndarray, tensorflow.Tensor, torch.Tensor, List[Any]]',
                       Y: 'Union[np.ndarray, tensorflow.Tensor, torch.Tensor]') \
             -> np.ndarray:
-        """Computes predictor parameter gradients and returns a flattened `numpy` array."""
-
+        """ Computes predictor parameter gradients and returns a flattened `numpy` array."""
```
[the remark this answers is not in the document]
RobertSamoilescu: Will remove it. Don't know how that happened.

alibi/explainers/similarity/base.py
```diff
@@ -79,21 +82,42 @@ def fit(self,
         self
             Returns self.
         """
-        self.X_train: np.ndarray = X_train
+        self.X_train: Union[np.ndarray, List[Any]] = X_train
         self.Y_train: np.ndarray = Y_train
```
jklaise, line 86: Not sure these type hints are needed here as they haven't changed from the definitions in the `fit` signature? `mypy` should be able to infer that the types are the same.

alibi/explainers/similarity/base.py
```diff
@@ -79,21 +82,42 @@ def fit(self,
...
             for X, Y in tqdm(zip(self.X_train, self.Y_train), disable=not self.verbose):
-                grad_X_train = self._compute_grad(X[None], Y[None])
+                grad_X_train = self._compute_grad(self._format(X), Y[None])
                 grads.append(grad_X_train[None])
+
             self.grad_X_train = np.concatenate(grads, axis=0)
         return self
 
+    @staticmethod
+    def _is_tensor(x: Any):
```
jklaise, line 103: Return type should be `bool`

jklaise (changes requested):
Looks good, a few minor comments.

Was wondering if there would be much effort to add a small end-to-end test calculating explanations on a model operating on `List[Any]`?

mauicv (approved):
LGTM

PR #925, opened 2023-05-19
Separate code quality step and only run on 3.10

jklaise:
It makes sense to run lints and type-checking only on the latest supported version (currently 3.10) as this will unblock improvement in tooling (e.g. see #874).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,6 @@ on:
 jobs:
   build:
     if: github.event.pull_request.draft == false
-    needs: code-quality
 
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


jklaise:
Ironically using `code-quality` as a pre-requisite for the test jobs takes overall longer just because of the length of time needed to install our dependencies. Perhaps better to remove the pre-requisite so that jobs can start asap (they would be cancelled if `code-quality` fails? @ascillitoe

jklaise:
@ascillitoe I was actually wrong, the test jobs wouldn't be cancelled (perhaps there is a setting). Currently they only get cancelled if one of the tests in the build fails.

PR #940, opened 2023-07-04
Replace grid_resolution np.inf value with large number

mauicv:
## What is this

Fixes CI failing due to sklearn parameter validation in `partial_dependence` function

Sklearn has added parameter validation for `partial_dependence` [here](https://github.com/scikit-learn/scikit-learn/blob/7f9bad99d6e0a3e8ddf92a7e5561245224dab102/sklearn/inspection/_partial_dependence.py#L367) which we use in the tests for our partial dependence implementation. This means the ci is failing as [we set `grid_resolution` to `np.inf`](https://github.com/SeldonIO/alibi/blob/c41b6d886e4c740272d2a49cf7181d31676277a5/alibi/explainers/tests/test_partial_dependence.py#L435C1-L436C1) where now Sklearn expects that parameter to be less than infinity.

This PR just replaces `np.inf` in the relevant tests to a large number. This is to check if the test suite failed elsewhere. This is still a work in progress.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest ]
-        python-version: [ '3.7', '3.8', '3.9', '3.10', '3.11']
+        python-version: [ '3.8', '3.9', '3.10', '3.11']
         include: # Run macos and windows tests on only one python version
           - os: windows-latest
             python-version: '3.9'  # PyTorch doesn't yet have 3.10 support on Windows (https://pytorch.org/get-started/locally/#windows-python)
--- a/.github/workflows/test_all_notebooks.yml
+++ b/.github/workflows/test_all_notebooks.yml
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false  # Continue to run other builds despite a failure
       matrix:
         os: [ubuntu-latest]
-        python-version: [ '3.7', '3.8', '3.9', '3.10' ]
+        python-version: [ '3.8', '3.9', '3.10' ]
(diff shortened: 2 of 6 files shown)


RobertSamoilescu:
The `inf` value was used to avoid the division of the categorical feature axis into equidistant bins controlled by the `grid_resolution` parameter. In principle, any value bigger than the cardinal of all categorical feature values should work. 

At that time, `sklearn` was supporting only numerical features, and using the `inf` value was a way around to support categorical features as well. Note that at the moment, `sklearn` included support for categorical features (see PR [here](https://github.com/scikit-learn/scikit-learn/pull/18298)). Thus, I would recommend to use this new feature instead of the large number, only if we don't run into any dependency problems.

mauicv:
I've updated the sklearn `partial_dependence` call to use `categorical_features` instead of `grid_resolution`. This fixes the issue for python >= 3.8 but sklearn 1.3.0 is not supported for python == 3.7 so the `categorical_features` functionality has not been added and this introduces an error. I've added some logic so that the test checks for which sklearn version we're using.

ascillitoe:
Ci failing due to https://github.com/SeldonIO/alibi/issues/943

ascillitoe:
I'm in favour of the second option i.e. keeping the conditional sklearn version behaviour in to support running tests with a new Python version but older sklearn version.

ascillitoe:
@mauicv sorry to be a pain, now we know that the tests pass with `shap 0.42.0`, would you mind reverting the bump in this PR? Then we can merge this PR and do a final check of https://github.com/SeldonIO/alibi/pull/944 before merging that. (just to keep the two changes separate in case of needing to revert etc...)

PR #947, opened 2023-07-17
Run Windows and MacOS CI on Python 3.11

jklaise:
Seems that `torch` support is there and since we only run one version on each non-linux platform we might as well run on the latest supported.


jklaise:
Requires investigation why `torch` 1.x is being pulled in since we support 2.x now.

jklaise:
Ah there's a bug in `extras_require` as `torch` version hasn't been bumped. Will resolve here and in `alibi-detect`.

jklaise:
Fixed another bug with missing forwardref causing an import error when `ray` isn't present: https://github.com/SeldonIO/alibi/commit/17b7267b3a1e3941f1249d1f0d9da6cf98251d4c

jklaise:
Tests passed so will merge for 3.11 testing.

PR #1009, opened 2024-04-18
build(make): Add helper target and install all optionals by default

jesse-c:
This is how it's done in Alibi Detect [1].

[1] https://github.com/jesse-c/alibi-detect/blob/cefb3dd253595ab2385db130269536beac85287f/Makefile#L1-L7


mauicv (approved):
LGTM!

PR #1033, opened 2024-12-05
Included Python3.12 support

RobertSamoilescu:


.github/workflows/ci.yml
```diff
@@ -35,7 +35,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest ]
-        python-version: ['3.9', '3.10', '3.11']
+        python-version: ['3.9', '3.10', '3.11', '3.12']
         include: # Run windows tests on only one python version
           - os: windows-latest
             python-version: '3.11'
```
sakoush, line 41: should this be 3.12 as well?

.github/workflows/ci.yml
```diff
@@ -35,7 +35,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest ]
-        python-version: ['3.9', '3.10', '3.11']
+        python-version: ['3.9', '3.10', '3.11', '3.12']
```
sakoush, line 38: do we also support 3.13? if so we can add it here for the tests?

sakoush (approved):
LGTM

.github/workflows/ci.yml
RobertSamoilescu: I would rather do it in a separate PR.

.github/workflows/ci.yml
RobertSamoilescu: Maybe another PR. Not really a priority atm

